PR #25, opened 2023-03-21
Use panic-probe behavior for defmt panic handler

ia0:


ia0:
Waiting for `panic-probe` to bump after 0.3.0 and include https://github.com/knurling-rs/defmt/pull/737. The PR would need to be adapted to remove the `[patch.crates-io]` section.

PR #99, opened 2023-04-06
Upgrade all dependencies

ia0:


ia0:
The defmt ecosystem is currently broken. Will upgrade dependencies at a later time.

PR #144, opened 2023-05-08
Support in-place AES-256-GCM

ia0:


crates/api-desc/src/crypto/gcm.rs
```diff
@@ -47,6 +47,9 @@ pub(crate) fn new() -> Item {
                 length: usize,
 
                 /// The clear text.
+                ///
+                /// A null pointer indicates that the clear text is in the cipher text and should be
+                /// encrypted in place.
```
ia0, line 52: Let's change `is_supported` to return a bit flag:
- 1 if 2-buffers supported
- 2 if in-place supported
- 3 if both supported (without copy)

ia0: We could also encode in those flags whether there is hardware support and whether the implementation passed CAVP tests or other certifications.

PR #211, opened 2023-08-25
Add debug::perf() for performance measurement

ia0:


crates/scheduler/src/lib.rs
```diff
@@ -267,7 +281,9 @@ impl<B: Board> Scheduler<B> {
     fn call(&mut self, inst: InstId, name: &'static str, args: &[u32]) {
         debug!("Schedule thread {}{:?}.", name, args);
         let args = args.iter().map(|&x| Val::I32(x)).collect();
+        self.perf.record(perf::Slot::Platform);
         let answer = self.applet.store_mut().invoke(inst, name, args).map(|x| x.forget());
+        self.perf.record(perf::Slot::Applets);
```
ia0, line 286: This code should be gated behind a `debug` feature.

PR #268, opened 2023-10-24
Notify the runner when the web UI is ready

invernizzi:


ia0 (approved):

crates/runner-host/crates/web-server/public/board.js
```diff
@@ -345,6 +345,9 @@ class Board {
       this._components.push(component);
     }
     this._drawer.appendStatus("Board is ready!");
+    this._channel.send({
+      type: "board_ready",
```
ia0, line 349: I agree it's a bit ugly, but right now the messages from runner to board use `type` while those from board to runner use `componentType`. Do you see a better alternative?

```suggestion
      componentType: "board_ready",
```

ia0 (changes requested):

ia0 (approved):

PR #288, opened 2023-11-09
Support native applets

### LINKED ISSUE #31: Support native Rust applets
When the applet is written in Rust and it is not a security problem for the applet to be compiled as a native static library and linked with the platform (for example if the applet doesn't use unsafe and is cooperative), then users should be able to do so for performance, footprint, or other reasons.

A possible idea to do so would be to require the applet to be async. When a host function is called, the following would happen:
- The host function would store some information for the scheduler and yield.
- The scheduler would execute the function, store the result for the applet, and resume the applet.
- The host function would then see the result and return it.

Scheduling between applets and threads can only be done when applets call a host function. But given that this feature should be used when the applet is somehow trusted, it seems reasonable that it is also well-behaving and yielding regularly enough.

ia0:
Fixes #31

### DIFF
--- a/crates/scheduler/src/event.rs
+++ b/crates/scheduler/src/event.rs
@@ -121,8 +121,6 @@ pub fn process<B: Board>(scheduler: &mut Scheduler<B>, event: Event<B>) {
         use alloc::boxed::Box;
 
         use crate::native;
-        #[cfg(feature = "debug")]
-        use crate::perf;
 
         #[derive(Copy, Clone)]
         #[repr(transparent)]
@@ -151,8 +149,6 @@ pub fn process<B: Board>(scheduler: &mut Scheduler<B>, event: Event<B>) {
                 native::schedule_callback(Box::new(move || unsafe { $cb(ptr, this $(, $x)*) }));
             }};
         }
-        #[cfg(feature = "debug")]
-        scheduler.perf.record(perf::Slot::Platform);
         match params.len() - 2 {
             0 => schedule!(applet_cb0()),
             1 => schedule!(applet_cb1(x0)),
@@ -160,7 +156,5 @@ pub fn process<B: Board>(scheduler: &mut Scheduler<B>, event: Event<B>) {
             3 => schedule!(applet_cb3(x0, x1, x2)),
             _ => unimplemented!(),
         }
-        #[cfg(feature = "debug")]
-        scheduler.perf.record(perf::Slot::Applets);
     }
 }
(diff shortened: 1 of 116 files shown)


crates/scheduler/src/lib.rs
```diff
@@ -192,17 +223,62 @@ impl<B: Board> Scheduler<B> {
...
+            Err(_) => log::panic!("Failed to find {:?}.", name),
+        };
+        let api_id = self.host_funcs[index].id();
+        let desc = api_id.descriptor();
+        let params = unsafe { core::slice::from_raw_parts(params, desc.params) };
+        let results = unsafe { core::slice::from_raw_parts_mut(results, desc.results) };
+        let erased = SchedulerCallT { scheduler: self, params, results };
+        let call = api_id.merge(erased);
+        log::debug!("Calling {}", log::Debug2Format(&call.id()));
+        call::process(call);
```
ia0, line 262: This should be accounted to the platform timing.

crates/scheduler/src/event.rs
```diff
@@ -104,10 +110,57 @@ pub fn process<B: Board>(scheduler: &mut Scheduler<B>, event: Event<B>) {
...
+        scheduler.perf.record(perf::Slot::Platform);
+        match params.len() - 2 {
+            0 => schedule!(applet_cb0()),
+            1 => schedule!(applet_cb1(x0)),
+            2 => schedule!(applet_cb2(x0, x1)),
+            3 => schedule!(applet_cb3(x0, x1, x2)),
+            _ => unimplemented!(),
+        }
+        #[cfg(feature = "debug")]
+        scheduler.perf.record(perf::Slot::Applets);
```
ia0, line 164: We only schedule here, this should not be accounted to the applet.

crates/scheduler/src/native.rs
```diff
@@ -0,0 +1,98 @@
...
+    lock(|| f(unsafe { SCHEDULER.as_deref_mut().unwrap() }))
+}
+
+pub(crate) fn schedule_callback(callback: Box<dyn Fn() + Send>) {
+    assert!(critical_section::with(|cs| CALLBACK.replace(cs, Some(callback))).is_none());
+}
+
+pub(crate) fn execute_callback() {
+    if let Some(callback) = critical_section::with(|cs| CALLBACK.take(cs)) {
+        callback();
```
ia0, line 86: We should account this to the applet.

crates/scheduler/src/native.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+#[no_mangle]
+extern "C" fn env_dispatch(link: *const c_char, params: *const u32, results: *mut u32) {
+    let link = unsafe { CStr::from_ptr(link) };
+    with_scheduler(|scheduler| {
```
ia0, line 93: We should stop accounting time to the applet here.

PR #324, opened 2023-12-13
Support BLE advertisement listening

jmichelp:


crates/api-desc/src/radio.rs
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::*;
+
+pub(crate) fn new() -> Item {
+    let docs = docs! {
+        /// Radio operations.
+    };
+    let name = "radio".into();
```
ia0, line 21: How general is the term radio? Does it cover other things than what this API is doing? If yes, then we should nest those functions below another module named according to what this API is doing within the radio realm.

For example, if we were to support Bluetooth, Zigbee, Wifi, or other such APIs, they should be at the same level as this API (which is suspect is something like BLE).

crates/api-desc/src/radio.rs
```diff
@@ -0,0 +1,63 @@
...
+                handler_func: fn { data: *mut u8 },
+
+                /// The opaque data to use when calling the handler function.
+                handler_data: *mut u8,
+            } -> {}
+        },
+        item! {
+            /// Unregister handlers for radio events.
+            fn unregister "rd" {
+            } -> {}
```
ia0, line 53: nit

```suggestion
            fn unregister "rd" {} -> {}
```

crates/board/src/radio.rs
```diff
@@ -0,0 +1,58 @@
...
+// limitations under the License.
+
+//! Radio interface.
+
+use crate::{Error, Support, Unsupported};
+
+/// Radio event.
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Event {
+    /// A radio packet has been received
```
ia0, line 22: nit
```suggestion
    /// A radio packet has been received.
```

crates/board/src/radio.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl<B: crate::Api> From<Event> for crate::Event<B> {
+    fn from(event: Event) -> Self {
+        crate::Event::Radio(event)
+    }
+}
+
+/// Radio interface.
+pub trait Api: Support<bool> {
+    /// Enables radio events
```
ia0, line 34: nit
```suggestion
    /// Enables radio events.
```

crates/board/src/radio.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn enable() -> Result<(), Error>;
+
+    /// Disables radio events
```
ia0, line 37: nit
```suggestion
    /// Disables radio events.
```

crates/board/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Major
 
+- Add `radio` module
```
ia0, line 7: nit (to be adapted based on the new naming scheme)

```suggestion
- Add `Api::Radio` for radio APIs
- Add `radio::Api::Ble` for Bluetooth Low Energy
```

crates/runner-nordic/src/main.rs
```diff
@@ -77,6 +86,10 @@ struct State {
     timers: Timers,
     ccm: Ccm,
     leds: [Pin<Output<PushPull>>; <led::Impl as Support<usize>>::SUPPORT],
+    ble_radio: BleRadio,
+    ble_scanner: BeaconScanner<BleAdvScanCallback, AllowAll>,
+    ble_timer: BleTimer<TIMER0>,
+    ble_packet_queue: VecDeque<BlePacket>,
```
ia0, line 92: Could all those fields (and most of the rest of the radio code in this file) be moved to a single struct defined in the `radio` (or ultimately `radio::ble`) module? Ideally the main should only use high-level functions for setup.

crates/runner-nordic/src/main.rs
```diff
@@ -95,6 +109,9 @@ fn with_state<R>(f: impl FnOnce(&mut State) -> R) -> R {
 fn main() -> ! {
     static mut CLOCKS: MaybeUninit<Clocks> = MaybeUninit::uninit();
     static mut USB_BUS: MaybeUninit<UsbBusAllocator<Usb>> = MaybeUninit::uninit();
+    // TX buffer is mandatory even when we only listen
```
ia0, line 112: nit

```suggestion
    // TX buffer is mandatory even when we only listen.
```

crates/runner-nordic/src/main.rs
```diff
@@ -159,6 +200,80 @@ fn main() -> ! {
...
 
+pub struct RadioMetadata {
+    ticks: u32,
+    freq: u16,
+    rssi: i8,
+    pdu_type: u8,
+}
+
+impl RadioMetadata {
+    pub fn len(&self) -> usize {
```
ia0, line 211: If I understand correctly, those `len()` functions are only used for debugging right? If yes, then let's have them `#[cfg(feature = "debug")]`. Also, not sure they need to be `pub` given they're at the root.

crates/runner-nordic/src/radio.rs
```diff
@@ -0,0 +1,120 @@
...
+            channel: AdvertisingChannel::first(),
+        }
+    }
+
+    /// Configures the `BleTrackerScanner` and returns a `Cmd` to apply to the radio.
+    ///
+    /// The `next_update` field of the returned `Cmd` specifies when to call `timer_update` the next
+    /// time. The timer used for this needs to be pretty accurate because of advanced advertising packets,
+    /// which require to precisely switch channel to a data channel to capture the whole advertisement.
+    /// Otherwise it is only used to switch to the next advertising channel after `interval` elapses.
```
ia0, line 56: nit: Those comments are more than 100 columns and should be wrapped.

crates/runner-nordic/src/radio.rs
```diff
@@ -0,0 +1,120 @@
...
+        Cmd {
+            // Switch channels
+            next_update: NextUpdate::At(now + self.interval),
+
+            radio: RadioCmd::ListenAdvertising {
+                channel: self.channel,
+            },
+
+            queued_work: false,
+        }
```
ia0, line 70: nit (could keep the comment if useful, but end the sentence with a dot)

```suggestion
        Cmd {
            next_update: NextUpdate::At(now + self.interval),
            radio: RadioCmd::ListenAdvertising {
                channel: self.channel,
            },
            queued_work: false,
        }
```

crates/runner-nordic/src/radio.rs
```diff
@@ -0,0 +1,120 @@
...
```
ia0, line 88: same as above

```suggestion
        Cmd {
            next_update: NextUpdate::At(now + self.interval),
            radio: RadioCmd::ListenAdvertising {
                channel: self.channel,
            },
            queued_work: false,
        }
```

crates/runner-nordic/src/radio.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// This should be called whenever the radio receives a packet on the configured advertising
+    /// channel.
+    pub fn process_adv_packet(
+        &mut self,
+        header: Header,
+        payload: &[u8],
+        metadata: Metadata,
+    ) -> Cmd {
+        if metadata.crc_ok && header.type_().is_beacon() {
+            // Partially decode to get the device ID and run it through the filter
```
ia0, line 102: nit

```suggestion
            // Partially decode to get the device ID and run it through the filter.
```

crates/runner-nordic/src/radio.rs
ia0, line 1: It looks like this could be within the `src/tasks/radio.rs` (or ultimately `src/tasks/radio/ble.rs`) given it only exists to implement the API.

crates/runner-nordic/src/tasks/radio.rs
```diff
@@ -0,0 +1,68 @@
...
+            Ok(())
+        })
+    }
+
+    fn read(output: &mut [u8]) -> Result<usize, Error> {
+        with_state(|state| {
+            let packet = state.ble_packet_queue.pop_front().ok_or(Error::World)?;
+            let len = packet.len();
+            let mut writer = ByteWriter::new(output);
+            // Serialize RadioMetadata first, then BDADDR and append the packet
```
ia0, line 50: nit

```suggestion
            // Serialize RadioMetadata first, then BDADDR and append the packet.
```

crates/runner-nordic/src/tasks/radio.rs
```diff
@@ -0,0 +1,68 @@
...
+            state.ble_timer.configure_interrupt(NextUpdate::Disable);
+            state.ble_radio.configure_receiver(RadioCmd::Off);
+            Ok(())
+        })
+    }
+
+    fn read(output: &mut [u8]) -> Result<usize, Error> {
+        with_state(|state| {
+            let packet = state.ble_packet_queue.pop_front().ok_or(Error::World)?;
+            let len = packet.len();
```
ia0, line 48: I guess this is where `len()` is used and it seems it's only for debugging. I'll have to think a bit about what's the best option. For now, let's keep it like that.

crates/runner-nordic/src/tasks/radio.rs
```diff
@@ -0,0 +1,68 @@
...
+            res.map_err(|_| Error::World)?;
+            logger::trace!("{}{:?} = read({})", len, &output[.. len], output.len());
+            Ok(len)
+        })
+    }
+}
+
+impl Support<bool> for Impl {
+    const SUPPORT: bool = true;
+}
```
ia0, line 68: nit

```suggestion
impl Supported for Impl {}
```

crates/scheduler/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Major
 
+- Add `radio` module
```
ia0, line 7: This is not major since it doesn't change the API of the scheduler. It's a minor change:

```
- Support `radio::ble`
```

ia0 (commented):
Thanks! I did a first pass. I think the only issue right now is that the API should probably be nested to indicate what kind of radio it is. Otherwise it will block the generic term radio for this specific use. Not sure what's the best name to describe this new API, I guess `radio::ble`, but maybe it should be `radio::ble_advertisement` or `radio::ble_adv` for short.

ia0 (approved):

PR #327, opened 2023-12-15
Add impl for WasefireWrite

kofls:


examples/rust/opensk/Cargo.toml
```diff
@@ -11,6 +11,7 @@ persistent_store = { path = "../../../third_party/OpenSK/libraries/persistent_st
 rand_core = "0.6.4"
 wasefire = { path = "../../../crates/prelude" }
 wasefire-stub = { path = "../../../crates/stub", optional = true }
+wasefire-logger = { path = "../../../crates/logger" }
```
ia0, line 14: This crate is only for platform implementation. For applets, you have the `debug!()` macro.

examples/rust/opensk/src/write.rs
```diff
@@ -12,11 +12,23 @@
...
-    fn write_str(&mut self, _: &str) -> core::fmt::Result {
-        todo!()
+    fn write_str(&mut self, data: &str) -> core::fmt::Result {
+        self.buf.write_str(data)
+    }
+}
+
+impl Drop for WasefireWrite {
+    fn drop(&mut self) {
+        log::debug!(self.buf)
```
ia0, line 32: This should be `wasefire::debug!("{}", self.buf)`.

examples/rust/opensk/Cargo.toml
kofls: Removed this.

examples/rust/opensk/src/write.rs
kofls: Done.

ia0 (approved):

ia0:
cc @kaczmarczyck

PR #330, opened 2023-12-15
Add impl for WasefireClock

kofls:


examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
...
+
+impl Handler for ClockHandler {
+    fn event(&self) {
+        // TODO: What should we do here?
+        wasefire::debug!("timer elapsed");
+    }
+}
+
+pub struct WasefireTimer {
+    timer: Timer<ClockHandler>,
```
ia0, line 30: We need a field to know whether the timer as triggered.

```suggestion
    timer: Timer<ClockHandler>,
    elapsed: Rc<Cell<bool>>,
```

crates/prelude/src/clock.rs
```diff
@@ -101,6 +101,14 @@ impl<H: Handler> Timer<H> {
         core::mem::forget(self);
     }
 
+    /// Returns true if the timer has elapsed.
+    pub fn is_elapsed(&self) -> bool {
```
ia0, line 105: This function is not needed. It can be implemented by the user.

examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
...
+}
+
+impl Default for WasefireTimer {
```
ia0, line 33: Let's not implement `Default`. We can only create a timer with `make_timer`.

examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
 // limitations under the License.use opensk_lib::api::clock::Clock;
 
 use opensk_lib::api::clock::Clock;
+use wasefire::clock::Mode::Oneshot;
+use wasefire::clock::{Handler, Timer};
 
 #[derive(Default)]
-pub struct WasefireClock {}
+struct ClockHandler;
```
ia0, line 20:
```suggestion
struct ClockHandler {
    triggered: Rc<Cell<bool>>,
}
```

examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
...
+
+impl Handler for ClockHandler {
+    fn event(&self) {
+        // TODO: What should we do here?
+        wasefire::debug!("timer elapsed");
```
ia0, line 25:
```suggestion
        self.triggered.set(true);
```

examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
...
 
 impl Clock for WasefireClock {
-    type Timer = Self;
+    type Timer = WasefireTimer;
 
     fn make_timer(&mut self, milliseconds: usize) -> Self::Timer {
-        todo!()
+        let timer = Timer::new(ClockHandler {});
+        timer.start_ms(Oneshot, milliseconds);
+        WasefireTimer { timer }
```
ia0, line 49:
```suggestion
        let elapsed = Rc::new(Cell::new(false));
        let triggered = elapsed.clone();
        let timer = Timer::new(ClockHandler { triggered });
        timer.start_ms(Oneshot, milliseconds);
        WasefireTimer { timer, elapsed }
```

examples/rust/opensk/src/clock.rs
```diff
@@ -13,18 +13,43 @@
...
     }
 
     fn is_elapsed(&mut self, timer: &Self::Timer) -> bool {
-        todo!()
+        timer.timer.is_elapsed()
```
ia0, line 53:
```suggestion
        timer.elapsed.get()
```

examples/rust/opensk/src/clock.rs
kofls: This is required from the `Clock` trait: https://github.com/google/OpenSK/blob/5fdc6e0739b1a9e758d99fa0dc268282e7bc8aae/libraries/opensk/src/api/clock.rs#L22

crates/prelude/src/clock.rs
kofls: Removed this.

examples/rust/opensk/src/clock.rs
ia0: Oh thanks, I wasn't aware of that.

examples/rust/opensk/src/clock.rs
```diff
@@ -12,19 +12,53 @@
...
+}
+
+pub struct WasefireTimer {
+    timer: Timer<ClockHandler>,
+    elapsed: Rc<Cell<bool>>,
+}
+
+impl Default for WasefireTimer {
+    fn default() -> Self {
+        let elapsed = Rc::new(Cell::new(false));
```
ia0, line 40: According to the link you sent, the default value should be a timer that has elapsed.

```suggestion
        let elapsed = Rc::new(Cell::new(true));
```

examples/rust/opensk/src/clock.rs
```diff
@@ -12,19 +12,53 @@
...
+        let triggered = elapsed.clone();
+        let timer = Timer::new(ClockHandler { triggered });
```
ia0, line 42: I'm wondering if we shouldn't try to optimize this case. Let's maybe just add a note.

```suggestion
        // This is a bit wasteful to allocate a timer that we don't need. This could be optimized later.
        let timer = Timer::new(ClockHandler { triggered });
```

ia0 (approved):

ia0:
cc @kaczmarczyck

PR #349, opened 2023-12-30
Add impl for UserPresence

kofls:
cc: @kaczmarczyck

### DIFF
--- a/examples/rust/opensk/src/user_presence.rs
+++ b/examples/rust/opensk/src/user_presence.rs
@@ -29,36 +29,25 @@ impl UserPresence for WasefireEnv {
 
     fn wait_with_timeout(&mut self, timeout_ms: usize) -> UserPresenceResult {
         // Make sure there is at least one button and one LED.
-        let num_buttons = button::count();
-        let num_leds = led::count();
-        if num_buttons == 0 || num_leds == 0 {
+        if button::count() == 0 || led::count() == 0 {
             return Err(Fail);
         }
 
         // Start listening for button presses.
-        // TODO: Should we be checking for any button pressed or only index 0?
         let button_pressed = Rc::new(Cell::new(false));
-        for index in 0 .. num_buttons {
+        let listener = Listener::new(0, {
             let button_pressed = button_pressed.clone();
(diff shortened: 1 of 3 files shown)


examples/rust/opensk/src/user_presence.rs
```diff
@@ -12,22 +12,75 @@
...
+                button::Pressed => button_pressed.set(true),
+                button::Released => (),
+            };
+            let listener = Listener::new(index, handler);
+            listener.leak();
+        }
+
+        // Start a periodic timer blinking the LED.
+        // TODO: Similarly, should we be blinking all LEDs or only index 0?
+        let blinking = Rc::new(Cell::new(false));
```
ia0, line 53: We don't need this variable (and you probably meant to initialize it to true). It is enough to do:

```rust
let blink_timer = clock::Timer::new(|| led::set(0, !led::get(0)));
led::set(0, led::On);
blink_timer.start(Periodic, Duration::from_millis(200));
// [...]
scheduling::wait_until(|| button_pressed.get() || timeout.get());
drop(blink_timer);
led::set(0, led::Off);
```

ia0 (commented):
Looks good modulo the 2 comments.

examples/rust/opensk/src/user_presence.rs
```diff
@@ -12,22 +12,75 @@
...
+        // TODO: Should we be checking for any button pressed or only index 0?
+        let button_pressed = Rc::new(Cell::new(false));
+        for index in 0 .. num_buttons {
+            let button_pressed = button_pressed.clone();
+            let handler = move |state| match state {
+                button::Pressed => button_pressed.set(true),
+                button::Released => (),
+            };
+            let listener = Listener::new(index, handler);
+            listener.leak();
```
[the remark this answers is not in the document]
kofls: Ok, got it. Removed `leak()`

examples/rust/opensk/src/user_presence.rs
kofls: Done.

ia0 (approved):

Issue #50, opened 2023-03-23
Platform and APIs slicing


ia0:
It should be possible to select at compilation time, which parts of the board and applet APIs should be supported. This would reduce the footprint of the platform and possibly the applets too when the same selection is done in the prelude (unless relying on dead-code elimination is enough).

Ideally this selection should be done at a unique place for all of the board API, the applet API, and the preludes. Even better is if this selection may be generated given a set of applets (selecting the minimum set of covering features), see #49.

Note that if this is not feasible, then the platform should always accept unknown API and just trap, otherwise we lose compatibility (a platform not aware of APIs linked in the applet but not used).

### Open questions

- For applet API, we need a way to handle unlinked functions polymorphically. One solution would be that all applet API functions return `isize` (or `i32`) such with negative values for errors. This way it's possible to return `World::NotImplemented` for unlinked functions. Created #355 for this.

### Using features versus `cfg`s

Pro features:
- Optional dependencies may be enabled only if some features are enabled.
- Works nice with `rust-analyzer` because it doesn't rely on `build.rs` and a correct environment.

Pro `cfg`s:
- Much less boilerplate in `Cargo.toml`. Single source of truth.

PR #351, opened 2024-01-04
Mention the board implementation in all events

ia0:
This is useful for #50 where only a subset of events may be present, because if none of them mentions the board implementation, then the compiler cannot guess the board implementation.


ia0:
This actually doesn't work. If there are no events we have the same problem.

PR #354, opened 2024-01-05
Add applet and board API features

### LINKED ISSUE #50: Platform and APIs slicing

ia0:
Fixes #50 although before releasing, it would be nice to remove `multivalue` feature and make all functions return `isize`, such that it's easy to provide a default for missing functions by returning `World::NotImplemented` (see #355).


ia0:
What's the footprint impact?

PR #363, opened 2024-01-08
Make all applet API functions return isize

### LINKED ISSUE #355: Make return type of all applet API functions uniform
This is needed for compatibility between platform and applets across versions. By always returning `isize`, the platform may return `World:NotImplemented` for function the applet wants to import that don't exist.

ia0:
Fixes #355

### DIFF
--- a/crates/api-desc/src/scheduling.rs
+++ b/crates/api-desc/src/scheduling.rs
@@ -22,20 +22,19 @@ pub(crate) fn new() -> Item {
             /// Waits until a callback is scheduled.
             ///
             /// This can be used as power management, since the CPU will sleep while waiting.
-            fn wait_for_callback "sw" {} -> {}
+            ///
+            /// Returns zero on success.
+            fn wait_for_callback "sw" {}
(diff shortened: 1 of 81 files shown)


crates/api-desc/src/scheduling.rs
```diff
@@ -22,20 +22,19 @@ pub(crate) fn new() -> Item {
...
-            }
+            fn num_pending_callbacks "sh" {}
         },
         item! {
             /// Aborts the applet.
             ///
-            /// This call doesn't return.
-            fn abort "sa" {} -> {}
+            /// Does not return on success.
+            fn abort "sa" {}
```
ia0, line 37: It might be worth trying to statically ensure some properties on the resulting type:
- `-> !` only returns negative values
- `-> ()` does not return positive values
- `-> bool` only returns values less than 2
- `-> usize` returns all possible `isize`

Besides making `api-desc` slightly more readable, this could add some additional type-checking in the scheduler.

PR #368, opened 2024-01-10
Rename clock to timer in applet API

ia0:


ia0:
Also run `./scripts/hwci.sh nordic --features=software-crypto` successfully.

PR #401, opened 2024-02-23
Correctly remove UART control flow for Nordic

### LINKED ISSUE #392: UART implementation is resulting in data loss and data corruption
Using UART without hardware control flow, could result in data loss or data corruption.
This problem might not be easy to reproduce without using an external UART.
Connecting a logic analyzer confirmed that the UART data is correct at the Nordic GPIO level.

The bug shouldn't depend on the baudrate because the root cause is on the FIFO handling.
Current RX buffer is 32 bytes and increasing it to 2048 solved the issue.

Here is my analysis:
* External device sends more data than what the RX buffer can hold
* the Nordic triggers an interrupt with an ENDRX event.
* Wasefire then stops receiving while transferring data out of the RX buffer and issues a STARTRX at the end to resume receiving data
* in the meantime the external device continued transmitting data which is now lost and when the STARTRX task starts, the Nordic will receive in the middle of an ongoing transmission, resulting in a framing error (one bit belonging to user data will be mistaken as a start bit)

Expected behavior:
* in continuous RX, Wasefire should use the SHORTS to quickly trigger a STARTRX task when an ENDRX event occurs
* the interrupt handler should update the RXD.PTR to avoid overwritting previously received data -- that requires double buffering
* When asked to stop receiving, the SHORTS should be cleared to stay in ENDRX
* the platform should log (warning or error) the ERRORSRC register to the applet, if at least one of the first three is occurring (PARITY, FRAMING and OVERRUN) to help catch such issue in the future.

ia0:
UART control flow was not correctly removed in #391 for Nordic, because reception may start in the middle of a frame resulting in corrupted data. By always keeping the UARTE in reception, we avoid corruption. Data may still be lost when the buffer is full but that's by design of not using control flow.

This implementation does not provide access to the most recent data and is only meant for streaming scenario when data is constantly received. See https://github.com/google/wasefire/pull/401/commits/1e47e9cf3a6b6bd231ba7240b915773d917390f2 for the commit that removes the logic to access most recent data. It didn't work because STOPRX doesn't seem to work in the presence of a short from ENDRX to STARTRX.

Fixes #392


ia0:
@jmichelp Can you test if this is fixing your issue? In my test environment I still see issues.

PR #428, opened 2024-04-15
Remove unecessary calls to call.read()

lukeyeh:
I believe these calls do not require any args and that having them around adds some minor confusion.


ia0:
It's true those applet API functions do not have any parameters at the moment (the API is not stable yet) and those argument unwrapping calls are unnecessary for the purpose of code behavior. However they serve 2 other purposes:
- Type-checked documentation: One can locally know which parameters an applet API function task (without having to jump to the api-desc crate).
- Maintainability: When changing the applet API (e.g. by adding, removing, renaming a parameter), one only needs to follow the compiler errors to update the code, ensuring that nothing is forgotten.

So I would prefer to keep those calls and more generally keep the applet API as statically typed as possible for now.

lukeyeh:
Makes sense, thank you for the thorough explanation!

PR #429, opened 2024-04-17
Add LastError trait to check for error after using RustCrypto API

### LINKED ISSUE #176: Add error support when RustCrypto doesn't
After #174 the hash API doesn't support returning errors because the RustCrypto API for digest (e.g. `Update`, `FixedOutput`) don't return a `Result`. We should somehow add the possibility to report errors in those operations, for example using a side channel that may be checked. This could be a `LastError` trait that would be used after each operation to check for error.

```rust
trait LastError {
    type Error: Copy;
    fn last_error(&self) -> Option<Error>;
}
```

zhouwfang:
To resolve #176.

Please let me know any issue or improvement. Thanks!

1) I was wondering if there is any existing test in `wasefire/examples` that could be used to check if these changes break anything, especially using `FixedOutputReset` for the `Hmac` trait.

2) What Rust format tool do you use? As you can see, I use a different format tool and I'm happy to change to yours.

### DIFF
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -203,14 +203,14 @@ pub trait LastError {
 #[cfg(feature = "software-crypto-sha256")]
 impl LastError for sha2::Sha256 {
     fn last_error(&self) -> Result<(), Error> {
-        todo!()
+        Ok(())
     }
 }
 
 #[cfg(feature = "software-crypto-sha384")]
 impl LastError for sha2::Sha384 {
     fn last_error(&self) -> Result<(), Error> {
-        todo!()
+        Ok(())
     }
 }
 
@@ -226,7 +226,7 @@ impl<
     > LastError for hmac::SimpleHmac<D>
 {
     fn last_error(&self) -> Result<(), Error> {
-        todo!()
+        Ok(())
     }
 }
 
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -200,6 +200,36 @@ pub trait LastError {
     fn last_error(&self) -> Result<(), Error>;
 }
 
+#[cfg(feature = "software-crypto-sha256")]
+impl LastError for sha2::Sha256 {
+    fn last_error(&self) -> Result<(), Error> {
+        todo!()
+    }
+}
+
+#[cfg(feature = "software-crypto-sha384")]
+impl LastError for sha2::Sha384 {
+    fn last_error(&self) -> Result<(), Error> {
+        todo!()
+    }
+}
+
+#[cfg(feature = "internal-software-crypto-hmac")]
+impl<
+        D: Support<bool>
+            + Default
+            + BlockSizeUser
+            + Update
+            + FixedOutputReset
+            + HashMarker
+            + LastError,
+    > LastError for hmac::SimpleHmac<D>
+{
+    fn last_error(&self) -> Result<(), Error> {
+        todo!()
+    }
+}
+
 /// Wrapper API around Hash that calls last_error.
 #[cfg(feature = "internal-api-crypto-hash")]
 pub struct HashApi<T: Hash>(T);
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -16,7 +16,6 @@
 
 #[cfg(feature = "internal-api-crypto-hash")]
 use crypto_common::BlockSizeUser;
-use crypto_common::Key;
 #[cfg(feature = "internal-api-crypto-hmac")]
 use crypto_common::KeyInit;
 #[cfg(any(feature = "internal-api-crypto-hash", feature = "internal-api-crypto-hmac"))]
@@ -197,7 +196,7 @@ impl<T: Hash> HashApi<T> {
         let hash = T::default();
         let error = T::last_error(&hash);
         match error {
-            Ok(()) => hash,
+            Ok(()) => Ok(hash),
             Err(error) => Err(error),
         }
     }
@@ -205,10 +204,10 @@ impl<T: Hash> HashApi<T> {
 
 impl<T: Hmac> HmacApi<T> {
     pub fn new(key: &[u8]) -> Result<Self, Error> {
-        let hash = T::new_from_slice(key)?;
+        let hash = T::new_from_slice(key).map_err(|_| Error)?;
         let error = T::last_error(&hash);
         match error {
-            Ok(()) => hash,
+            Ok(()) => Ok(hash),
             Err(error) => Err(error),
         }
     }
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -191,41 +191,29 @@ pub trait LastError {
     fn last_error(&self) -> Result<(), Error>;
 }
 
-// Wrapper API around Hash that calls last_error().
+/// Wrapper API around Hash that calls last_error.
 #[cfg(feature = "internal-api-crypto-hash")]
-#[allow(missing_docs)]
 pub struct HashApi<T: Hash>(T);
 
-// Wrapper API around Hmac that calls last_error().
+/// Wrapper API around Hmac that calls last_error.
 #[cfg(feature = "internal-api-crypto-hmac")]
-#[allow(missing_docs)]
 pub struct HmacApi<T: Hmac>(T);
 
-#[allow(missing_docs)]
 #[cfg(feature = "internal-api-crypto-hash")]
 impl<T: Hash> HashApi<T> {
-    // Constructor of HashApi
+    /// Constructor of HmacApi
     pub fn new() -> Result<Self, Error> {
         let hash = T::default();
-        let error = T::last_error(&hash);
-        match error {
-            Ok(()) => Ok(HashApi(hash)),
-            Err(error) => Err(error),
-        }
+        hash.last_error().map(|()| HashApi(hash))
     }
 }
 
-#[allow(missing_docs)]
 #[cfg(feature = "internal-api-crypto-hmac")]
 impl<T: Hmac> HmacApi<T> {
-    // Constructor of HmacApi
+    /// Constructor of HmacApi
     pub fn new(key: &[u8]) -> Result<Self, Error> {
         let hash =
             T::new_from_slice(key).map_err(|_| Error::new(Space::User, Code::InvalidLength))?;
-        let error = T::last_error(&hash);
-        match error {
-            Ok(()) => Ok(HmacApi(hash)),
-            Err(error) => Err(error),
-        }
+        hash.last_error().map(|()| HmacApi(hash))
     }
 }
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -210,19 +210,19 @@ pub struct HmacApi<T: Hmac>(T);
 
 #[cfg(feature = "internal-api-crypto-hash")]
 impl<T: Hash> HashApi<T> {
-    /// Create a hash wrapper.
+    /// Creates a hash wrapper.
     pub fn new() -> Result<Self, Error> {
         let hash = T::default();
         hash.last_error().map(|()| HashApi(hash))
     }
 
-    /// Update the hash with the provided data.
+    /// Updates the hash with the provided data.
     pub fn update(&mut self, data: &[u8]) -> Result<(), Error> {
         self.0.update(data);
         self.0.last_error()
     }
 
-    /// Finalize the hash to the provided output.
+    /// Finalizes the hash to the provided output.
     pub fn finalize_into(mut self, out: &mut Output<T>) -> Result<(), Error> {
         self.0.finalize_into_reset(out);
         self.0.last_error()
@@ -231,7 +231,7 @@ impl<T: Hash> HashApi<T> {
 
 #[cfg(feature = "internal-api-crypto-hmac")]
 impl<T: Hmac> HmacApi<T> {
-    /// Create a HMAC wrapper.
+    /// Creates a HMAC wrapper.
     pub fn new(key: &[u8]) -> Result<Self, Error> {
         let hash = T::new_from_slice(key).map_err(|_| Error::user(Code::InvalidLength))?;
         hash.last_error().map(|()| HmacApi(hash))
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -70,41 +70,6 @@ pub trait Api: Send {
     type Sha384: Hash<BlockSize = typenum::U128, OutputSize = typenum::U48>;
 }
 
-/// Extends APIs with infallible operations to support errors.
-///
-/// Infallible operations may return incorrect values. Those results must not be used until
-/// `last_error()` is called and returns `Ok(())`, signifying that the return value is correct.
-pub trait LastError {
-    /// Returns the last error of infallible operations.
-    fn last_error(&self) -> Result<(), Error>;
-}
-
-// Wrapper API around Hash/Hmac that calls last_error().
-pub struct HashApi<T: Hash>(T);
-pub struct HmacApi<T: Hmac>(T);
-
-impl<T: Hash> HashApi<T> {
-    pub fn new() -> Result<Self, Error> {
-        let hash = T::default();
-        let error = T::last_error(&hash);
-        match error {
-            Ok(()) => hash,
-            Err(error) => Err(error),
-        }
-    }
-}
-
-impl<T: Hmac> HmacApi<T> {
-    pub fn new(key: &[u8]) -> Result<Self, Error> {
-        let hash = T::new_from_slice(key).map_err(Error)?;
-        let error = T::last_error(&hash);
-        match error {
-            Ok(()) => hash,
-            Err(error) => Err(error),
-        }
-    }
-}
-
 /// Hash interface.
 #[cfg(feature = "internal-api-crypto-hash")]
 pub trait Hash:
@@ -212,4 +177,39 @@ impl<D: Support<bool> + Default + BlockSizeUser + Update + FixedOutput + HashMar
     for hmac::SimpleHmac<D>
 {
     const SUPPORT: bool = D::SUPPORT;
+}
+
+/// Extends APIs with infallible operations to support errors.
+///
+/// Infallible operations may return incorrect values. Those results must not be used until
+/// `last_error()` is called and returns `Ok(())`, signifying that the return value is correct.
+pub trait LastError {
+    /// Returns the last error of infallible operations.
+    fn last_error(&self) -> Result<(), Error>;
+}
+
+// Wrapper API around Hash/Hmac that calls last_error().
+pub struct HashApi<T: Hash>(T);
+pub struct HmacApi<T: Hmac>(T);
+
+impl<T: Hash> HashApi<T> {
+    pub fn new() -> Result<Self, Error> {
+        let hash = T::default();
+        let error = T::last_error(&hash);
+        match error {
+            Ok(()) => hash,
+            Err(error) => Err(error),
+        }
+    }
+}
+
+impl<T: Hmac> HmacApi<T> {
+    pub fn new(key: &[u8]) -> Result<Self, Error> {
+        let hash = T::new_from_slice(key)?;
+        let error = T::last_error(&hash);
+        match error {
+            Ok(()) => hash,
+            Err(error) => Err(error),
+        }
+    }
 }
--- a/crates/board/src/crypto.rs
+++ b/crates/board/src/crypto.rs
@@ -19,7 +19,7 @@ use crypto_common::BlockSizeUser;
 #[cfg(feature = "internal-api-crypto-hmac")]
 use crypto_common::KeyInit;
 #[cfg(any(feature = "internal-api-crypto-hash", feature = "internal-api-crypto-hmac"))]
-use crypto_common::{Output, OutputSizeUser};
+use crypto_common::Output;
 #[cfg(feature = "internal-api-crypto-hash")]
 use digest::HashMarker;
 #[cfg(feature = "internal-api-crypto-hmac")]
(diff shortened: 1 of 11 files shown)


crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -63,9 +64,17 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
     let result = try {
         let context = match convert_hash_algorithm::<B>(*algorithm)?? {
             #[cfg(feature = "board-api-crypto-sha256")]
-            Algorithm::Sha256 => HashContext::Sha256(board::crypto::Sha256::<B>::default()),
+            Algorithm::Sha256 => {
+                let default = board::crypto::Sha256::<B>::default();
+                default.last_error()?;
+                HashContext::Sha256(default)
+            }
```
zhouwfang, line 71: @ia0 You mentioned we could use `Result::map` for this. Could you explain more? Thanks.

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -63,9 +64,17 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
     let result = try {
         let context = match convert_hash_algorithm::<B>(*algorithm)?? {
             #[cfg(feature = "board-api-crypto-sha256")]
-            Algorithm::Sha256 => HashContext::Sha256(board::crypto::Sha256::<B>::default()),
+            Algorithm::Sha256 => {
+                let default = board::crypto::Sha256::<B>::default();
+                default.last_error()?;
```
zhouwfang, line 69: @ia0 You mentioned we could use `Result::map` for this. Could you explain more? Thanks.

ia0: The `let context =` needs to become `let (context, error) =` and each branch needs to return both, e.g.:

```rust
let context = board::crypto::Sha256::<B>::default();
let error = context.last_error();
(context, error)
```

Then the final `Ok(scheduler.applet.hashes.insert(context)? as u32)` _could_ become:

```rust
error.map(|()| scheduler.applet.hashes.insert(context)? as u32)
```

The problem is that we capture the `?` in the closure, while it should return to the try-block. So this actually doesn't work. We need to write the map manually instead:

```rust
match error {
    Ok(()) => Ok(scheduler.applet.hashes.insert(context)? as u32),
    Err(error) => Err(error),
}
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -83,9 +92,15 @@ fn update<B: Board>(mut call: SchedulerCall<B, api::update::Sig>) {
...
+            }
             #[cfg(feature = "board-api-crypto-sha384")]
-            HashContext::Sha384(context) => context.update(data),
+            HashContext::Sha384(context) => {
+                context.update(data);
+                context.last_error()?
+            }
             _ => trap_use!(data),
         }
         Ok(())
```
ia0, line 106: This must now be removed. The `last_error()` is the result.

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -83,9 +92,15 @@ fn update<B: Board>(mut call: SchedulerCall<B, api::update::Sig>) {
...
-            HashContext::Sha256(context) => context.update(data),
+            HashContext::Sha256(context) => {
+                context.update(data);
+                context.last_error()?
+            }
             #[cfg(feature = "board-api-crypto-sha384")]
-            HashContext::Sha384(context) => context.update(data),
+            HashContext::Sha384(context) => {
+                context.update(data);
+                context.last_error()?
```
ia0, line 102: We don't want to "unwrap" the error. Within the try-block, errors are only for traps. Here, we want to return the error to the user.
```suggestion
                context.last_error()
```

ia0 (commented):
Thanks for the PR!

1. You can just run `./scripts/ci.sh` to run all CI locally. However, if you don't want to modify the changelogs just yet, you can run `./scripts/ci-tests.sh` which runs the tests of all crates, and `./scripts/hwci.sh host --no-default-features` which runs all applet tests. This last one will exerce those APIs (in particular the `hash_test` and `ec_test` for example).

Actually, thinking about it, I think a better approach that would statically avoid mistakes, is to define a wrapper API on top of board::crypto::Api that actually return errors such that last_error is handled internally. It would look like this:

```rust
pub struct HashApi<T: Hash>(T);
pub struct HmacApi<T: Hmac>(T);

impl<T: Hash> HashApi<T> {
    pub fn new() -> Result<Self, Error> { ... last_error ... }
    pub fn update(&mut self, data: &[u8]) -> Result<(), Error> { ... last_error ... }
    ...
}
// same for Hmac
```

2. I use rustfmt (through rust-analyzer, but this shouldn't change anything) which should use the top-level rustfmt.toml configuration file. `cargo fmt` should also work the same.

crates/board/src/crypto.rs
```diff
@@ -158,8 +175,57 @@ impl crate::Supported for sha2::Sha256 {}
...
+    /// Returns the last error of infallible operations.
+    fn last_error(&self) -> Result<(), Error>;
+}
+
+// Wrapper API around Hash that calls last_error().
+#[cfg(feature = "internal-api-crypto-hash")]
+#[allow(missing_docs)]
+pub struct HashApi<T: Hash>(T);
+
+// Wrapper API around Hmac that calls last_error().
```
ia0, line 199: nit: those could be proper documentation and not just comments
```suggestion
/// Wrapper API around Hmac that calls last_error.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +175,57 @@ impl crate::Supported for sha2::Sha256 {}
...
+#[allow(missing_docs)]
+pub struct HmacApi<T: Hmac>(T);
+
+#[allow(missing_docs)]
+#[cfg(feature = "internal-api-crypto-hash")]
+impl<T: Hash> HashApi<T> {
+    // Constructor of HashApi
+    pub fn new() -> Result<Self, Error> {
+        let hash = T::default();
+        let error = T::last_error(&hash);
```
ia0, line 210: nit: it should be possible to call last_error as a method
```suggestion
        let error = hash.last_error();
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +175,57 @@ impl crate::Supported for sha2::Sha256 {}
...
+        match error {
+            Ok(()) => Ok(HashApi(hash)),
+            Err(error) => Err(error),
+        }
```
ia0, line 214: nit: this would be an example where Result::map should work
```suggestion
        hash.last_error().map(|()| HashApi(hash))
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -61,15 +65,30 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
     let api::initialize::Params { algorithm } = call.read();
     let scheduler = call.scheduler();
     let result = try {
-        let context = match convert_hash_algorithm::<B>(*algorithm)?? {
+        let context_or_error = match convert_hash_algorithm::<B>(*algorithm)?? {
```
ia0, line 68: nit: it's ok to call a `Result<T, Error>` the same as a `T`, otherwise the codebase will be full of `or_error`
```suggestion
        let context = match convert_hash_algorithm::<B>(*algorithm)?? {
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -61,15 +65,30 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
...
             #[cfg(feature = "board-api-crypto-sha256")]
-            Algorithm::Sha256 => HashContext::Sha256(board::crypto::Sha256::<B>::default()),
+            Algorithm::Sha256 => {
+                let hash_or_error = HashApi::<board::crypto::Sha256<B>>::new();
+                match hash_or_error {
+                    Ok(hash) => Ok(HashContext::Sha256(hash)),
+                    Err(error) => Err(error),
+                }
```
ia0, line 75: I didn't think of returning the result directly, this is actually much better.

nit: you can actually improve the style a bit here
```suggestion
                HashApi::<board::crypto::Sha256<B>>::new().map(HashContext::Sha256)
```

ia0 (commented):
Thanks for the work! This is going in the right direction from what I can see. It seems you still need to do the same as `new` but for `update` and `finalize`.

crates/scheduler/src/call/crypto/hash.rs
zhouwfang: Nice, and I had to make a modification -- `HashContext::Sha256()` does not take `HashApi` but the hash in it.

zhouwfang:
> Actually, thinking about it, I think a better approach that would statically avoid mistakes, is to define a wrapper API on top of board::crypto::Api that actually return errors such that last_error is handled internally. It would look like this:
> 
> ```rust
> pub struct HashApi<T: Hash>(T);
> pub struct HmacApi<T: Hmac>(T);
> 
> impl<T: Hash> HashApi<T> {
>     pub fn new() -> Result<Self, Error> { ... last_error ... }
>     pub fn update(&mut self, data: &[u8]) -> Result<(), Error> { ... last_error ... }
>     ...
> }
> // same for Hmac
> ```

Good idea! (I also had a similar feeling -- need something similar to RAII in C++.)

zhouwfang:
> Thanks for the work! This is going in the right direction from what I can see. It seems you still need to do the same as `new` but for `update` and `finalize`.

The current error is that `LastError` is not implemented for `hmac::simple::SimpleHmac` (and `CoreWrapper`). Should I add a trivial implementation (no error)?

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+        self.0.update(data);
+        self.0.last_error()
+    }
+
+    /// Call finalize_into_reset from RustCrypto API and then last_error.
+    pub fn finalize_into_reset(&mut self, out: &mut Output<Self>) -> Result<(), Error> {
+        self.0.finalize_into_reset(out);
+        self.0.last_error()
+    }
+}
```
zhouwfang, line 233: to me: think of a way to factor out these common code for both structs

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -156,14 +169,17 @@ fn hmac_update<B: Board>(mut call: SchedulerCall<B, api::hmac_update::Sig>) {
...
-            HashContext::HmacSha256(context) => context.update(data),
+            HashContext::HmacSha256(hmac) => HmacApi(hmac).update(data),
             #[cfg(feature = "board-api-crypto-hmac-sha384")]
-            HashContext::HmacSha384(context) => context.update(data),
+            HashContext::HmacSha384(hmac) => HmacApi(hmac).update(data),
             _ => trap_use!(data),
+        };
+        match context {
+            Ok(()) => Ok(()),
+            Err(error) => Err(error),
```
zhouwfang, line 181: @ia0 Without these lines, the compiler asked me to explicitly write out the type of result as it couldn't figure it out. There is probably a more elegant way.

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn update(&mut self, data: &[u8]) -> Result<(), Error> {
+        self.0.update(data);
+        self.0.last_error()
+    }
+
+    /// Call finalize_into_reset from RustCrypto API and then last_error.
+    pub fn finalize_into_reset(&mut self, out: &mut Output<Self>) -> Result<(), Error> {
+        self.0.finalize_into_reset(out);
+        self.0.last_error()
+    }
```
zhouwfang, line 254: The problem with using `finalize_into` is that `self` would be moved into it so that we could not call `last_error` on it afterwards.

But I'm entirely sure `finalize_into_reset` should be used because it does seem to make sense to call `last_error` after `self.0` is reset to default.

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+
+#[cfg(feature = "internal-api-crypto-hash")]
+impl<T: Hash> OutputSizeUser for HashApi<T> {
+    type OutputSize = T::OutputSize;
+}
+
+#[cfg(feature = "internal-api-crypto-hmac")]
+impl<T: Hmac> OutputSizeUser for HmacApi<T> {
+    type OutputSize = T::OutputSize;
+}
```
ia0, line 212: Are those really needed? It should be possible to use `OutputSize` directly on their type parameter.

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+
+#[cfg(feature = "internal-api-crypto-hash")]
+impl<T: Hash> HashApi<T> {
+    /// Constructor of HmacApi.
```
ia0, line 216:
```suggestion
    /// Creates a hash wrapper.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn new() -> Result<Self, Error> {
+        let hash = T::default();
+        hash.last_error().map(|()| HashApi(hash))
+    }
+
+    /// Call update from RustCrypto API and then last_error.
```
ia0, line 222: nit: I think the type documentation is enough context regarding last_error, the function type also reminds it. We can just document the high-level operation.
```suggestion
    /// Updates the hash with the provided data.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn update(&mut self, data: &[u8]) -> Result<(), Error> {
+        self.0.update(data);
+        self.0.last_error()
+    }
+
+    /// Call finalize_into_reset from RustCrypto API and then last_error.
```
ia0, line 228:
```suggestion
    /// Finalizes the hash to the provided output.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn finalize_into_reset(&mut self, out: &mut Output<Self>) -> Result<(), Error> {
```
ia0, line 229: Let's forbid usage of self after this call (like it used to be the case). The fact that we use `into_reset` is an implementation detail of this type.

Also, I believe using `Output<T>` here should work and prevent implementing `OutputSizeUser` on the wrapper types.

```suggestion
    pub fn finalize_into(self, out: &mut Output<T>) -> Result<(), Error> {
```

crates/board/src/crypto.rs
ia0: I agree it looks like duplicated code a bit. But given we only have 2 instances (hash and hmac) and don't expect to get more, this duplication is actually fine (and the gain in readability by having the duplication is worth the duplication).

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+        self.0.finalize_into_reset(out);
+        self.0.last_error()
+    }
+}
+
+#[cfg(feature = "internal-api-crypto-hmac")]
+impl<T: Hmac> HmacApi<T> {
+    /// Constructor of HmacApi.
```
ia0, line 237:
```suggestion
    /// Creates an HMAC wrapper.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn new(key: &[u8]) -> Result<Self, Error> {
+        let hash =
+            T::new_from_slice(key).map_err(|_| Error::new(Space::User, Code::InvalidLength))?;
```
ia0, line 240: nit: let's use the `Error::user()` shorthand
```suggestion
            T::new_from_slice(key).map_err(|_| Error::user(Code::InvalidLength))?;
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+        hash.last_error().map(|()| HmacApi(hash))
+    }
+
+    /// Call update from RustCrypto API and then last_error.
```
ia0, line 244:
```suggestion
    /// Updates the HMAC with the provided data.    
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+        hash.last_error().map(|()| HmacApi(hash))
+    }
+
+    /// Call update from RustCrypto API and then last_error.    
+    pub fn update(&mut self, data: &[u8]) -> Result<(), Error> {
+        self.0.update(data);
+        self.0.last_error()
+    }
+
+    /// Call finalize_into_reset from RustCrypto API and then last_error.
```
ia0, line 250:
```suggestion
    /// Finalizes the HMAC to the provided output.
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+    pub fn finalize_into_reset(&mut self, out: &mut Output<Self>) -> Result<(), Error> {
```
ia0, line 251: Same here
```suggestion
    pub fn finalize_into(self, out: &mut Output<T>) -> Result<(), Error> {
```

crates/board/src/crypto.rs
ia0: Yes, we can't use `finalize_into` in the implementation and have to use `finalize_into_reset`, but we are hiding this fact (and preventing further usage of `self`) by taking `self` as parameter (and renaming to `finalize_into`).

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -63,13 +67,18 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
     let result = try {
         let context = match convert_hash_algorithm::<B>(*algorithm)?? {
             #[cfg(feature = "board-api-crypto-sha256")]
-            Algorithm::Sha256 => HashContext::Sha256(board::crypto::Sha256::<B>::default()),
+            Algorithm::Sha256 => HashApi::<board::crypto::Sha256<B>>::new()
+                .map(|hash_api| HashContext::Sha256(hash_api.0)),
```
ia0, line 71: I think we can change the definition of HashContext to make ShaX and HmacShaX take the wrappers. Which means here you should be able to write:
```suggestion
                .map(HashContext::Sha256),
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -81,14 +90,17 @@ fn update<B: Board>(mut call: SchedulerCall<B, api::update::Sig>) {
     let memory = scheduler.applet.store.memory();
     let result = try {
         let data = memory.get(*data, *length)?;
-        match scheduler.applet.hashes.get_mut(*id as usize)? {
+        let context = match scheduler.applet.hashes.take(*id as usize)? {
             #[cfg(feature = "board-api-crypto-sha256")]
-            HashContext::Sha256(context) => context.update(data),
+            HashContext::Sha256(hash) => HashApi(hash).update(data),
```
ia0, line 95: And here you wouldn't need to rewrap, because Sha256 would already contain the wrapped type.
```suggestion
            HashContext::Sha256(context) => context.update(data),
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -156,14 +169,17 @@ fn hmac_update<B: Board>(mut call: SchedulerCall<B, api::hmac_update::Sig>) {
     let memory = scheduler.applet.store.memory();
     let result = try {
         let data = memory.get(*data, *length)?;
-        match scheduler.applet.hashes.get_mut(*id as usize)? {
+        let context = match scheduler.applet.hashes.take(*id as usize)? {
```
ia0, line 172: It shouldn't be needed to `take` here.

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+/// `last_error()` is called and returns `Ok(())`, signifying that the return value is correct.
+#[cfg(any(feature = "internal-api-crypto-hash", feature = "internal-api-crypto-hmac"))]
+pub trait LastError {
+    /// Returns the last error of infallible operations.
+    fn last_error(&self) -> Result<(), Error>;
+}
+
+/// Wrapper API around Hash that calls last_error.
+#[cfg(feature = "internal-api-crypto-hash")]
+pub struct HashApi<T: Hash>(pub T);
```
ia0, line 198: We don't need to expose the internals. If a method is needed, we should add it to the wrapper.
```suggestion
pub struct HashApi<T: Hash>(T);
```

crates/board/src/crypto.rs
```diff
@@ -158,8 +177,79 @@ impl crate::Supported for sha2::Sha256 {}
...
+
+/// Wrapper API around Hmac that calls last_error.
+#[cfg(feature = "internal-api-crypto-hmac")]
+pub struct HmacApi<T: Hmac>(pub T);
```
ia0, line 202:
```suggestion
pub struct HmacApi<T: Hmac>(T);
```

crates/scheduler/src/call/crypto/hash.rs
ia0: I can't reproduce this problem. Maybe it has been fixed in the meantime.

ia0 (commented):
Thanks! Looks good, it's getting closer. I suspect there might be some difficulties with imports eventually, but we'll deal with them once the rest is fixed.

crates/scheduler/src/call/crypto/hash.rs
zhouwfang: But we need to call last error right?

crates/scheduler/src/call/crypto/hash.rs
zhouwfang: With `get_mut`, the `hash` in `HashContext::Sha256(hash) => HashApi(hash).update(data)` is `&mut`, and it couldn't be moved.

crates/board/src/crypto.rs
zhouwfang: I'll change this after understanding the other comment on `HashApi(hash).update(data)`, since it may require creating a constructor.

crates/board/src/crypto.rs
```diff
@@ -195,60 +202,49 @@ pub trait LastError {
...
-#[cfg(feature = "internal-api-crypto-hmac")]
-impl<T: Hmac> OutputSizeUser for HmacApi<T> {
-    type OutputSize = T::OutputSize;
-}
+pub struct HmacApi<T: Hmac>(T);
 
 #[cfg(feature = "internal-api-crypto-hash")]
 impl<T: Hash> HashApi<T> {
-    /// Constructor of HmacApi.
+    /// Create a hash wrapper.
```
ia0, line 213: nit: this project follows [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Function_Comments) for function documentation
```suggestion
    /// Creates a hash wrapper.
```
(similarly for the other functions below)

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -63,13 +67,20 @@ fn initialize<B: Board>(mut call: SchedulerCall<B, api::initialize::Sig>) {
     let result = try {
         let context = match convert_hash_algorithm::<B>(*algorithm)?? {
             #[cfg(feature = "board-api-crypto-sha256")]
-            Algorithm::Sha256 => HashContext::Sha256(board::crypto::Sha256::<B>::default()),
+            Algorithm::Sha256 => {
+                HashApi::<board::crypto::Sha256<B>>::new().map(HashContext::Sha256)
```
ia0, line 71: nit: the type can be inferred (same for the 3 other occurrences below)
```suggestion
                board::crypto::HashApi::new().map(HashContext::Sha256)
```
And by fully qualifying the HashApi type, no need to import anything.

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -14,9 +14,13 @@
...
 #![allow(unused_imports)]
 
-use digest::{FixedOutput, InvalidLength, KeyInit, Output, Update};
+use std::ptr::hash;
+
+use digest::{FixedOutput, InvalidLength, KeyInit, Mac, Output, Update};
 use generic_array::GenericArray;
 use wasefire_applet_api::crypto::hash::{self as api, Algorithm, Api};
+use wasefire_applet_api::crypto::Api::Hash;
+use wasefire_board_api::crypto::{HashApi, HmacApi, LastError};
```
ia0, line 23: You can remove this line (see below why).

Actually, you shouldn't need to touch any import.

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -14,9 +14,13 @@
 
 #![allow(unused_imports)]
 
-use digest::{FixedOutput, InvalidLength, KeyInit, Output, Update};
+use std::ptr::hash;
+
```
ia0, line 18: looks like a mistake
```suggestion
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -14,9 +14,13 @@
...
+use digest::{FixedOutput, InvalidLength, KeyInit, Mac, Output, Update};
```
ia0, line 19:
```suggestion
use digest::{FixedOutput, InvalidLength, KeyInit, Output, Update};
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -14,9 +14,13 @@
...
 use generic_array::GenericArray;
 use wasefire_applet_api::crypto::hash::{self as api, Algorithm, Api};
+use wasefire_applet_api::crypto::Api::Hash;
```
ia0, line 22:
```suggestion
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -88,7 +99,6 @@ fn update<B: Board>(mut call: SchedulerCall<B, api::update::Sig>) {
             HashContext::Sha384(context) => context.update(data),
             _ => trap_use!(data),
         }
-        Ok(())
     };
     call.reply(result);
```
ia0, line 103: For the inference problem you get here, you have to annotate the result variable:

```rust
let result: Result<Result<(), _>, _> = try {
```

(some for the other occurrence below)

ia0 (commented):
Looks good. I suspect after this iteration, it should be pretty close to code-complete. Then it will be about following the CI errors and fixing them.

ia0 (approved):

Issue #43, opened 2023-03-22
Improve Rust prelude error messages


ia0:
When calling the prelude with wrong arguments, instead of panicking without error message, an error message could describe the problem. For example when calling `led::set()` with an out-of-bound index.

lukeyeh:
> Prelude functions should panic on errors that the user could have prevented (e.g. out of bound index).

Hey just to make sure I understand this point. By panicking do you mean using [`scheduling::abort`](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/api-desc/src/scheduling.rs#L33) somewhere in the applet? or panicking somewhere in the scheduler using [`log::panic!`](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/logger/src/lib.rs#L73)

For example, for dealing with an index out of bounds error on `led::set`. I assume we would have to get the out-of-bounds-error from [here](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/scheduler/src/call/led.rs#L57). Then,

* If we wanted to panic in the applet we would have to somehow push the error back to the applet and then call another `scheduling::abort`  some where [here](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/prelude/src/led.rs#L49) ?
* panicking in the scheduler I believe would be as simple as just calling `log::panic!` [here](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/scheduler/src/call/led.rs#L57) instead of sending the error back to the scheduler to trap [here](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/scheduler/src/lib.rs#L188)

PR #434, opened 2024-04-19
`led::{get, set}`: panic with an `InvalidArgument` error instead of trapping

lukeyeh:
panic with an `InvalidArgument` error instead of trapping, also ensure we track the callers for #43.

For example the editing [`examples/rust/blink/src/lib.rs`](https://github.com/google/wasefire/blob/main/examples/rust/blink/src/lib.rs) to the following,

```rust
#![no_std]
wasefire::applet!();

use core::time::Duration;

fn main() {
    // Make sure there is at least one LED.
    let num_leds = led::count();
    assert!(num_leds > 0, "Board has no LEDs.");
    led::set(led_index, led::On);
}
```

and running `cargo xtask applet rust blink runner host `, would result in the following error:

```rust
Board initialized. Starting scheduler.                                                                                                                                                    
0.000848: panicked at src/lib.rs:29:5:                                                                                                                                                   
called `Result::unwrap()` on an `Err` value: User:InvalidArgument                                                                                                                        
Applet aborted (probably a panic).
```

I couldn't find an `OutOfBounds` error  [here](https://github.com/google/wasefire/blob/33fe4839e4927cf19cfde86c2b56dd63df65954a/crates/error/src/lib.rs#L141-L173), so decided to use `InvalidArgument`. Let me know if we should add an `OutOfBounds` error if that would be better.

### DIFF
--- a/crates/prelude/CHANGELOG.md
+++ b/crates/prelude/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Document all public API
 - Remove deprecated `println!` macro
 - Add `Sha384` and `HmacSha384` to the `rust-crypto` feature
+- Led returns error instead of trapping when indexing out of bounds.
 
 ## 0.5.0
 
--- a/crates/prelude/CHANGELOG.md
+++ b/crates/prelude/CHANGELOG.md
@@ -4,10 +4,10 @@
 
 ### Minor
 
+- Change `led::{get,set}()` to never trap and return an error instead
 - Document all public API
 - Remove deprecated `println!` macro
 - Add `Sha384` and `HmacSha384` to the `rust-crypto` feature
-- Led returns error instead of trapping when indexing out of bounds.
 
 ## 0.5.0
 
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.3.1-git
 
+### Minor
+
+- Change `led::{get,set}()` to never trap and return an error instead
+
 ### Patch
 
 - Simplify `#[cfg(all)]` attributes between board and applet features
@@ -38,7 +42,6 @@
 - Permit applets to call `debup::println()` during `init()`
 - Use `debug::exit()` board API when the applet traps
 - Use `log::panic!()` on interpreter errors
-- Change led scheduler calls to return out of bounds errors instead of trapping
 
 ### Patch
 
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Permit applets to call `debup::println()` during `init()`
 - Use `debug::exit()` board API when the applet traps
 - Use `log::panic!()` on interpreter errors
+- Change led scheduler calls to return out of bounds errors instead of trapping
 
 ### Patch
 
--- a/crates/scheduler/src/call/led.rs
+++ b/crates/scheduler/src/call/led.rs
@@ -18,9 +18,9 @@ use wasefire_board_api::led::Api as _;
 use wasefire_board_api::Api as Board;
 #[cfg(feature = "board-api-led")]
 use wasefire_board_api::{self as board, Id, Support};
+#[cfg(feature = "board-api-led")]
 use wasefire_error::{Code, Error};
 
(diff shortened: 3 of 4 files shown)


crates/scheduler/src/call/led.rs
```diff
@@ -44,8 +44,10 @@ fn count<B: Board>(call: SchedulerCall<B, api::count::Sig>) {
 fn get<B: Board>(call: SchedulerCall<B, api::get::Sig>) {
     let api::get::Params { led } = call.read();
     let result = try {
-        let id = Id::new(*led as usize).ok_or(Trap)?;
-        board::Led::<B>::get(id)
+        match Id::new(*led as usize) {
+            Some(id) => board::Led::<B>::get(id),
+            None => Err(Error::new(Space::User, Code::InvalidArgument)),
```
ia0, line 49: nit: if I remember correctly there's a shorthand for that
```suggestion
            None => Err(Error::user(Code::InvalidArgument)),
```

crates/scheduler/src/call/led.rs
```diff
@@ -44,8 +44,10 @@ fn count<B: Board>(call: SchedulerCall<B, api::count::Sig>) {
 fn get<B: Board>(call: SchedulerCall<B, api::get::Sig>) {
     let api::get::Params { led } = call.read();
     let result = try {
-        let id = Id::new(*led as usize).ok_or(Trap)?;
-        board::Led::<B>::get(id)
+        match Id::new(*led as usize) {
```
ia0, line 47: For convenience, we should probably change `Id::new()` to return `Result<Self, Error>`. Then, I would expect this function to look like:

```rust
let result = try {
    board::Led::<B>::get(Id:new(*led as usize)?)
};
call.reply(Ok(result));
```

Note how the reply is `Ok(result)` showing that `led::get()` never traps. (Maybe type annotations are needed though.)

ia0 (commented):
Thanks! Looks good to me. I'll check on Monday if I don't have some `OutOfBounds` error somewhere, otherwise I would suggest adding it. And I think we should change `Id::new()`. This will factorize code from many places.

crates/scheduler/src/call/led.rs
lukeyeh: Done.

crates/scheduler/src/call/led.rs
lukeyeh: Would you like me to change it in this PR? I ask because it seems like I would have to touch other modules given the type change, and perhaps we want to keep this PR isolated.

crates/prelude/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Document all public API
 - Remove deprecated `println!` macro
 - Add `Sha384` and `HmacSha384` to the `rust-crypto` feature
+- Led returns error instead of trapping when indexing out of bounds.
```
ia0, line 10: nit: the changelog is "append to top" and ideally starts with a verb describing the change, for example

```
- Change `led::{get,set}()` to never trap and return an error instead
```

crates/scheduler/CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - Permit applets to call `debup::println()` during `init()`
 - Use `debug::exit()` board API when the applet traps
 - Use `log::panic!()` on interpreter errors
+- Change led scheduler calls to return out of bounds errors instead of trapping
```
ia0, line 41: Same here, let's move this line at the top. Also this is under the wrong version. You need to create a minor subsection under 0.3.1-git and above the patch subsection already present. Also nit regarding the description, I would suggest something a bit more concise (the word scheduler is superfluous in the scheduler changelog) and actually reusing the same line as for the prelude:

```
- Change `led::{get,set}()` to never trap and return an error instead
```

crates/scheduler/src/call/led.rs
ia0: Yes, another PR is probably better. This PR is already improving the situation and only has a small code duplication which is going to be fixed by the other PR.

ia0 (changes requested):

crates/scheduler/src/call/led.rs
ia0: I've created #435 to track that.

ia0 (approved):

PR #441, opened 2024-04-23
Change `scheduler::call::button::new` to never trap

lukeyeh:
This PR changes `scheduler::call::button::new` to never trap and return a `Result<(), Error>` instead.

### DIFF
--- a/book/src/applet/prelude/button1.rs
+++ b/book/src/applet/prelude/button1.rs
@@ -38,7 +38,7 @@ fn main() {
 
         //{ ANCHOR: listener
         // We start listening for state changes with the handler.
-        let listener = button::Listener::new(index, handler);
+        let listener = button::Listener::new(index, handler).unwrap();
         //} ANCHOR_END: listener
 
         //{ ANCHOR: leak
--- a/crates/prelude/CHANGELOG.md
+++ b/crates/prelude/CHANGELOG.md
@@ -4,9 +4,7 @@
 
 ### Major
 
-- Change `button::Listener::new()` to never panic and return an error instead.
-
-## 0.5.1-git
+- Change `button::Listener::new()` to never panic and return an error instead
 
 ### Minor
 
--- a/crates/prelude/CHANGELOG.md
+++ b/crates/prelude/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.6.0-git
+
+### Major
+
+- Change `button::Listener::new()` to never panic and return an error instead.
+
 ## 0.5.1-git
 
 ### Minor
--- a/crates/prelude/src/button.rs
+++ b/crates/prelude/src/button.rs
@@ -21,6 +21,7 @@
 use alloc::boxed::Box;
 
 use wasefire_applet_api::button as api;
+use wasefire_error::Error;
 
 pub use self::api::State;
 pub use self::api::State::*;
@@ -64,15 +65,15 @@ impl<H: Handler> Listener<H> {
     /// # Examples
     ///
     /// ```ignore
-    /// Listener::new(index, |state| debug!("Button has been {state:?}"))
+    /// Listener::new(index, |state| debug!("Button has been {state:?}"))?
     /// ```
-    pub fn new(button: usize, handler: H) -> Self {
+    pub fn new(button: usize, handler: H) -> Result<Self, Error> {
         let handler_func = Self::call;
         let handler = Box::into_raw(Box::new(handler));
         let handler_data = handler as *const u8;
         let params = api::register::Params { button, handler_func, handler_data };
-        convert_unit(unsafe { api::register(params) }).unwrap();
-        Listener { button, handler }
+        convert_unit(unsafe { api::register(params) })?;
+        Ok(Listener { button, handler })
     }
 
     /// Stops listening.
--- a/crates/scheduler/src/call/button.rs
+++ b/crates/scheduler/src/call/button.rs
@@ -16,11 +16,11 @@ use wasefire_applet_api::button::{self as api, Api};
 use wasefire_board_api::Api as Board;
 #[cfg(feature = "board-api-button")]
 use wasefire_board_api::{self as board, button::Api as _, Id, Support};
+#[cfg(feature = "board-api-button")]
+use wasefire_error::{Code, Error};
 
 #[cfg(feature = "board-api-button")]
 use crate::event::{button::Key, Handler};
-#[cfg(feature = "board-api-button")]
-use crate::Trap;
 use crate::{DispatchSchedulerCall, SchedulerCall};
 
 pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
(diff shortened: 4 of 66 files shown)


footprint.toml
ia0, line 1: This file should not be committed. It's temporarily generated to compute the footprint impact.

book/src/applet/prelude/button1.rs
```diff
@@ -30,15 +30,15 @@ fn main() {
     //{ ANCHOR: loop
     // For each button on the board.
     for index in 0 .. count {
-    //} ANCHOR_END: loop
+        //} ANCHOR_END: loop
```
ia0, line 33: The idea was to keep the ANCHOR_END aligned with its ANCHOR, even though from a pure "Rust comment" perspective this is wrong. Those comments are actually outside Rust and happens before. They are just comments for the file to still have a Rust meaning.

crates/prelude/src/button.rs
```diff
@@ -66,13 +67,19 @@ impl<H: Handler> Listener<H> {
     /// ```ignore
     /// Listener::new(index, |state| debug!("Button has been {state:?}"))
     /// ```
-    pub fn new(button: usize, handler: H) -> Self {
+    #[track_caller]
+    pub fn new(button: usize, handler: H) -> Result<Self, Error> {
```
ia0, line 71: There's a good design question here. Do we want to split error handling in the prelude? There are essentially 3 high-level categories:
- Never return error and always panic.
- Always return error and never panic.
- Return non-user error and panic on user error.

The first 2 have a signature that describe exactly what they do. Either the function returns a `Result` or not. The last one is more weird, because it returns a `Result` forcing the user to do something but still takes decision for them. So it's like a lose-lose situation, where the user loses the convenience of not dealing with errors but also loses the option to deal with its own errors. So I would suggest to keep prelude signatures to either of the first 2 alternatives (either always panic, or always return an error). It's not yet clear to me what's best between both and can be decided on a function basis. Ultimately I would expect return `Result<_, Error>` to be the best option and having the `main()` function return `Result<_, Error>` too, such that users can simply use `?` if they don't care about the error.

So in this particular case, I would suggest return a `Result` as you did (updating the documentation by adding `?`) and never panicking, letting the user handle all errors.

crates/scheduler/src/call/button.rs
```diff
@@ -42,30 +40,35 @@ fn count<B: Board>(call: SchedulerCall<B, api::count::Sig>) {
...
+        let button = Id::new(*button as usize)?;
+        call.scheduler()
+            .applet
+            .enable(Handler {
+                key: Key { button }.into(),
+                inst,
+                func: *handler_func,
+                data: *handler_data,
+            })
+            .map_err(|_| Error::internal(Code::InvalidState))?;
```
ia0, line 56: That's actually a user error. And we should actually update `enable` and `disable` to return an `Error` instead of `Trap`. The error here is that the user called `register` twice without calling `unregister` in between.

ia0 (commented):
Quick review even though it's a draft, just to avoid going in the wrong direction.

book/src/applet/prelude/button1.rs
lukeyeh: Removed, I think my formatter did this.

crates/prelude/src/button.rs
lukeyeh: I've made it such that we always return an error here as you suggested However, do you mind if I open another PR to have the `Result` be propagated up to `main`'s return value and leave the calls to `button::Listener::new` to unwrap for now? I think I'll have to touch many of the other examples that don't use `button::Listener::new`, as I'll be changing the function signature of main.

crates/scheduler/src/call/button.rs
lukeyeh: Changed it to an internal error. Do you think `InvalidState` is the correct code here?

crates/prelude/src/button.rs
ia0: > do you mind if I open another PR to have the `Result` be propagated up to `main`'s return value

Yes, that's actually what I had in mind.

crates/prelude/src/button.rs
```diff
@@ -66,13 +67,14 @@ impl<H: Handler> Listener<H> {
     /// ```ignore
     /// Listener::new(index, |state| debug!("Button has been {state:?}"))
```
ia0, line 68:
```suggestion
    /// Listener::new(index, |state| debug!("Button has been {state:?}"))?
```

crates/scheduler/src/call/button.rs
```diff
@@ -42,30 +40,35 @@ fn count<B: Board>(call: SchedulerCall<B, api::count::Sig>) {
 
 #[cfg(feature = "board-api-button")]
 fn register<B: Board>(mut call: SchedulerCall<B, api::register::Sig>) {
+    use wasefire_error::{Code, Error};
```
ia0, line 43: Let's add this at the top instead (like you did in the other modules):

```rust
#[cfg(feature = "board-api-button")]
use wasefire_error::{Code, Error};
```

I think rust-analyzer adds imports to the function when the function is feature-gated. That's sometimes the right choice, but not always (e.g. when multiple functions with the same feature-gate need the same imports).

ia0 (commented):
LGTM modulo that the prelude change is a major change.

crates/scheduler/src/call/button.rs
ia0: Yes, that's a good choice.

crates/scheduler/src/call/button.rs
lukeyeh: Done.

crates/prelude/src/button.rs
lukeyeh: Done.

crates/prelude/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Minor
 
+- Change `button::new()` to never panic and return a `Result<(), Error>` instead.
```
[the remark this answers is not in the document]
lukeyeh: Done.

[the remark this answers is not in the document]
lukeyeh: Done.

crates/prelude/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.5.2-git
+
+### Major
+
+- Change `button::Listener::new()` to never panic and return a `Result<(), Error>` instead.
+
 ## 0.5.1-git
```
ia0, line 9: Sorry, I wasn't clear. What I meant by "bump the version" was going from a patch/minor bump (0.5.1-git) to a major bump (0.6.0-git). See the [Cargo Book](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) and the [SemVer](https://semver.org/) specification. A changelog files is a sequence of versions, where a version is a level-2 header with a semver. Only the very first version may have a pre-release version (and when it does, it must be -git). Versions are sorted by precedence (highest precedence first). Each version may have a major, minor, and/or patch level-3 header describing the changes of that category for this version against the previous one. The version number must match the highest level-3 header (major version if there's a major section, minor if minor, patch if patch) including the cargo rule that for 0.x.y, x is major and y minor/patch (and for 0.0.x, x is bumped for all changes, but wasefire doesn't use this numbering scheme). I'll probably have to move this information to some `docs/changelog.md` file.

```suggestion
## 0.6.0-git

### Major

- Change `button::Listener::new()` to never panic and return a `Result<(), Error>` instead.
```

crates/prelude/src/button.rs
```diff
@@ -64,15 +65,16 @@ impl<H: Handler> Listener<H> {
     /// # Examples
     ///
     /// ```ignore
-    /// Listener::new(index, |state| debug!("Button has been {state:?}"))
+    /// Listener::new(index, |state| debug!("Button has been {state:?}"))?
     /// ```
-    pub fn new(button: usize, handler: H) -> Self {
+    #[track_caller]
```
ia0, line 70: No need for track-caller if it never panics.

ia0 (changes requested):

crates/prelude/CHANGELOG.md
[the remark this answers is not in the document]
ia0: Only the first part seems done, but there was also using a shorter textual form for "error", which contains the same information.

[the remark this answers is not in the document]
lukeyeh: sorry about missing that!

crates/prelude/CHANGELOG.md
lukeyeh: Thanks! I wasn't too familiar with SemVer and the Cargo book thank you for the references :+1:

crates/prelude/src/button.rs
lukeyeh: Removed.

crates/prelude/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.6.0-git
+
+### Major
+
+- Change `button::Listener::new()` to never panic and return an error instead.
+
 ## 0.5.1-git
```
ia0, line 9: The 0.6.0-git section is not a new one, but a major bump of the current -git version. I'll apply the change and merge.

```suggestion
- Change `button::Listener::new()` to never panic and return an error instead
```

ia0 (commented):
Thanks! LGTM. I'll apply the suggestion and merge.

ia0 (approved):

PR #445, opened 2024-04-25
Warn and detect edition of generated files

### LINKED ISSUE #444: Better error messages when book and examples are out of sync.
Noticed this while trying to run `scripts/ci.sh` on some changes that touched `examples/`. I believe these are the steps causing the issues:

1. Edit any of some of the examples under `examples/` and commit
2. `scripts/ci.sh` will [run `scripts/sync.sh`](https://github.com/google/wasefire/blob/fe57cd1562c3414a8d3649305c75a1935d85b360/scripts/ci.sh#L24). This will [then copy](https://github.com/google/wasefire/blob/fe57cd1562c3414a8d3649305c75a1935d85b360/scripts/sync.sh#L23-L32) over the example in book to the examples in `examples/` and "undo" changes in made in `examples/`
3. Now repo isn't clean so we'll hit the error `Repository is not clean` in [`scripts/publish.sh`](https://github.com/google/wasefire/blob/fe57cd1562c3414a8d3649305c75a1935d85b360/scripts/publish.sh#L22)

I get the point here (we wan't to keep the book and `examples/` in sync), but it took me a while to figure out that some of the scripts were editing the files under me and the error was a little opaque. Maybe we can report error message in `sync.sh` instead?

ia0:
If the detection happens to have false positive, it will be removed in favor of keeping the warning only.

Fixes #444

### DIFF
--- a/scripts/sync.sh
+++ b/scripts/sync.sh
@@ -21,7 +21,23 @@ set -e
 cargo xtask update-apis
 
 book_example() {
-  sed '/ANCHOR/d' book/src/applet/prelude/$1.rs > examples/rust/$2/src/lib.rs
+  local src=book/src/applet/prelude/$1.rs
+  local dst=examples/rust/$2/src/lib.rs
+  if [ $dst -nt $src ]; then
+    t "Update $src instead of $dst"
+    e "$dst seems to have been manually modified"
+  fi
+  # Besides removing all anchors, we insert a warning before the #![no_std] line, which all examples
+  # should have near the beginning of the file.
+  sed '/ANCHOR/d;/^#!\[no_std\]$/{i \
+// DO NOT EDIT MANUALLY:\'"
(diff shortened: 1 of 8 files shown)


scripts/sync.sh
```diff
@@ -21,7 +21,23 @@ set -e
 cargo xtask update-apis
 
 book_example() {
-  sed '/ANCHOR/d' book/src/applet/prelude/$1.rs > examples/rust/$2/src/lib.rs
+  local src=book/src/applet/prelude/$1.rs
+  local dst=examples/rust/$2/src/lib.rs
+  if [ -z "$CI" -a $dst -nt $src ]; then
```
ia0, line 26: Actually the problem occurs when cloning a repository (and probably also when changing branch). We don't control in which order git will update them. So we need to compare with a delta (dst is not newer than src by more than 2 seconds).

PR #449, opened 2024-05-03
Add std feature to error crate

ia0:


ia0:
This was kind of reverted by #500 because we use nightly and can simply implement `core::error::Error` instead.

PR #453, opened 2024-05-04
Change gpio functions to never trap

lukeyeh:
* Changes `scheduler/call/gpio.rs` to not trap and always return errors.
* Changes `prelude/gpio` to never panic and always return errors.

### DIFF
--- a/crates/prelude/CHANGELOG.md
+++ b/crates/prelude/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Major
 
+- Change `Gpio` to never panic and return an error instead
 - Change `button::Listener::new()` to never panic and return an error instead
 
 ### Minor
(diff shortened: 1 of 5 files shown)


crates/prelude/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Major
 
+- Change `Gpio` to never panic and return an error instead
```
ia0, line 7: nit
```suggestion
- Change `gpio` module to never panic and return an error instead
```

crates/scheduler/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Minor
 
+- Change gpio functions to never trap and return an error instead
```
ia0, line 7: nit
```suggestion
- Change `gpio` module to never trap and return an error instead
```

ia0 (commented):
Thanks! LGTM. I'll commit the suggestions and merge.

ia0 (approved):

PR #461, opened 2024-05-09
Impl Storage

kofls:


examples/rust/opensk/Cargo.toml
```diff
@@ -10,6 +10,7 @@ opensk-lib = { path = "../../../third_party/OpenSK/libraries/opensk", package =
 persistent_store = { path = "../../../third_party/OpenSK/libraries/persistent_store" }
 rand_core = "0.6.4"
 wasefire = { path = "../../../crates/prelude" }
+wasefire-store = { path = "../../../crates/store", features = ["std"] }
```
ia0, line 13: We shouldn't use `std`.

examples/rust/opensk/src/storage.rs
ia0, line 1: This file can now be deleted.

examples/rust/opensk/src/lib.rs
```diff
@@ -64,12 +68,26 @@ struct WasefireEnv {
...
+
+    fn remove(&mut self, key: usize) -> CtapResult<()> {
+        Ok(store::remove(key)?)
+    }
+
+    fn iter(&self) -> CtapResult<PersistIter<'_>> {
+        // Ok(Box::new(self.store.iter()?.map(|handle| match handle {
+        //     Ok(handle) => Ok(handle.get_key()),
+        //     Err(error) => Err(error.into()),
+        // })))
```
ia0, line 90: For this we need to extend the applet store API. I'll take a look at how to do it, because either we use allocation or we do many syscalls. I expect the first to be better, but need to take a look.

examples/rust/opensk/src/storage.rs
kofls: Deleted the file.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/Cargo.toml
kofls: I'm not sure what I'm doing wrong here but without the `std` feature, the test throws the error:
```
error[E0432]: unresolved import `crate::Nat`
   --> /usr/local/google/home/suryamidatala/Documents/repos/wasefire/crates/store/src/format/bitfield.rs:20:13
    |
20  | use crate::{Nat, StoreError, StoreResult};
    |             ^^^ no `Nat` in the root
    |
note: found an item that was configured out
   --> /usr/local/google/home/suryamidatala/Documents/repos/wasefire/crates/store/src/lib.rs:407:10
    |
407 | pub type Nat = u32;
    |          ^^^
```

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
 
-impl AttestationStore for WasefireEnv {
-    fn get(&mut self, id: &Id) -> Result<Option<Attestation>, Error> {
-        todo!()
+impl Persist for WasefireEnv {
+    fn find(&self, key: usize) -> CtapResult<Option<Vec<u8>>> {
+        let Ok(res) = store::find(key) else {
+            // TODO: What is the correct error code to use here? There doesn't
+            // seem to be an error for value not found or something like that.
+            return Err(CTAP1_ERR_INVALID_COMMAND);
```
kofls, line 75: @ia0 Do we need to convert every error code to a corresponding `Ctap2StatusCode`?
And is this the mapping for the error codes: https://github.com/google/wasefire/blob/dev/opensk/crates/error/src/lib.rs#L144-L179

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
+        };
+        match res {
+            Some(g) => Ok(Some(g.into_vec())),
+            None => Ok(None),
+        }
```
ia0, line 80: nit
```suggestion
        Ok(res.map(|x| x.into_vec()))
```

Alternatively, the whole function could be a single match:

```rust
    Err(_) => Err(Ctap2StatusCode::XXX),
    Ok(None) => Ok(None),
    Ok(Some(data)) => Ok(Some(data.into_vec())),
```

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
+    fn insert(&mut self, key: usize, value: &[u8]) -> CtapResult<()> {
+        // TODO: Impl `From<wasefire::Error>` for `Ctap2StatusCode`
+        if let Err(e) = store::insert(key, value) {
+            // TODO: What is the correct error code to use here? There doesn't
+            // seem to be an error for value not found or something like that.
+            // We probably need to convert `wasefire_error::Code` to `Ctap2StatusCode`.
+            Err(CTAP1_ERR_INVALID_COMMAND)
+        } else {
+            Ok(())
+        }
```
ia0, line 92: Same here, this can just be a match.
```suggestion
        match store::insert(key, value) {
            Err(_) => Err(Ctap2StatusCode::Xxx),
            Ok(()) => Ok(()),
        }
```

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
+        if let Err(e) = store::remove(key) {
+            // TODO: What is the correct error code to use here?
+            Err(CTAP1_ERR_INVALID_COMMAND)
+        } else {
+            Ok(())
+        }
+    }
+
+    fn iter(&self) -> CtapResult<PersistIter<'_>> {
+        let Ok(keys) = store::keys() else { return Err(CTAP1_ERR_INVALID_COMMAND) };
```
ia0, line 105: nit: this is much more idiomatic to use `?`
```suggestion
        let keys = store::keys().map_err(|_| CTAP1_ERR_INVALID_COMMAND)?;
```

ia0 (commented):
Looks mostly fine, thanks!

@kaczmarczyck can you help us with what errors are expected?

examples/rust/opensk/src/lib.rs
kaczmarczyck: In general, only 4 errors are relevant:
- `Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR`: We got into a internal state that shouldn't have been possible.
- `Ctap2StatusCode::CTAP2_ERR_VENDOR_HARDWARE_FAILURE`: Something broke, and storage detected that it was a hardware failure.
- `Ctap2StatusCode::CTAP2_ERR_KEY_STORE_FULL`: We can't do the operation because storage is full or out of lifetime.
- `Ctap2StatusCode::CTAP1_ERR_OTHER`: Something else entirely, but I don't think you need this.

See the `From` in `libraries/opensk/src/ctap/status_code.rs` in OpenSK for how it converts from the Store errors.

In this particular case, we return a `Result` of an `Option`. Key not found is `Ok(None)`. All errors are likely `Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)` or `Err(tap2StatusCode::CTAP2_ERR_VENDOR_HARDWARE_FAILURE)`, the idea of `.map_err(convert)?` seems reasonable.

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
     }
 
-    fn set(&mut self, id: &Id, attestation: Option<&Attestation>) -> Result<(), Error> {
-        todo!()
+    fn insert(&mut self, key: usize, value: &[u8]) -> CtapResult<()> {
+        // TODO: Impl `From<wasefire::Error>` for `Ctap2StatusCode`
+        if let Err(e) = store::insert(key, value) {
+            // TODO: What is the correct error code to use here? There doesn't
+            // seem to be an error for value not found or something like that.
+            // We probably need to convert `wasefire_error::Code` to `Ctap2StatusCode`.
```
kaczmarczyck, line 88: Same, call `.map_err(convert)?`.

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +67,55 @@ struct WasefireEnv {
...
+            Err(CTAP1_ERR_INVALID_COMMAND)
+        } else {
+            Ok(())
+        }
+    }
+
+    fn remove(&mut self, key: usize) -> CtapResult<()> {
+        if let Err(e) = store::remove(key) {
+            // TODO: What is the correct error code to use here?
```
kaczmarczyck, line 97: Same, `.map_err(convert)?`.

examples/rust/opensk/src/lib.rs
kaczmarczyck: Same `.map_err(convert)?`.

examples/rust/opensk/src/lib.rs
ia0: Thanks! So I would suggest the following mapping:

```rust
// I'm abusing notations because space is u8 and code is u16, but that's the idea.
match (error.space(), error.code()) {
    (User, NotEnough) => CTAP2_ERR_KEY_STORE_FULL,
    (User | Internal, _) => CTAP2_ERR_VENDOR_INTERNAL_ERROR,
    (World, _) => CTAP2_VENDOR_HARDWARE_FAILURE,
    _ => CTAP2_ERR_OTHER,
}
```

This can be improved in the future if needed.

examples/rust/opensk/Cargo.toml
kofls: I've removed `persistent_store` and `wasefire_stub`. But I'm using impl `CryptoRng` and `RngCore` traits in rng.rs. So shouldn't we need that?

Also I've added `wasefire_error` because we need `Space` and `Code` which I don't think are being re-exported in wasefire.

examples/rust/opensk/src/lib.rs
kofls: Ok, thank you!

I've made the change but i'm converting the u8 & u16 to `Space` and `Code` and then destructuring in the match. Not sure if there is a more idiomatic way to do this.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/src/lib.rs
kofls: Done.

examples/rust/opensk/Cargo.toml
ia0: > But I'm using impl `CryptoRng` and `RngCore` traits in rng.rs. So shouldn't we need that?

My bad, I forgot OpenSK uses those traits in its public API. That's expected to depend on `rand_core`, all good.

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +70,47 @@ struct WasefireEnv {
     // c: PhantomData<C>,
 }
 
-impl AttestationStore for WasefireEnv {
-    fn get(&mut self, id: &Id) -> Result<Option<Attestation>, Error> {
-        todo!()
+fn convert(error: wasefire::Error) -> Ctap2StatusCode {
+    match (Space::try_from(error.space()), Code::try_from(error.code())) {
+        (_, Ok(Code::NotEnough)) => CTAP2_ERR_KEY_STORE_FULL,
+        (Ok(Space::User) | Ok(Space::Internal), _) => CTAP2_ERR_VENDOR_INTERNAL_ERROR,
```
ia0, line 76: ultra nit
```suggestion
        (Ok(Space::User | Space::Internal), _) => CTAP2_ERR_VENDOR_INTERNAL_ERROR,
```

examples/rust/opensk/src/lib.rs
```diff
@@ -64,25 +70,47 @@ struct WasefireEnv {
...
+    }
+
+    fn remove(&mut self, key: usize) -> CtapResult<()> {
+        store::remove(key).map_err(convert)
+    }
+
+    fn iter(&self) -> CtapResult<PersistIter<'_>> {
+        let keys = store::keys().map_err(convert)?;
+        let res = keys.into_iter().map(|x| Ok(x as usize));
+        Ok(Box::new(res))
```
ia0, line 102: nit
```suggestion
        Ok(Box::new(keys.into_iter().map(|x| Ok(x as usize))))
```

ia0 (approved):
Thanks! LGTM, I'll commit the 2 nits and merge.

examples/rust/opensk/src/lib.rs
ia0: Looks good. That's even better than what I had in mind.

examples/rust/opensk/Cargo.toml
```diff
@@ -7,13 +7,12 @@ edition = "2021"
...
 rand_core = "0.6.4"
 wasefire = { path = "../../../crates/prelude" }
-wasefire-stub = { path = "../../../crates/stub", optional = true }
+wasefire-error = { path = "../../../crates/error" }
 
 [build-dependencies]
 uuid = "1.8.0"
 
 [features]
-test = ["dep:wasefire-stub", "opensk-lib/std", "persistent_store/std", "wasefire/test"]
```
ia0, line 19: Actually, you might need `wasefire-stub` for the tests. But let's see what the CI says. I don't understand why it's not running.

ia0 (dismissed):

examples/rust/opensk/Cargo.toml
ia0: Ok, I've figured it out. I forgot to update the workflow when I renamed the branch. I'll update the PR to see if it works.

ia0:
Let's try to fix the warnings, even if it's just adding `#[allow(unused_imports)]` or `#[allow(dead_code)]`.

kaczmarczyck (commented):
LGTM

ia0 (approved):

Issue #463, opened 2024-05-10
Simplify try-blocks in scheduler for API call replies


ia0:
A scheduler reply currently looks like `Result<Result<T, Error>, Trap>` which does not work nicely with try-blocks that only support one level of `Result`. This is problematic when within the same try-block some functions may return `Error` while others may return `Trap`. We can only use the `?` syntax for `Trap`. By introducing the following type:

```rust
enum Failure {
    Trap(Trap),
    Error(Error),
}
impl From<Trap> for Failure { ... }
impl From<Error> for Failure { ... }
```

It will be possible to use `?` for both `Trap` and `Error`. We also need to change the schedule reply signature to take `Result<T, Failure>` instead.

This issue is part of #43.

PR #465, opened 2024-05-11
Introduce Failure type for scheduler replies

lukeyeh:
This PR addresses the issue of handling both Trap and Error within the same try-block in the scheduler reply, resolving #463. Currently, the scheduler reply has the signature Result<Result<T, Error>, Trap>, which does not work well with try-blocks that only support one level of Result.

To resolve this, we introduce a new Failure enum that can represent either a Trap or an Error. The Failure type is defined in failure.rs, along with the Trap function. We also implement From traits for both Trap and Error to allow seamless conversion into the Failure type.

By updating the scheduler reply signature to Result<T, Failure>, we can now use the ? operator for both Trap and Error within the same try-block, making the code more concise and easier to read.

### DIFF
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Patch
 
+- Add `Failure` type to simplify try-blocks for `Scheduler::reply()` arguments
 - Update `store` errors mapping
 - Use explicit conversion from `Error` to `Trap`
 - Simplify `#[cfg(all)]` attributes between board and applet features
--- a/crates/scheduler/src/call.rs
+++ b/crates/scheduler/src/call.rs
@@ -23,7 +23,7 @@ macro_rules! or_trap {
         #[cfg(feature = $feature)]
         $name($call);
         #[cfg(not(feature = $feature))]
-        $call.reply_(Err(crate::Trap));
+        $call.reply_(Err(crate::Trap.into()));
     }};
 }
 
@@ -33,7 +33,9 @@ macro_rules! or_fail {
         #[cfg(feature = $feature)]
         $name($call);
         #[cfg(not(feature = $feature))]
-        $call.reply_(Ok(Err(wasefire_error::Error::world(wasefire_error::Code::NotImplemented))));
+        $call.reply_(Err(crate::Failure::Error(wasefire_error::Error::world(
+            wasefire_error::Code::NotImplemented,
+        ))));
     }};
 }
 
@@ -113,5 +115,5 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 
 fn syscall<B: Board>(call: SchedulerCall<B, api::syscall::Sig>) {
     let api::syscall::Params { x1, x2, x3, x4 } = call.read();
-    call.reply(B::syscall(*x1, *x2, *x3, *x4).ok_or(Trap));
+    call.reply(try { B::syscall(*x1, *x2, *x3, *x4).ok_or(Trap)?? });
 }
--- a/crates/scheduler/src/call/crypto/ec.rs
+++ b/crates/scheduler/src/call/crypto/ec.rs
@@ -39,7 +39,7 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 
 fn is_supported<B: Board>(call: SchedulerCall<B, api::is_supported::Sig>) {
     let api::is_supported::Params { curve } = call.read();
-    call.reply(convert_curve::<B>(*curve).map(|x| Ok(x.is_ok())))
+    call.reply(try { convert_curve::<B>(*curve)?.is_ok() })
 }
 
 #[cfg(feature = "internal-board-api-crypto-ecc")]
@@ -48,7 +48,7 @@ fn is_valid_scalar<B: Board>(mut call: SchedulerCall<B, api::is_valid_scalar::Si
     let scheduler = call.scheduler();
     let memory = scheduler.applet.memory();
     let result = try {
-        Ok(match convert_curve::<B>(*curve)?? {
+        match convert_curve::<B>(*curve)?? {
             #[cfg(feature = "board-api-crypto-p256")]
             Curve::P256 => {
                 let n = memory.get_array::<32>(*n)?.into();
@@ -61,7 +61,7 @@ fn is_valid_scalar<B: Board>(mut call: SchedulerCall<B, api::is_valid_scalar::Si
             }
             #[allow(unreachable_patterns)]
             _ => trap_use!(),
-        })
+        }
     };
     call.reply(result)
 }
@@ -72,7 +72,7 @@ fn is_valid_point<B: Board>(mut call: SchedulerCall<B, api::is_valid_point::Sig>
     let scheduler = call.scheduler();
     let memory = scheduler.applet.memory();
     let result = try {
-        Ok(match convert_curve::<B>(*curve)?? {
+        match convert_curve::<B>(*curve)?? {
             #[cfg(feature = "board-api-crypto-p256")]
             Curve::P256 => {
                 let x = memory.get_array::<32>(*x)?.into();
@@ -87,7 +87,7 @@ fn is_valid_point<B: Board>(mut call: SchedulerCall<B, api::is_valid_point::Sig>
             }
             #[allow(unreachable_patterns)]
             _ => trap_use!(),
-        })
+        }
     };
     call.reply(result)
 }
@@ -104,14 +104,14 @@ fn base_point_mul<B: Board>(mut call: SchedulerCall<B, api::base_point_mul::Sig>
                 let n = memory.get_array::<32>(*n)?.into();
                 let x = memory.get_array_mut::<32>(*x)?.into();
                 let y = memory.get_array_mut::<32>(*y)?.into();
-                board::crypto::P256::<B>::base_point_mul(n, x, y)
+                board::crypto::P256::<B>::base_point_mul(n, x, y)?
             }
             #[cfg(feature = "board-api-crypto-p384")]
             Curve::P384 => {
                 let n = memory.get_array::<48>(*n)?.into();
                 let x = memory.get_array_mut::<48>(*x)?.into();
                 let y = memory.get_array_mut::<48>(*y)?.into();
-                board::crypto::P384::<B>::base_point_mul(n, x, y)
+                board::crypto::P384::<B>::base_point_mul(n, x, y)?
             }
             #[allow(unreachable_patterns)]
             _ => trap_use!(),
@@ -134,7 +134,7 @@ fn point_mul<B: Board>(mut call: SchedulerCall<B, api::point_mul::Sig>) {
                 let in_y = memory.get_array::<32>(*in_y)?.into();
                 let out_x = memory.get_array_mut::<32>(*out_x)?.into();
                 let out_y = memory.get_array_mut::<32>(*out_y)?.into();
-                board::crypto::P256::<B>::point_mul(n, in_x, in_y, out_x, out_y)
+                board::crypto::P256::<B>::point_mul(n, in_x, in_y, out_x, out_y)?
             }
             #[cfg(feature = "board-api-crypto-p384")]
             Curve::P384 => {
@@ -143,7 +143,7 @@ fn point_mul<B: Board>(mut call: SchedulerCall<B, api::point_mul::Sig>) {
                 let in_y = memory.get_array::<48>(*in_y)?.into();
                 let out_x = memory.get_array_mut::<48>(*out_x)?.into();
                 let out_y = memory.get_array_mut::<48>(*out_y)?.into();
-                board::crypto::P384::<B>::point_mul(n, in_x, in_y, out_x, out_y)
+                board::crypto::P384::<B>::point_mul(n, in_x, in_y, out_x, out_y)?
             }
             #[allow(unreachable_patterns)]
             _ => trap_use!(),
@@ -165,7 +165,7 @@ fn ecdsa_sign<B: Board>(mut call: SchedulerCall<B, api::ecdsa_sign::Sig>) {
                 let message = memory.get_array::<32>(*message)?.into();
                 let r = memory.get_array_mut::<32>(*r)?.into();
                 let s = memory.get_array_mut::<32>(*s)?.into();
-                board::crypto::P256::<B>::ecdsa_sign(key, message, r, s)
+                board::crypto::P256::<B>::ecdsa_sign(key, message, r, s)?
             }
             #[cfg(feature = "board-api-crypto-p384")]
             Curve::P384 => {
@@ -173,7 +173,7 @@ fn ecdsa_sign<B: Board>(mut call: SchedulerCall<B, api::ecdsa_sign::Sig>) {
                 let message = memory.get_array::<48>(*message)?.into();
                 let r = memory.get_array_mut::<48>(*r)?.into();
                 let s = memory.get_array_mut::<48>(*s)?.into();
-                board::crypto::P384::<B>::ecdsa_sign(key, message, r, s)
+                board::crypto::P384::<B>::ecdsa_sign(key, message, r, s)?
             }
             #[allow(unreachable_patterns)]
             _ => trap_use!(),
@@ -196,7 +196,7 @@ fn ecdsa_verify<B: Board>(mut call: SchedulerCall<B, api::ecdsa_verify::Sig>) {
                 let y = memory.get_array::<32>(*y)?.into();
                 let r = memory.get_array::<32>(*r)?.into();
                 let s = memory.get_array::<32>(*s)?.into();
-                board::crypto::P256::<B>::ecdsa_verify(message, x, y, r, s)
+                board::crypto::P256::<B>::ecdsa_verify(message, x, y, r, s)?
             }
             #[cfg(feature = "board-api-crypto-p384")]
             Curve::P384 => {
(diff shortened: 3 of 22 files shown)


crates/scheduler/src/failure.rs
```diff
@@ -0,0 +1,70 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use wasefire_error::Error;
+
+#[derive(Debug, PartialEq)]
+enum Failure {
+    Trap(Trap),
+    Error(Error),
+}
```
ia0, line 21: I don't think we need a new file just for those 3 items. The `Failure` type and its implementations can simply be in the `lib.rs` next to the `Trap` type. In particular we don't need tests. Those are just normal usage of the type within the rest of the crate, so the tests are redundant.

crates/scheduler/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Minor
 
+- Add `Failure` type
```
ia0, line 7: This is not a minor change. This is purely an implementation detail, so a patch change. And the change should be something along the lines of "Simplify try-blocks for API call replies with a common `Failure` type".

ia0 (changes requested):
Sorry, I probably used a misleading title in #463. This PR is not fixing the core idea of #463 but just introducing an unused private type, which is not useful on its own. Instead, for this PR to fix #463, it should:
- Introduce the `Failure` type and its 2 implementations at top-level (the enabler).
- Change the `SchedulerCall::reply()` function to take a `Result<_, Failure>` instead (the actual change).
- Fix any compilation error when calling `reply()` (the consequences).
- Improve the try-blocks that currently have to juggle to return an `Error` (the benefit). This juggling can probably be identified by explicit use of `Ok` and `Err` within or right after the try-block .

lukeyeh:
> Sorry, I probably used a misleading title in #463. This PR is not fixing the core idea of #463 but just introducing an unused private type, which is not useful on its own. Instead, for this PR to fix #463, it should:
> 
>     * Introduce the `Failure` type and its 2 implementations at top-level (the enabler).
> 
>     * Change the `SchedulerCall::reply()` function to take a `Result<_, Failure>` instead (the actual change).
> 
>     * Fix any compilation error when calling `reply()` (the consequences).
> 
>     * Improve the try-blocks that currently have to juggle to return an `Error` (the benefit). This juggling can probably be identified by explicit use of `Ok` and `Err` within or right after the try-block .

I mainly wanted to add tests because I wanted to have this PR just add the failure type and following PRs to change `ScheduelrCall::reply`'s signature and improve the `try` blocks. I didn't want to have to add a `#[allow(dead_code)]` marker, so added some tests. I moved the `Trap` and `Failure` back to `lib.rs` and left a TODO instead.

crates/scheduler/src/failure.rs
lukeyeh: moved `Trap` back into lib.rs and added the `Failure` enum there too.

crates/scheduler/CHANGELOG.md
lukeyeh: Moved to patch, but kept the description the same as I plan to do the other parts in following PRs. Please let me know if you would like me to do all of the steps in a single PR.

ia0: Yes, I think it should all be in a single PR, since the changes independently don't make sense.

lukeyeh: Hey @ia0, I changed `Scheduler::reply()` to take the new `Result<_, Failure>` type as an argument and fixed all the compiler errors.

crates/scheduler/src/call/platform/protocol.rs
```diff
@@ -40,17 +40,17 @@ fn read<B: Board>(mut call: SchedulerCall<B, api::read::Sig>) {
...
                 let len = value.len() as u32;
                 let ptr = memory.alloc(len, 1)?;
                 memory.get_mut(ptr, len)?.copy_from_slice(&value);
                 memory.get_mut(*ptr_ptr, 4)?.copy_from_slice(&ptr.to_le_bytes());
                 memory.get_mut(*len_ptr, 4)?.copy_from_slice(&len.to_le_bytes());
-                Ok(true)
+                true
             }
-            Err(e) => Err(e),
+            Err(e) => Err(e)?,
```
ia0, line 53: This is actually not needed anymore, the whole match can be simplified by using `?` on the match expression.

```rust
match scheduler.applet.get_request()? {
```

And then, we're simply matching on an `Option`.

crates/scheduler/src/call/platform/update.rs
```diff
@@ -57,9 +57,8 @@ fn metadata<B: Board>(mut call: SchedulerCall<B, api::metadata::Sig>) {
                 memory.get_mut(ptr, len)?.copy_from_slice(&metadata);
                 memory.get_mut(*ptr_ptr, 4)?.copy_from_slice(&ptr.to_le_bytes());
                 memory.get_mut(*len_ptr, 4)?.copy_from_slice(&len.to_le_bytes());
-                Ok(())
             }
-            Err(error) => Err(error),
+            Err(error) => Err(error)?,
```
ia0, line 61: Same here, we can get rid of this branch, and actually the whole match by having:

```rust
let metadata = board::platform::Update::<B>::metadata()?;
```

crates/scheduler/src/call/platform/update.rs
```diff
@@ -86,14 +85,13 @@ fn process_<B: Board>(mut call: SchedulerCall<B, api::process::Sig>) {
...
     };
     call.reply(result);
 }
 
 #[cfg(feature = "board-api-platform-update")]
 fn finalize<B: Board>(call: SchedulerCall<B, api::finalize::Sig>) {
     let api::finalize::Params {} = call.read();
-    let result = try { board::platform::Update::<B>::finalize() };
-    call.reply(result);
+    call.reply(board::platform::Update::<B>::finalize().map_err(|x| x.into()));
```
ia0, line 96: nit: this could be simplified to
```suggestion
    call.reply(try { board::platform::Update::<B>::finalize()? });
```

This should work everywhere this pattern happens except when the never type `!` is used. I've created a compiler bug here: https://github.com/rust-lang/rust/issues/125364

crates/scheduler/src/call/store.rs
```diff
@@ -57,16 +57,16 @@ fn insert<B: Board>(mut call: SchedulerCall<B, api::insert::Sig>) {
...
     };
     call.reply(result);
 }
 
 #[cfg(feature = "board-api-storage")]
 fn remove<B: Board>(mut call: SchedulerCall<B, api::remove::Sig>) {
     let api::remove::Params { key } = call.read();
-    let res = call.scheduler().store.remove(*key as usize).map_err(convert);
-    call.reply(Ok(res));
+    let res = call.scheduler().store.remove(*key as usize).map_err(|x| convert(x).into());
```
ia0, line 68: nit: a try-block would help here
```suggestion
    let res = try { call.scheduler().store.remove(*key as usize).map_err(convert)? };
```

crates/scheduler/src/call/store.rs
```diff
@@ -76,16 +76,16 @@ fn find<B: Board>(mut call: SchedulerCall<B, api::find::Sig>) {
...
                 let len = value.len() as u32;
                 let ptr = memory.alloc(len, 1)?;
                 memory.get_mut(ptr, len)?.copy_from_slice(&value);
                 memory.get_mut(*ptr_ptr, 4)?.copy_from_slice(&ptr.to_le_bytes());
                 memory.get_mut(*len_ptr, 4)?.copy_from_slice(&len.to_le_bytes());
-                Ok(true)
+                true
             }
-            Err(e) => Err(convert(e)),
+            Err(e) => Err(convert(e))?,
```
ia0, line 88: this branch could be removed with `match scheduler.store.find(*key as usize).map_err(convert)? {`

crates/scheduler/src/call/store.rs
```diff
@@ -105,15 +105,15 @@ fn keys<B: Board>(mut call: SchedulerCall<B, api::keys::Sig>) {
...
             Ok(keys) => {
                 let len = keys.len() as u32;
                 let ptr = memory.alloc(2 * len, 2)?;
                 memory.get_mut(ptr, 2 * len)?.copy_from_slice(bytemuck::cast_slice(&keys));
                 memory.get_mut(*ptr_ptr, 4)?.copy_from_slice(&ptr.to_le_bytes());
-                Ok(len)
+                len
             }
-            Err(e) => Err(convert(e)),
+            Err(e) => Err(convert(e))?,
```
ia0, line 116: This could also be removed, in particular, the first try block to build keys could be removed, by using `map_err(convert)?` instead of `?`

crates/scheduler/src/call/store.rs
```diff
@@ -123,8 +123,8 @@ fn keys<B: Board>(mut call: SchedulerCall<B, api::keys::Sig>) {
 fn clear<B: Board>(mut call: SchedulerCall<B, api::clear::Sig>) {
     let api::clear::Params {} = call.read();
     let scheduler = call.scheduler();
-    let result = scheduler.store.clear(0).map_err(convert);
-    call.reply(Ok(result));
+    let result = scheduler.store.clear(0).map_err(|x| convert(x).into());
```
ia0, line 126: same here
```suggestion
    let result = try { scheduler.store.clear(0).map_err(convert)? };
```

ia0 (commented):
Thanks! This looks good. Most of the comments are the same and I did not mention it in every file, so you'll have to search for them.

lukeyeh:
Made some of the changes and applied changes to equivalent areas that you didn't leave comments PTAL

crates/scheduler/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### Patch
 
+- Add `Failure` type and change `Scheduler::reply()` to use it as a parameter.
```
ia0, line 17: nit
```suggestion
- Add `Failure` type to simplify try-blocks for `Scheduler::reply()` arguments
```

crates/scheduler/src/call.rs
```diff
@@ -52,7 +54,7 @@ macro_rules! or_false {
 macro_rules! trap_use {
     ($($x:ident),* $(,)?) => {{
         $( let _ = $x; )*
-        Err(crate::Trap)?
+        Err(crate::Failure::Trap(crate::Trap))?
```
ia0, line 57: I'm surprised this is needed. The conversion should be automatic and using `Err(crate::Trap)?` is more generic.

crates/scheduler/src/call.rs
```diff
@@ -113,5 +115,9 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 
 fn syscall<B: Board>(call: SchedulerCall<B, api::syscall::Sig>) {
     let api::syscall::Params { x1, x2, x3, x4 } = call.read();
-    call.reply(B::syscall(*x1, *x2, *x3, *x4).ok_or(Trap));
+    let result = match B::syscall(*x1, *x2, *x3, *x4) {
+        Some(result) => try { result? },
+        None => Err(Trap.into()),
+    };
+    call.reply(result);
```
ia0, line 122: nit: `try { x?? }` is actually the implementation of `Failure:join()` I mentioned before
```suggestion
    call.reply(try { B::syscall(*x1, *x2, *x3, *x4).ok_or(Trap)?? });
```

crates/scheduler/src/call/crypto/ec.rs
```diff
@@ -39,7 +39,7 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 
 fn is_supported<B: Board>(call: SchedulerCall<B, api::is_supported::Sig>) {
     let api::is_supported::Params { curve } = call.read();
-    call.reply(convert_curve::<B>(*curve).map(|x| Ok(x.is_ok())))
+    call.reply(try { convert_curve::<B>(*curve).map(|x| x.is_ok())? })
```
ia0, line 42: nit: by moving the `?` we can skip the `map()`
```suggestion
    call.reply(try { convert_curve::<B>(*curve)?.is_ok() })
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -18,12 +18,18 @@ use digest::{FixedOutput, InvalidLength, KeyInit, Output, Update};
...
+use wasefire_error::{Code, Error};
 
 use crate::applet::store::{MemoryApi, StoreApi};
 #[cfg(feature = "internal-hash-context")]
 use crate::applet::HashContext;
-use crate::{DispatchSchedulerCall, SchedulerCall, Trap};
+use crate::{DispatchSchedulerCall, Failure, SchedulerCall, Trap};
+
+impl From<InvalidLength> for Failure {
+    fn from(_value: InvalidLength) -> Self {
```
ia0, line 29: nit: to prove we are not hiding/discarding information in this conversion
```suggestion
    fn from(InvalidLength: InvalidLength) -> Self {
```

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -53,7 +59,7 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 #[cfg(feature = "applet-api-crypto-hash")]
 fn is_supported<B: Board>(call: SchedulerCall<B, api::is_supported::Sig>) {
     let api::is_supported::Params { algorithm } = call.read();
-    call.reply(convert_hash_algorithm::<B>(*algorithm).map(|x| Ok(x.is_ok())))
+    call.reply(try { convert_hash_algorithm::<B>(*algorithm).map(|x| x.is_ok())? })
```
ia0, line 62: same trick here (moving `?` up to avoid `map()`)

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -154,13 +154,13 @@ fn hmac_update<B: Board>(mut call: SchedulerCall<B, api::hmac_update::Sig>) {
     let api::hmac_update::Params { id, data, length } = call.read();
     let scheduler = call.scheduler();
     let memory = scheduler.applet.store.memory();
-    let result: Result<Result<(), _>, _> = try {
+    let result: Result<(), Failure> = try {
```
ia0, line 157: the whole type annotation can disappear
```suggestion
    let result = try {
```

crates/scheduler/src/call/timer.rs
```diff
@@ -39,7 +39,7 @@ pub fn process<B: Board>(call: Api<DispatchSchedulerCall<B>>) {
 #[cfg(not(feature = "board-api-timer"))]
 fn allocate<B: Board>(call: SchedulerCall<B, api::allocate::Sig>) {
     use wasefire_error::{Code, Error};
-    call.reply_(Ok(Err(Error::world(Code::NotEnough))))
+    call.reply_(Err(crate::Failure::Error(Error::world(Code::NotEnough))))
```
ia0, line 42: fyi: using `.into()` would have been fine too (in case you wondered)

crates/scheduler/src/call/uart.rs
```diff
@@ -87,7 +87,7 @@ fn read<B: Board>(mut call: SchedulerCall<B, api::read::Sig>) {
     let result = try {
         let uart = Id::new(*uart as usize).map_err(|_| Trap)?;
         let output = memory.get_mut(*ptr, *len)?;
-        board::Uart::<B>::read(uart, output).map(|x| x as u32)
+        board::Uart::<B>::read(uart, output).map(|x| x as u32)?
```
ia0, line 90: we can also do the trick here (and in the rest of the file)
```suggestion
        board::Uart::<B>::read(uart, output)? as u32
```

crates/scheduler/src/call/usb/serial.rs
```diff
@@ -46,7 +46,7 @@ fn read<B: Board>(mut call: SchedulerCall<B, api::read::Sig>) {
     let memory = scheduler.applet.memory();
     let result = try {
         let output = memory.get_mut(*ptr, *len)?;
-        board::usb::Serial::<B>::read(output).map(|x| x as u32)
+        board::usb::Serial::<B>::read(output).map(|x| x as u32)?
```
ia0, line 49: same in this file

lukeyeh:
went through the new pass of comments with the new pattern and other various nits

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -123,7 +126,7 @@ fn finalize<B: Board>(mut call: SchedulerCall<B, api::finalize::Sig>) {
 #[cfg(feature = "applet-api-crypto-hmac")]
 fn is_hmac_supported<B: Board>(call: SchedulerCall<B, api::is_hmac_supported::Sig>) {
     let api::is_hmac_supported::Params { algorithm } = call.read();
-    call.reply(convert_hmac_algorithm::<B>(*algorithm).map(|x| Ok(x.is_ok())))
+    call.reply(try { convert_hmac_algorithm::<B>(*algorithm).map(|x| x.is_ok())? })
```
ia0, line 129: that's an occurrence of `map()?`

crates/scheduler/src/call/crypto/hash.rs
```diff
@@ -195,7 +195,7 @@ fn hmac_finalize<B: Board>(mut call: SchedulerCall<B, api::hmac_finalize::Sig>)
 #[cfg(feature = "applet-api-crypto-hkdf")]
 fn is_hkdf_supported<B: Board>(call: SchedulerCall<B, api::is_hkdf_supported::Sig>) {
     let api::is_hkdf_supported::Params { algorithm } = call.read();
-    call.reply(convert_hmac_algorithm::<B>(*algorithm).map(|x| Ok(x.is_ok())))
+    call.reply(try { convert_hmac_algorithm::<B>(*algorithm).map(|x| x.is_ok())? })
```
ia0, line 198: and another one

ia0 (commented):
Thanks! There's 2 occurrences of the `map()?` that seems to have been missed, otherwise everything else is good to merge now.

ia0 (approved):
Thanks!

PR #466, opened 2024-05-13
Parsing and validation of atomic instructions.

resulknad:
Looking for early feedback on general direction, content and style. Thank you.


crates/interpreter/src/valid.rs
```diff
@@ -818,4 +893,10 @@ impl<'a, 'm> Expr<'a, 'm> {
         self.push(dst.into());
         Ok(())
     }
+    #[cfg(feature = "threads")]
+    fn check_aligned(&mut self, m: MemArg, n: usize) -> CheckResult {
+        let n_len: u32 = u32::try_from(n).unwrap();
+        check(u32::pow(2, m.align) == (n_len / 8))?;
```
ia0, line 899: nit
```suggestion
        check(1 << m.align == n / 8)?;
```

crates/interpreter/tests/spec.rs
```diff
@@ -24,7 +24,16 @@ use wast::token::Id;
 use wast::{parser, QuoteWat, Wast, WastArg, WastDirective, WastExecute, WastInvoke, WastRet, Wat};
 
 fn test(name: &str) {
-    let path = format!("../../third_party/WebAssembly/spec/test/core/{}.wast", name);
+    test_custom_path(name, "");
+}
+
+fn test_custom_path(name: &str, path: &str) {
```
ia0, line 30: I would instead add parameters to `test()` and update the `test!()` macro. I'll push something on main and then merge it into this branch.

ia0 (commented):
Thanks! Looks good so far. I didn't review in details though, I'll do so once it's closer to mergeable.

I'm not sure what's the best solution for the tests is. I think the simplest (I think that's what I did in my dev branch when I wrote the interpreter initially), is to add `#[ignore]` on the tests that fail, like `atomic`, and just remove the `#[ignore]` locally to know where you are, and remove it definitely once the full test passes. This way you don't need to modify the tests and can use them directly from the wasm/threads repo. You can always modify them there locally for testing purposes.

ia0:
I've merge #467 which let's you write `test!("threads", atomic);` or actually `test!(#[ignore] "threads", atomic);` initially.

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "third_party/SchemaStore/schemastore"]
 	path = third_party/SchemaStore/schemastore
 	url = https://github.com/SchemaStore/schemastore.git
+[submodule "third_party/WebAssembly/threads"]
```
ia0, line 10: By the way, feel free to do a first PR with just the submodule. You'll also have to update `scripts/sync.sh` and make sure the entry is sorted, but both of those are caught by tests.

crates/cli-tools/src/action.rs
```diff
@@ -106,6 +106,10 @@ impl RustAppletBuild {
         cargo.args(["rustc", "--lib"]);
         match &self.native {
             None => {
+                if package.features.contains_key("threads") {
+                    rustflags.push("-C target-feature=+atomics,+bulk-memory".to_owned());
```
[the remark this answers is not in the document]
resulknad: I'm not sure what the implications are exactly. I have created a CL however to see if anything breaks. Resolving for now.

crates/interpreter/Cargo.toml
```diff
@@ -23,12 +23,15 @@ lazy_static = "1.4.0"
 wast = "70.0.0"
 
 [features]
+default = ["threads"]
```
ia0, line 26: This shouldn't be enabled by default. Instead you can add new lines to `test.sh`. If this is for another reason that is not easy to fix, then let's add a TODO to remove it from default, because otherwise we are only testing with threads enabled.

ia0 (commented):
I didn't review in details because the PR is not up-to-date with the branch (let's merge the submodule PR first). Once done, I'll do a thorough review. But from a quick glance over the PR I'd say the overall approach is sound, so it's just about me checking the spec on one side and the PR on the other.

crates/interpreter/Cargo.toml
resulknad: Yeah this shouldn't be here, I only added it for local development as a hint to the language server.

crates/interpreter/src/valid.rs
```diff
@@ -645,6 +645,68 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                self.check_aligned(m, n.into())?;
+                self.store(NumType::i(n), n.into(), m)?;
+            }
+            #[cfg(feature = "threads")]
+            IAtomicStore_(b, m) => {
+                self.check_aligned(m, b.into())?;
+                self.store(NumType::i(b.into()), b.into(), m)?;
+            }
+            #[cfg(feature = "threads")]
+            AtomicOp(n, syntax::AtomicOp::Cmpxchg, m) => {
```
ia0, line 683: We import all the syntax in scope, so this shouldn't be needed
```suggestion
            AtomicOp(n, AtomicOp::Cmpxchg, m) => {
```

crates/interpreter/src/valid.rs
```diff
@@ -645,6 +645,68 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                let num = NumType::i(n);
+                self.pops([ValType::I32, num.into()][..].into())?;
+                self.push(num.into());
+            }
+            #[cfg(feature = "threads")]
+            AtomicOp_(b, _, _, m) => {
+                self.check_aligned(m, b.into())?;
+                let num = NumType::i(b.into());
+                self.pops([ValType::I32, num.into()][..].into())?;
+                self.push(NumType::i(b.into()).into())
```
ia0, line 708: nit
```suggestion
                self.push(num.into())
```

crates/interpreter/test.sh
```diff
@@ -28,3 +29,4 @@ cargo check --target=riscv32imc-unknown-none-elf \
 RUSTFLAGS=--cfg=portable_atomic_unsafe_assume_single_core \
   cargo check --target=riscv32imc-unknown-none-elf
 cargo test --features=debug,toctou,float-types,vector-types
+cargo test --features=debug,toctou,float-types,vector-types,threads
```
ia0, line 32: Let's always test with threads. Otherwise we should also test without floats.
```suggestion
cargo test --features=debug,toctou,float-types,vector-types,threads
```

ia0 (commented):
Thanks! It seems almost there.

crates/interpreter/src/parser.rs
```diff
@@ -272,6 +284,30 @@ impl<'m, M: Mode> Parser<'m, M> {
...
+            x @ 23 ..= 24 => Instr::AtomicStore((x - 23).into(), self.parse_memarg()?),
+            x @ 25 ..= 29 => Instr::AtomicStore_((x - 25).into(), self.parse_memarg()?),
+            x @ 30 ..= 78 => {
+                let op: AtomicOp = ((x - 30) / 7).into();
+                match (x - 30) % 7 {
+                    x @ 0 ..= 1 => Instr::AtomicOp(x.into(), op, self.parse_memarg()?),
+                    x => Instr::AtomicOp_((x - 2).into(), op, self.parse_memarg()?),
+                }
+            }
+            _x => M::unsupported(if_debug!(Unsupported::Opcode(_x)))?,
```
ia0, line 307: That's similar to the top-level byte. You can add your own debug-print for now with a TODO(threads), but those bytes are always invalid for the spec we target.
```suggestion
            _ => M::invalid()?,
```

crates/interpreter/src/syntax.rs
```diff
@@ -60,10 +60,28 @@ pub struct FuncType<'m> {
     pub results: ResultType<'m>,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Share {
```
ia0, line 64: I'm surprised you don't need to gate those too. I guess it's because they are eventually exported. I'm adding lints to catch those hopefully.

crates/interpreter/src/valid.rs
ia0: I see. We don't have this issue with `IBinOp` and the other ops because the behavior is always the same for all ops.

Let's keep it with `syntax::AtomicOp` then.

ia0 (commented):
Thanks!

crates/interpreter/src/syntax.rs
resulknad: So for my own understanding, you would expect rustc to complain about an unused enum? But because it's exported it cannot actually be sure about whether it is being used or not?

crates/interpreter/CHANGELOG.md
```diff
@@ -55,4 +55,4 @@
 
 ## 0.1.0
 
-<!-- Increment to skip CHANGELOG.md test: 4 -->
+<!-- Increment to skip CHANGELOG.md test: 5 -->
```
ia0, line 58: Sorry I missed that file. Actually this crate is published so you need to update the changelog. However, we could do it later, so let's keep it like that for now.

crates/interpreter/src/syntax.rs
ia0: Yes, the lint I'm thinking about is called `unreachable-pub`. It's when an item is `pub` but is actually not reachable from the public API. The `syntax` module is not public and we only re-export a subset of the stuff, so without the `threads` feature, the `Share` enum should be unreachable triggering a warning.

crates/scheduler/CHANGELOG.md
```diff
@@ -132,4 +132,4 @@
 
 ## 0.1.0
 
-<!-- Increment to skip CHANGELOG.md test: 35 -->
+<!-- Increment to skip CHANGELOG.md test: 36 -->
```
ia0, line 135: I don't understand this change, but it doesn't hurt.

ia0 (approved):
Thanks! Looks good enough to merge. The details can be handled when we'll merge this branch into main.

ia0:
The changelog test is a bit noisy for a dev branch, let me remove it.

ia0:
I've resolved the comments, but 2 things you could do in `main` are:
- Always enable the mutable globals since we always support them.
- Move the FC parsing to a `parse_instr_fc` similar to the `_fe` you just did.

PR #468, opened 2024-05-15
Use ubuntu repository for markdownlint

ia0:
Now that `ubuntu-latest` is Ubuntu 24.04, see https://github.com/actions/runner-images/issues/9691


ia0:
Ubuntu 24.04 is not yet `ubuntu-latest`. It's only available as `ubuntu-24.04`. This PR confirms that it would work. But let's wait until Ubuntu 24.04 is GA to avoid switch to `ubuntu-24.04` explicitly.

Issue #423, opened 2024-04-04
Support WebAssembly threads proposal


ia0:
The [WebAssembly Threads](https://github.com/WebAssembly/threads) proposal is close to being standardized (in final phase) and probably stable enough to implement. This will enable #30 and may need some amount of decision from #58.

PR #475, opened 2024-05-17
Adding wasm/threads submodule.

resulknad:
Moving from #466 into its own PR. #423


ia0 (approved):
Thanks!

The sort behavior was intended (i.e. using ASCII order), but I guess you have a different LC_COLLATE. I made the script more robust.

ia0 (approved):

PR #476, opened 2024-05-18
Enable +bulk-memory for wasm by default.

resulknad:
As discussed in #466 , let's enable bulk-memory operations by default.

### DIFF
--- a/crates/cli-tools/src/action.rs
+++ b/crates/cli-tools/src/action.rs
@@ -107,6 +107,7 @@ impl RustAppletBuild {
         match &self.native {
             None => {
                 rustflags.push(format!("-C link-arg=-zstack-size={}", self.stack_size));
+                rustflags.push("-C target-feature=+bulk-memory".to_string());
                 cargo.args(["--crate-type=cdylib", "--target=wasm32-unknown-unknown"]);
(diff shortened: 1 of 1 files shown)


ia0 (dismissed):
Thanks!

You'll need to increment [this](https://github.com/google/wasefire/blob/ac32ec709398c325c0d0293e84f85345123e3041/crates/cli-tools/CHANGELOG.md?plain=1#L5) number from 2 to 3 to disable the CHANGELOG test. (If this crate was published, then a proper entry in the changelog would have been needed. But since the crate is still 0.1.0-git, no need to track changes yet.)

resulknad:
> Thanks!
> 
> You'll need to increment [this](https://github.com/google/wasefire/blob/ac32ec709398c325c0d0293e84f85345123e3041/crates/cli-tools/CHANGELOG.md?plain=1#L5) number from 2 to 3 to disable the CHANGELOG test. (If this crate was published, then a proper entry in the changelog would have been needed. But since the crate is still 0.1.0-git, no need to track changes yet.)

Got it, incremented the version and the test is passing now. Thank you.

ia0 (approved):

PR #513, opened 2024-06-12
Execution of wasm atomic ops with portable_atomics in the interpreter.

resulknad:
Execution up to the point that `atomic.wast` test passes. Notably that means notify, wait and fence are nor properly implemented yet, but everything else is. For the last three some basic checks are implemented though, like out of bounds and memory alignment.

Also changed the submodule to the feature branch with the fixed tests.

### DIFF
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -1421,7 +1421,7 @@ impl<'m> Thread<'m> {
             return Err(trap());
         }
         self.push_value(count);
-        return Ok(());
+        Ok(())
     }
 
     #[cfg(feature = "threads")]
@@ -1438,7 +1438,7 @@ impl<'m> Thread<'m> {
         if read != expected {
             self.push_value(Val::I32(1));
         }
-        return Ok(());
+        Ok(())
     }
 
     #[cfg(feature = "threads")]
@@ -1532,21 +1532,17 @@ impl<'m> Thread<'m> {
                 c.unwrap_i64() as i64,
                 c2.unwrap_i64() as i64,
             ) as u64),
-            (NumType::I32, 8) => Val::I32(op.u32_u8(
-                mem.as_mut_ptr() as *mut u8,
-                c.unwrap_i32() as u8,
-                c2.unwrap_i32() as u8,
-            )),
+            (NumType::I32, 8) => {
+                Val::I32(op.u32_u8(mem.as_mut_ptr(), c.unwrap_i32() as u8, c2.unwrap_i32() as u8))
+            }
             (NumType::I32, 16) => Val::I32(op.u32_u16(
                 mem.as_mut_ptr() as *mut u16,
                 c.unwrap_i32() as u16,
                 c2.unwrap_i32() as u16,
             )),
-            (NumType::I64, 8) => Val::I64(op.u64_u8(
-                mem.as_mut_ptr() as *mut u8,
-                c.unwrap_i64() as u8,
-                c2.unwrap_i64() as u8,
-            )),
+            (NumType::I64, 8) => {
+                Val::I64(op.u64_u8(mem.as_mut_ptr(), c.unwrap_i64() as u8, c2.unwrap_i64() as u8))
+            }
             (NumType::I64, 16) => Val::I64(op.u64_u16(
                 mem.as_mut_ptr() as *mut u16,
                 c.unwrap_i64() as u16,
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -17,7 +17,23 @@ use alloc::vec;
 use alloc::vec::Vec;
 
 #[cfg(feature = "threads")]
-use portable_atomic::*;
+use portable_atomic::AtomicI16;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicI32;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicI64;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicI8;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicU16;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicU32;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicU64;
+#[cfg(feature = "threads")]
+use portable_atomic::AtomicU8;
+#[cfg(feature = "threads")]
+use portable_atomic::Ordering;
 
 use crate::error::*;
 use crate::module::*;
@@ -859,6 +875,7 @@ impl<'m> Thread<'m> {
             }
             MemorySize => self.push_value(Val::I32(store.mem(inst_id, 0).size())),
             MemoryGrow => {
+                // TODO(threads): Add mutex for growing shared memory.
                 let n = self.pop_value().unwrap_i32();
                 self.push_value(Val::I32(grow(store.mem(inst_id, 0), n, ())));
             }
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +548,46 @@ impl CvtOp {
     }
 }
 
+#[cfg(feature = "threads")]
+macro_rules! impl_atomic_op {
+    ($n:ident, $u:ident) => {
+        paste::paste! {
+            impl AtomicOp {
+                pub fn [< $n _ $u >](&self, mem:*mut $u, c:$u, c2:$u) -> $n {
```
ia0, line 556: It seems to me this could be simplified by returning `$u`, reducing the combinatorics explosion. The conversion from `$u` to `$n` can be done by the caller.

Also, maybe we should handle cmpxchg separately to reduce fake arguments.

ia0 (commented):
Quick review. I think the main issue is using `i` instead of `ea` to check alignment (but I might be wrong).

I also commented on notify/wait, but we can keep them for a later PR, when we'll extend what the shared memory is.

crates/interpreter/src/exec.rs
```diff
@@ -1423,6 +1406,158 @@ impl<'m> Thread<'m> {
...
+
+    #[cfg(feature = "threads")]
+    fn atomic_load(
+        &mut self, mem: &mut Memory<'m>, t: NumType, n: usize, m: MemArg,
+    ) -> Result<(), Error> {
+        let i = self.pop_value().unwrap_i32();
+        if i % (n as u32 / 8) != 0 {
+            return Err(trap());
+        }
+        let mem = match self.mem_slice(mem, m, i, n / 8) {
```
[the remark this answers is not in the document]
resulknad: Right, [`wasmtime` employs a RwLock](https://github.com/bytecodealliance/wasmtime/blob/a068bfe06435a048502bf4afc426451dab5294af/crates/wasmtime/src/runtime/vm/threads/shared_memory.rs#L13-L25), acquiring the readers lock for accessing size and the writers lock when growing.

IIUC from [the discussion](https://github.com/WebAssembly/threads/issues/26#issuecomment-433930711), an atomic load of the length and a mutex around grow should be sufficient as well, as you suggested. Added a ToDo for a mutex around grow.

crates/interpreter/src/exec.rs
```diff
@@ -1423,6 +1406,158 @@ impl<'m> Thread<'m> {
...
+
+    #[cfg(feature = "threads")]
+    fn atomic_notify(&mut self, _mem: &mut Memory<'m>, _m: MemArg) -> Result<(), Error> {
+        let count = self.pop_value();
+        let i = self.pop_value().unwrap_i32();
+
+        // Trap if memory access is OOB.
+        let _mem = match self.mem_slice(_mem, _m, i, 4) {
+            None => return Err(trap()),
+            Some(x) => x,
```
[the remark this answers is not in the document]
resulknad: Yes, we will certainly need some data for waiting threads. Maybe something like [parking_lot](https://docs.rs/parking_lot_core/latest/parking_lot_core/) would be handy. I have added a TODO for now.

resulknad:
Thank you for the review.

You are right about using `ea` when checking for alignment. Let's evolve the shared memory datastructure in a next PR, I left a todo in the code linking to your suggestion.

crates/interpreter/src/exec.rs
```diff
@@ -1423,6 +1430,185 @@ impl<'m> Thread<'m> {
...
+    ) -> Result<(), Error> {
+        let c = self.pop_value();
+        let c2 = if op == AtomicOp::Cmpxchg { self.pop_value() } else { c };
+        let i = self.pop_value().unwrap_i32();
+        let mem = match self.mem_slice(mem, m, i, n / 8, n as u32 / 8) {
+            None => return Err(trap()),
+            Some(x) => x,
+        };
+        let c = match (&op, t, n) {
+            (AtomicOp::Cmpxchg, NumType::I32, 32) => Val::I32(op.i32_binary(
```
ia0, line 1550: I suspect it might be worth splitting `AtomicOp` into `AtomicUnOp` and `AtomicBinOp` the same way non-atomic operations are. It will change a bit the syntax, but I don't think it should be too bad. Then it should even be possible to move the `fn $n(ptr: *mut $u, x: $u, y: $u)` definitions to the big `impl_op!` macro.

crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +558,50 @@ impl CvtOp {
...
 }
 
+#[cfg(feature = "threads")]
+macro_rules! impl_atomic_op {
+    ($u:ident) => {
+        paste::paste! {
+            impl AtomicOp {
+                pub fn [< $u >](&self, mem: *mut $u, c: $u) -> $u {
+                    match self {
+                        AtomicOp::Add => unsafe { [< Atomic $u:upper >]::from_ptr(mem).fetch_add(c, Ordering::SeqCst) }
```
ia0, line 568: Those `unsafe { [< Atomic $u:upper >]::from_ptr(mem) }` can be factored out of the match. Such that the match branches only have the relevant operation.

crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +558,50 @@ impl CvtOp {
     }
 }
 
+#[cfg(feature = "threads")]
+macro_rules! impl_atomic_op {
+    ($u:ident) => {
+        paste::paste! {
+            impl AtomicOp {
+                pub fn [< $u >](&self, mem: *mut $u, c: $u) -> $u {
```
ia0, line 566: nit: using paste seems unecessary here
```suggestion
                pub fn $u(&self, mem: *mut $u, c: $u) -> $u {
```
If/When moving this to the big `impl_op!` macro. I guess the function could be named `$n` instead.

ia0 (commented):
Thanks! I left some more comments.

crates/interpreter/src/syntax.rs
resulknad: Removed usage of paste for this. See comment in other thread, `impl_op!` and `impl_op!` are called for different types.

crates/interpreter/src/syntax.rs
resulknad: Much better, thanks.

crates/interpreter/src/exec.rs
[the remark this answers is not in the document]
resulknad: Ah yes, I'm mistaken. I somehow thought parking-lot was non-std and would allow you to pass in a synchronization primitive depending on the platform.

I will have to think about this a bit more and get back to you. Probably best to start simple though before generalizing further.

resulknad (commented):
Thanks. Not sure if it makes sense to move the atomic binary op macro into `impl_op!` - I left a comment.

crates/interpreter/src/exec.rs
```diff
@@ -1423,6 +1426,188 @@ impl<'m> Thread<'m> {
...
+            None => return Err(trap()),
+            Some(x) => x,
+        };
+
+        macro_rules! convert {
+            ($s:ident, $t:ident) => {{
+                let atomic_val;
+                unsafe {
+                    paste::paste! {
+                        atomic_val = [< Atomic $t:upper >]::from_ptr(mem.as_mut_ptr() as *mut $t);
```
ia0, line 1510: You can do the same change here as in `atomic_load()`.

crates/interpreter/src/exec.rs
```diff
@@ -1515,6 +1652,8 @@ impl<'m> Memory<'m> {
         self.data.fill(0);
         self.size = limits.min;
         self.max = limits.max;
+        #[cfg(feature = "threads")]
+        (self.share = limits.share == Share::Shared);
```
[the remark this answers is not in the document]
ia0: Sorry, I wrote it wrong. I meant `let () = ` not `let _ = `.

crates/interpreter/src/valid.rs
```diff
@@ -669,28 +669,28 @@ impl<'a, 'm> Expr<'a, 'm> {
...
-            AtomicOp(n, _, m) => {
+            AtomicUnOp(n, _, m) => {
                 self.check_mem(true, n.into(), m)?;
                 let num = NumType::i(n);
                 self.pops([ValType::I32, num.into()][..].into())?;
                 self.push(num.into());
             }
             #[cfg(feature = "threads")]
-            AtomicOp_(b, _, m) => {
+            AtomicUnOp_(b, _, m) => {
```
ia0, line 693: Now that we can match in the order we want, we can put UnOp back first again.

crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +557,47 @@ impl CvtOp {
...
+                }
+            }
+        }
+    };
+}
+
+#[cfg(feature = "threads")]
+impl_atomic_op!(i64);
+#[cfg(feature = "threads")]
+impl_atomic_op!(i32);
```
ia0, line 593: I don't think the spec uses signed integers for atomics, because all operations are sign-agnostics. I guess the confusion may come from the nomenclature between Rust and Wasm:
- `iNN` means signed integer in Rust but it means "no-signedness" integer in Wasm, see https://webassembly.github.io/spec/core/syntax/values.html#integers
- `uNN` means unsigned integer in both Rust and Wasm
- `sNN` means signed integer in Wasm

crates/interpreter/src/exec.rs
```diff
@@ -1423,6 +1426,188 @@ impl<'m> Thread<'m> {
...
+    ) -> Result<(), Error> {
+        let c = self.pop_value();
+        let i = self.pop_value().unwrap_i32();
+        let mem = match self.mem_slice(mem, m, i, n / 8, true) {
+            None => return Err(trap()),
+            Some(x) => x,
+        };
+        let c = match (t, n) {
+            (NumType::I32, 32) => {
+                Val::I32(op.i32(mem.as_mut_ptr() as *mut i32, c.unwrap_i32() as i32) as u32)
```
ia0, line 1544: Maybe after the sign issue is fixed, it would be possible to use a macro like:

```rust
($n:ident, $u:ident) => (op.$n(mem.as_mut_ptr() as *mut $u, c.unwrap_$n() as $u)
```
(probably using paste for the unwrap)

crates/interpreter/src/syntax.rs
ia0: Indeed, it doesn't work.

crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +557,47 @@ impl CvtOp {
     }
 }
 
+#[cfg(feature = "threads")]
+macro_rules! impl_atomic_op {
+    ($u:ident) => {
+        paste::paste! {
+            impl AtomicUnOp {
+                pub fn $u(&self, mem: *mut $u, c: $u) -> $u {
+                    let atomic_val = unsafe { [< Atomic $u:upper >]::from_ptr(mem) };
```
ia0, line 566: nit: can just use `x` as variable name. It's local and repetitive.

You can also rename `mem` to `ptr`.

crates/interpreter/src/exec.rs
resulknad: Yeah, looks great.

crates/interpreter/src/syntax.rs
resulknad: Yeah, thanks for the pointers. I was indeed thinking `iNN` is signed in Rust and wasm. I'm now using unsigned atomics everywhere.

IIUC this works because `iNN` in wasm can be signed or unsigned, but is in twos complement in the signed case so unsigned addition will work just fine. And we always zero-extend by definition.

crates/interpreter/src/syntax.rs
```diff
@@ -546,6 +557,45 @@ impl CvtOp {
...
+                        AtomicUnOp::Or => x.fetch_or(c, Ordering::SeqCst),
+                        AtomicUnOp::Xor => x.fetch_xor(c, Ordering::SeqCst),
+                        AtomicUnOp::Xchg => x.swap(c, Ordering::SeqCst),
+                    }
+                }
+            }
+            impl AtomicBinOp {
+                pub fn $u(&self, ptr: *mut $u, c: $u, c2: $u) -> $u {
+                    let x = unsafe { [< Atomic $u:upper >]::from_ptr(ptr) };
+                    match x.compare_exchange(c2, c, Ordering::SeqCst, Ordering::SeqCst) {
```
ia0: Why are the arguments swapped? This feels counter-intuitive and I'm not sure it's correct.

ia0 (approved):
Thanks! I've pushed a few style changes (plus the weird argument swap in atomic bin op). I didn't do a complete review of correctness. I'll do that when we'll merge this branch to main.

Issue #46, opened 2023-03-22
Interpreter performance and footprint


ia0:
The current interpreter is very simple:
- It doesn't pre-compute anything.
- It executes directly from flash.

It is thus an in-place (non-optimized) interpreter according to [A fast in-place interpreter for WebAssembly](https://arxiv.org/pdf/2205.01183.pdf) by Ben L. Titzer.

We would like to give users control over the tradeoff between performance and footprint (both in flash and memory). The following improvements would go towards this direction:
- Implement the optimizations discussed in the paper above, and feature-gate those that don't improve both performance and footprint (like the sidetable).
- Implement optimizations similar to what wasm3 does, by translating the opcodes into their function addresses, and feature-gate them.
- Implement a feature to persist in flash the translated code in the optimization above. This optimization means that a platform update would invalidate the applet (unless the original bytecode is preserved).
- JIT and AOT are probably out of the question for now. If wasmtime or wasmer end up supporting no-std, the situation might simplify. The difficulty being that they would need to run in a different thread to not take control of the scheduler thread.

zhouwfang:
@ia0 What is the expected timeline for #458? IIUC, it might make this issue irrelevant.

zhouwfang:
Thanks for the clarification!

zhouwfang:
@ia0 I was wondering if it is worth considering using `wasmi`. 

In https://google.github.io/wasefire/faq.html#why-implement-a-new-interpreter, it says "`wasmi` consumes too much RAM for embedded". However, in the recent release `wasmi` has been migrated from stack-based IR to register-based IR, and the release doc [writes](https://wasmi-labs.github.io/blog/posts/wasmi-v0.32/#memory-consumption), "The new register-based IR was carefully designed to enhance execution performance and to _minimize memory usage_. As the vast majority of a Wasm binary is comprised of encoded instructions, _this substantially decreases memory usage and enhances cache efficiency_ when executing Wasm through Wasmi. [...] with a fantastic startup performance and _low memory consumption especially suited for embedded environments_."

(Of course, we would still need to implement streamed compilation to flash.)

ia0:
> I was wondering if it is worth considering using `wasmi`.

It makes sense to give it a try if they claim being now suited for embedded environments. That would be one more comparison point along the following dimensions: interpreter code size, interpreter RAM usage, interpreter performance.

[...]

PR #514, opened 2024-06-25
Support tail call in the WASM interpreter

zhouwfang:
I'd just like to get some early feedback. I guess we probably want to preprocess/cache the detection of tail call, otherwise the overhead of `is_tail_call()` may be larger than the benefits. WDYT? Thanks.

#46


ia0:
Thanks for the PR, but this is implementing tail-call optimization for the WASM byte code. What we want is tail-call optimization for the WASM interpreter (thus TCO in Rust).

Note that there's a tail-call proposal for WASM that is almost finished: https://github.com/WebAssembly/tail-call. We might want to implement it at some point, but that's not a priority and not useful for #46.

I'm closing as this is not a direction we want to go now.

PR #520, opened 2024-06-27
Trampolining draft

zhouwfang:
I was wondering if such trampolining could be viewed as tail-call optimization in the interpreter. Thanks!

#46


ia0:
> I was wondering if such trampolining could be viewed as tail-call optimization in the interpreter.

Sadly no, this is the same issue as #514. Tail-call optimization in the interpreter would mean that the `loop` in `exec::Thread::run()` would need to disappear (and simply be the result of recursive tail-calls).

PR #522, opened 2024-07-03
Serve the new web client instead of the old one.

chris-dietz:
In addition ensure that trunk is installed and use trunk to build the web client from source automatically.


chris-dietz:
I think the CI is broken at the head of dev/web-ui. Not sure what the best way of fixing that is, I don't want to clutter up this PR pulling in commits from main to fix it. Do you have a way to rebase/merge the dev/web-ui branch from main? I don't think I have the ability to do that anyways.

crates/runner-host/Cargo.toml
```diff
@@ -20,6 +20,7 @@ wasefire-board-api = { path = "../board", features = ["std"] }
 wasefire-error = { path = "../error" }
 wasefire-interpreter = { path = "../interpreter", optional = true }
 wasefire-logger = { path = "../logger" }
+wasefire-cli-tools = {path = "../cli-tools"}
```
ia0, line 23: CI should complain about ordering and formatting here.

crates/runner-host/src/main.rs
```diff
@@ -79,7 +81,11 @@ async fn main() -> Result<()> {
                 }
             }
         });
+        ensure_trunk()?;
         let url = format!("{}:{}", flags.web_options.web_host, flags.web_options.web_port);
+        let mut build_command = Command::new("trunk");
```
ia0, line 86: We could/should use the wrapper script for reproducibility and to avoid contaminating the user environment. However, we might want to only use the wrapper script based on some flag in the future (once/if the host runner becomes an independent binary). Or even better, we should use trunk as a library if it's possible, such that the host runner is a self-contained binary.

For this PR, let's keep things like that. Just FYI and future steps.

crates/runner-host/src/main.rs
```diff
@@ -124,3 +130,17 @@ async fn main() -> Result<()> {
...
+
+fn ensure_trunk() -> Result<()> {
+    let trunk_installed = Command::new("which").args(["trunk"]).spawn()?.wait()?.success();
+    if trunk_installed {
+        return Ok(());
+    }
+
+    let mut cargo_command = Command::new("cargo");
+    // Uses trunk version 0.19.3 due to the latest not compiling with the rustc version used by wasefire.
+    // When/ if the rustc version gets updated this should be updated as well.
```
ia0, line 142: fyi: Using the wrapper script will/should take care of this.

ia0 (approved):
I've merged main into dev/web-ui and updated your branch. You should pull before fixing the 2 CI failures.

crates/runner-host/src/main.rs
chris-dietz: I changed it to used the wrapper script. Thanks for pointing that out. 

And I was hoping to use trunk as a library as well and I looked into it somewhat. From what I could find though there doesn't seem to be a publicly documented api for using trunk like that.

There is an internal module that the CLI uses but idk if there's any guarantees about the stability of that interface. 

See https://github.com/trunk-rs/trunk/blob/main/src/build.rs

If you think it would be worthwhile to look into that though I'd be happy to.

crates/runner-host/src/main.rs
chris-dietz: Per my other comment, changed to use the wrapper script.

crates/runner-host/Cargo.toml
chris-dietz: Fixed with formatter

crates/runner-host/src/main.rs
ia0: Thanks for looking into it. I'm still unsure whether it's worth it. It will be useful for development, but I intend to have the host runner as a standalone released binary (or maybe part of the wasefire CLI), in which case it doesn't really matter because users are not supposed to build the host runner anymore.

ia0 (commented):
Thansk! LGTM

scripts/wrapper.sh
```diff
@@ -58,6 +59,7 @@ case "$1" in
   rust-objcopy|rust-size) ensure_cargo cargo-binutils 0.3.6 ;;
   taplo) ensure_cargo taplo-cli 0.9.0 ;;
   twiggy) ensure_cargo twiggy 0.7.0 ;;
+  trunk) ensure_cargo trunk 0.19.3 ;;
```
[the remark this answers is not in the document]
ia0: I'm fixing this in the next commit.

ia0 (approved):
Thanks! I'm gonna test some applets and merge.

ia0:
LGTM, it's gonna take some time to pass the CI because it needs to build trunk. I hope it will get cached after the merge. I didn't test this with a development branch yet.

ia0:
I guess now only the esthetic aspects remain right? Or do you see other technical aspects?

scripts/wrapper.sh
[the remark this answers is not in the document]
chris-dietz: Thanks sorry I missed this comment earlier.

ia0:
Tests are a good idea, but as you mentioned, not a regression. Also testing UIs is usually much harder than testing logic, so I wouldn't worry about it, but feel free to explore this direction if it's something that interests you.

To merge this development branch in main, some CSS is probably enough indeed. Then there's also the automatic websocket reconnection when the server dies, but maybe a manual reload is fine for now (or just opening a new window).

chris-dietz:
> To merge this development branch in main, some CSS is probably enough indeed. Then there's also the automatic websocket reconnection when the server dies, but maybe a manual reload is fine for now (or just opening a new window).

Sg, also I believe it recovers from the server going down already. The thing that doesn't quite seem to work is the web server seems to not realize the client is already open and tries to open a new tab anyways. I'm not quite sure the reason why but if you comment out the lines opening the browser it recovers like it's supposed to. 

I'll have to look into that more closely.

ia0:
> Sg, also I believe it recovers from the server going down already. The thing that doesn't quite seem to work is the web server seems to not realize the client is already open and tries to open a new tab anyways. I'm not quite sure the reason why but if you comment out the lines opening the browser it recovers like it's supposed to.

This seems to be because the web-socket retries to connect every 3 seconds instead of every second. The web server waits 2 seconds before opening a new window, so there's some chances that the web client won't connect in that timeframe. Instead of using `use_websocket()` you need to use `use_websocket_with_options()` and set [`reconnect_interval`](https://docs.rs/yew-hooks/latest/yew_hooks/struct.UseWebSocketOptions.html#structfield.reconnect_interval) to 1000. This could be its own PR.

PR #524, opened 2024-07-06
Add wasmi for benchmark

zhouwfang:
#46


crates/wasm-bench/Cargo.toml
```diff
@@ -43,6 +49,13 @@ target-nordic = [
 ]
 target-riscv = ["dep:embedded-alloc", "dep:portable-atomic", "dep:riscv", "dep:riscv-rt"]
 
+# https://github.com/wasmi-labs/wasmi/blob/main/docs/usage.md#optimizations
+[profile.wasmi_optimization]
```
ia0: I think this makes sense to have it for all runtimes. I was considering doing it anyway before.

ia0 (approved):
Thanks for the PR!

PR #526, opened 2024-07-10
Add wasmtime for benchmark

zhouwfang:
#46


ia0 (approved):
Thanks! So it looks like Wasmtime is not yet no-std.

PR #537, opened 2024-07-22
Also run cargo publish with dry-run

ia0:


scripts/publish.sh
```diff
@@ -71,13 +82,18 @@ for crate in "${TOPOLOGICAL_ORDER[@]}"; do
...
+if [ $DRY_RUN = n ]; then
+  i "Remove all -git suffixes"
+  sed -i 's/-git//' $(git ls-files '*/'{Cargo.{toml,lock},CHANGELOG.md})
+  if [ -n "$(git status -s)" ]; then
+    i "Commit release"
+    git commit -aqm'Release all crates'
+    t "Create a PR from this release commit"
+    d "Re-run from the merged PR"
+  fi
+  i "This looks like a release commit"
```
ia0-review, line 94: We should check the description to make sure it ends with `(#nnn)`. This is enough to catch most re-run issues since we create the commit (and in particular its description).

ia0-review (approved):

PR #539, opened 2024-07-22
WIP Add ctap hid api

kofls:


kofls:
This is a draft PR for the CTAP HID api. I still have to implement the actual read/write using the usbd-ctaphid crate. Just wanted to get initial feedback on the approach.

I'm copying the read/write signatures we have in the usb-serial API currently. Do we also want to have the asynchronous reads/writes like we have in the usb serial api?

crates/board/src/usb/ctap.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn from(event: Event) -> Self {
+        super::Event::Ctap(event).into()
+    }
+}
+
+pub trait Api: Send {
+    /// Reads from the USB serial into a buffer.
+    ///
+    /// Returns the number of bytes read. It could be zero if there's nothing to read.
+    fn read(output: &mut [u8]) -> Result<usize, Error>;
```
ia0, line 46: Should this be `&mut [u8; 64]`? This is the maximum packet size in full speed and most probably what will always be used. It's definitely what's used by OpenSK. See https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#usb-message-and-packet-structure for reference.

crates/board/src/usb/ctap.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// Writes from a buffer to the USB serial.
+    ///
+    /// Returns the number of bytes written. It could be zero if the other side is not ready.
+    fn write(input: &[u8]) -> Result<usize, Error>;
```
ia0, line 51: Same here, I would use `&[u8; 64]`.

We could also make clear in those function documentation that this is about reading and writing _packets_, not messages.

crates/board/src/usb/ctap.rs
```diff
@@ -0,0 +1,85 @@
...
+    Write,
+}
+
+impl<B: crate::Api> From<Event> for crate::Event<B> {
+    fn from(event: Event) -> Self {
+        super::Event::Ctap(event).into()
+    }
+}
+
+pub trait Api: Send {
```
ia0, line 42: We should add the event support with `enable()` and `disable()`. This is needed by OpenSK.

crates/board/src/usb/ctap.rs
```diff
@@ -0,0 +1,85 @@
...
+pub trait HasCtapHid: Send {
+    type UsbBus: UsbBus;
+
+    fn with_ctaphid<R>(f: impl FnOnce(&mut CtapHid<Self::UsbBus>) -> R) -> R;
+}
+
+/// Wrapper type for boards using the `usbd_ctaphid` crate.
+pub struct WithCtapHid<T: HasCtapHid> {
+    _never: !,
+    _has_serial: T,
```
ia0, line 63: nit: rename

crates/board/src/usb/ctap.rs
```diff
@@ -0,0 +1,85 @@
...
+/// Helper struct for boards using the `usbd_ctaphid` crate.
+pub struct CtapHid<'a, T: UsbBus> {
+    ctap_hid: usbd_ctaphid::CtapHid,
+    read_enabled: bool,
+    write_enabled: bool,
+}
+
+impl<T: HasCtapHid> Api for WithCtapHid<T> {
+    fn read(output: &mut [u8]) -> Result<usize, Error> {
+        match T::with_ctaphid(|ctap_hid| ctap_hid.pipe.read_and_handle_packet()) {
```
ia0, line 75: I have a hard time understanding how to use the `usbd_ctaphid` crate or if it is at all usable yet. Curious to see how this will look like once the PR is ready.

crates/prelude/src/ctap.rs
```diff
@@ -0,0 +1,372 @@
...
+///
+/// This trait should only be implemented by the prelude and is thus sealed. Its purpose is to
+/// provide a unique interface to the different CtapHids.
+#[sealed(pub(crate))]
+pub trait CtapHid {
+    /// Reads from the CtapHid into a buffer without blocking.
+    ///
+    /// Returns how many bytes were read (and thus written to the buffer). This function does not
+    /// block, so if there are no data available for read, zero is returned.
+    fn read(&self, buffer: &mut [u8]) -> Result<usize, Error>;
```
ia0, line 49: Same as in the board. I would fix the size to 64 bytes.

crates/prelude/src/lib.rs
```diff
@@ -51,6 +51,8 @@ pub mod button;
 mod callback;
 #[cfg(feature = "internal-api-crypto")]
 pub mod crypto;
+// TODO: Add feature
+pub mod ctap;
```
ia0, line 55: This should be under `usb` like `serial` (except it doesn't use the generic `Serial` API).

crates/prelude/src/usb/ctap.rs
```diff
@@ -0,0 +1,65 @@
...
+
+// use crate::ctap::Event;
+use crate::{convert, Error};
+// use crate::convert_unit;
+
+/// Implements the [`CtapHid`](crate::ctap::CtapHid) interface for USB transport.
+pub struct UsbCtapHid;
+
+#[sealed]
+impl crate::ctap::CtapHid for UsbCtapHid {
```
ia0, line 28: You actually don't need the `crate::ctap::CtapHid` abstraction. We can add it if we ever have 2 implementations. For now let's just directly implement free functions in this module.

ia0 (commented):
Thanks! And to answer your question: Yes, we should support asynchronous API.

crates/board/src/usb/ctap.rs
kofls: Done.

crates/board/src/usb/ctap.rs
kofls: Done.

crates/board/src/usb/ctap.rs
kofls: I wasn't able to figure out how to use this crate. Changed to use `ctap-hid-fido2` instead

crates/prelude/src/lib.rs
kofls: Done.

crates/prelude/src/usb/ctap.rs
kofls: Done.

crates/board/Cargo.toml
```diff
@@ -21,6 +21,7 @@ aes-gcm = { version = "0.10.3", default-features = false, optional = true }
 bytemuck = { version = "1.16.0", default-features = false, optional = true }
 ccm = { version = "0.5.0", default-features = false, optional = true }
 crypto-common = { version = "0.1.6", default-features = false, optional = true }
+ctap-hid-fido2 = { version = "3.5.1", default-features = false, optional = true }
```
ia0, line 24: This crate does not do what you think it does. It implements the client part of CTAP2, not the authenticator part. I don't think there are any crates. You might have to implement something yourself. You can implement a `UsbClass` directly like https://github.com/google/wasefire/blob/main/crates/protocol-usb/src/device.rs or use https://crates.io/crates/usbd-human-interface-device which might work.

ia0 (commented):
The PR is out-of-date, there was some merged changes from main. You'll need to merge into dev/opensk. Also, it would probably be good to start using the real CI. You can try to `git checkout main -- .github/workflows/ci.yml scripts/ci.sh` and fix any issues when running `./scripts/ci.sh`.

crates/board/Cargo.toml
kofls: I didn't realize that, thank you! I'll re-implement this.

ia0: No worries, I agree this is quite confusing. Ideally, this should be caught by CI, but due to the conflicts it didn't run. (But maybe because it's WASM, the fact that that crate was not no-std didn't fail.)

ia0:
ccing @kaczmarczyck who might also have ideas on how to implement the authenticator side of CTAP2

kaczmarczyck:
> ccing @kaczmarczyck who might also have ideas on how to implement the authenticator side of CTAP2

I assume you mean CTAP HID instead of CTAP2? And you need descriptors etc? OpenSK has this for Tock:

https://github.com/google/OpenSK/blob/2.1/patches/tock/03-add-ctap-modules.patch

PR #540, opened 2024-07-23
[Draft] Add side table

zhouwfang:
Just wanted to gather some initial feedback. Thanks!

#46


crates/interpreter/src/valid.rs
```diff
@@ -16,6 +16,7 @@ use alloc::collections::BTreeSet;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
+use std::collections::HashMap;
```
zhouwfang, line 19: Do you have a recommendation for the alternative in `no_std`?

zhouwfang:
Wanted to work more before review.

PR #549, opened 2024-07-26
Added CSS styling and fixed reconnecting to backend without reload.

chris-dietz:
- Added CSS to match existing web client look.
- Changed timeout on websockets to 1 second (1000 ms)
- Also made the client reset and disable input when the websocket connection goes down.

Should be the final bit that needs to be done to replace the old web client.


ia0 (commented):
Thanks! It looks pretty good to final merge.

The only issue I have is that we shouldn't clear the logs when disconnecting (just print that we disconnected), and only clear when connecting. In particular I'd like the `applet rust panic` example to actually show the applet message and the panic message. I tried to remove the `history.clear()` but this doesn't seem to be enough. There's maybe some timing issue. When I add `timer::sleep_ms(2000)` before the `panic!("ready to die")` in `examples/rust/panic/src/lib.rs` then it works (the history is present after disconnect).

One way to fix this would be for the host runner to stay alive, but it could die for other reasons, so it would be nice if the log is not cleared when disconnected.

ia0:
By the way, I've fixed an issue with `xtask` not using the correct applet. You should merge `dev/web-ui` in your branch before testing switching between `hello`, `panic`, and `button_abort` examples.

ia0:
And once this PR is merged, you can create a PR from dev/web-ui to main with https://github.com/google/wasefire/compare/dev/web-ui?expand=1.

crates/runner-host/crates/web-client/src/app.rs
```diff
@@ -60,6 +60,11 @@ pub fn app() -> Html {
 
     html! {
         <main>
+        <object
```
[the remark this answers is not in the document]
chris-dietz: Fixed

crates/runner-host/crates/web-client/src/console.rs
```diff
@@ -50,11 +51,26 @@ pub fn console(Props { id, command_state, on_new_console_msg }: &Props) -> Html
...
                 info!("Command: {command:?}");
-                if let Command::Log { message } = command {
-                    history.push(format!("[recv]: {message}"));
+
+                match command {
+                    Command::Log { message } => {
+                        history.push(format!("[recv]: {message}"));
+                    }
+                    Command::Disconnected => {
+                        history.clear();
```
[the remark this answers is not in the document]
chris-dietz: Makes sense I removed clearing on disconnect.

crates/runner-host/crates/web-client/src/hooks.rs
```diff
@@ -37,23 +37,50 @@ impl UseRunnerConnectionHandle {
...
-    use_effect_with(ws.message.clone(), {
-        let command_state = command_state.clone();
-        move |message| {
-            if let Some(message) = &**message {
-                info!("Message: {message}");
-                match serde_json::from_str::<Command>(message) {
-                    Ok(command) => command_state.set(Some(command)),
-                    Err(err) => warn!("Error parsing message: {err}"),
+    let ws = use_websocket_with_options(
+        backend_address.clone(),
```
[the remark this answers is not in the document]
chris-dietz: We don't, good point! Fixed

chris-dietz:
> Thanks! It looks pretty good to final merge.
> 
> The only issue I have is that we shouldn't clear the logs when disconnecting (just print that we disconnected), and only clear when connecting. In particular I'd like the `applet rust panic` example to actually show the applet message and the panic message. I tried to remove the `history.clear()` but this doesn't seem to be enough. There's maybe some timing issue. When I add `timer::sleep_ms(2000)` before the `panic!("ready to die")` in `examples/rust/panic/src/lib.rs` then it works (the history is present after disconnect).
> 
> One way to fix this would be for the host runner to stay alive, but it could die for other reasons, so it would be nice if the log is not cleared when disconnected.

Huh, I'm having trouble replicating that on my end. It seems to keep the history when I run the panic applet now. I was having a issue where the server would relaunch the web app even though it was already open sometimes. I fixed it by increasing the time the server waits for the client to connect. (Changing the retry interval on the client didn't have an effect, I think there's just too much overhead for it to reconnect within 2 seconds consistently.)

Maybe that fixed the problem you were having as well? Could you try again and see if it's still happening?

chris-dietz:
Also I made trunk build the web client in "release" mode now. Hopefully that will make things faster and possibly fix some timing issues.

crates/runner-host/crates/web-client/src/console.rs
```diff
@@ -63,11 +77,12 @@ pub fn console(Props { id, command_state, on_new_console_msg }: &Props) -> Html
...
-            <p><b>{ "Log history: " }</b></p>
-            { for history.current().iter().map(|message| html!(<p>{ message }</p>)) }
-            <form onsubmit={onsubmit}>
+            <div class="console-display">
+                { for history.current().iter().rev().map(|message| html!(<div>{ message }</div>)) }
+            </div>
+            <form class="console-form" onsubmit={onsubmit}>
                 <input ref={console_ref} type="text" id="consolein" />
-                <input  type="submit" value="Send"/>
+                <input disabled={!*button_enabled}  type="submit" value="Send"/>
```
ia0, line 85: nit

```suggestion
                <input disabled={!*button_enabled}  type="submit" value="Send" />
```

crates/runner-host/crates/web-client/src/hooks.rs
```diff
@@ -37,23 +37,50 @@ impl UseRunnerConnectionHandle {
     }
 }
 
+fn set_disconnected(command_state: UseStateHandle<Option<Command>>) {
+    if let Some(command) = &*command_state {
+        if *command != Command::Disconnected {
+            command_state.set(Some(Command::Disconnected))
+        }
+    }
```
ia0, line 45: nit
```suggestion
    match &*command_state {
        None | Some(Command::Disconnected) => (),
        _ => command_state.set(Some(Command::Disconnected)),
    }
```

crates/runner-host/crates/web-client/data/index.scss
```diff
@@ -23,10 +23,49 @@ svg {
 }
 
 .button {
```
ia0, line 25: Is it possible to add some margin (or any other CSS technique) to get the buttons and LEDs horizontally centered in those 79 pixels of the left panel?

ia0 (commented):
LGTM thanks!

Regarding the 2 to 3 seconds change, can you try again with 2 seconds? On my side I didn't get any issue. And I would prefer to reduce the starting time (which affects all users all the time) than increase it to fix possible reopens (which affects less users some of the time).

If trunk doesn't build in release by default, it's definitely needed to do that. And speaking of release, maybe there's too much logging to the console. Instead of use `wasm_logger::Config::default()` we should use `wasm_logger::Config::new(level)` with appropriate level. This probably needs to be done at compile-time using features (one per `log::Level`) and set in the host runner when building trunk using `log::max_level().to_level()` to know which feature to use. If no feature is passed (i.e. when `to_level()` returns `None`), then logging should be completely disabled. But this should be done in a separate PR.

crates/runner-host/crates/web-client/data/index.scss
chris-dietz: Definitely, should be centered now.

chris-dietz:
> LGTM thanks!
> 
> Regarding the 2 to 3 seconds change, can you try again with 2 seconds? On my side I didn't get any issue. And I would prefer to reduce the starting time (which affects all users all the time) than increase it to fix possible reopens (which affects less users some of the time).
> 
> If trunk doesn't build in release by default, it's definitely needed to do that. And speaking of release, maybe there's too much logging to the console. Instead of use `wasm_logger::Config::default()` we should use `wasm_logger::Config::new(level)` with appropriate level. This probably needs to be done at compile-time using features (one per `log::Level`) and set in the host runner when building trunk using `log::max_level().to_level()` to know which feature to use. If no feature is passed (i.e. when `to_level()` returns `None`), then logging should be completely disabled. But this should be done in a separate PR.

I tried it again and it doesn't seem to be an issue anymore. So I reverted to 2 seconds. 

And yeah I agree, I'll set the logging level based on compilation flags. I'll create a follow up PR doing that. :)

ia0 (approved):
Thanks!

Once merged, do you mind [creating a PR](https://github.com/google/wasefire/compare/main...dev/web-ui?expand=1) from `dev/web-ui` to `main`? This way you will be the main author. We can also decide to make you code owner of the `dev/web-ui` if that's a role you're willing to take.

The PR for logging can still be based on `dev/web-ui` and updated to `main` once the `dev/web-ui` is merged in `main`.

chris-dietz:
> Thanks!
> 
> Once merged, do you mind [creating a PR](https://github.com/google/wasefire/compare/main...dev/web-ui?expand=1) from `dev/web-ui` to `main`? This way you will be the main author. We can also decide to make you code owner of the `dev/web-ui` if that's a role you're willing to take.
> 
> The PR for logging can still be based on `dev/web-ui` and updated to `main` once the `dev/web-ui` is merged in `main`.

PR is is created. :) 

And yeah I would have no problems with taking ownership over the 'dev/web-ui' branch. ^^

PR #551, opened 2024-07-29
Side table interface design

zhouwfang:
Would like to get feedback on the side table interface design before diving into the implementation details. Thanks!

#46


crates/interpreter/src/module.rs
```diff
@@ -19,14 +19,14 @@ use crate::cache::Cache;
...
-use crate::valid::validate;
+use crate::valid::{validate, FuncInfo};
 use crate::*;
 
 /// Valid module.
 #[derive(Debug)]
 pub struct Module<'m> {
     binary: &'m [u8],
-    types: Vec<FuncType<'m>>,
+    functions_info: Vec<FuncInfo<'m>>,
```
ia0, line 29: You want a new field for that (see the comment in the other file).

ia0 (commented):
Thanks! The side table design looks fine so far (except for the confusion between `types` and `funcs`). For now let's ignore the fact that we want to persist to flash. It will complicate more than needed.

I've merged `dev/wasm-bench` into main and created `dev/fast-interp` for this work instead. You may need to pull to take this merge into account.

PR #555, opened 2024-07-31
Store all source files under /src/

ia0:
In particular, this is an alternative solution to #545.


crates/scheduler/Cargo.toml
```diff
@@ -7,7 +7,8 @@ publish = true
 edition = "2021"
 description = "Wasefire scheduler"
 repository = "https://github.com/google/wasefire"
-include = ["/LICENSE", "/build.rs", "/src/"]
+include = ["/LICENSE", "/src/"]
+build = "src/build.rs"
```
ia0-review, line 11: Did you try `/src/build.rs`? Not sure why `include` has it, but better be explicit if it works.

ia0-review (approved):

crates/scheduler/Cargo.toml
ia0-review: Just tried and it doesn't work. The [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-build-field) is also clear that the file is search at the package root.

PR #557, opened 2024-08-01
New frontend web client

### LINKED ISSUE #261: Write the web client of the host runner in Rust
The web client is currently written in Javascript. This means duplication of the transport format (besides the issues specific to Javascript as a language). By writing the web client in Rust compiled to Wasm, the `web-common` crate can be used by both the client and server, enforcing a single source of truth for the transport format.

Tasks:
- [ ] Make sure `trunk build` is called during build (worst case using a `build.rs` script).
- [ ] Move all assets from web-server to web-client.
- [ ] Update README once feature complete.

chris-dietz:
- New client is written using rust and compiled to web asm using trunk. 
- See #459 #522 and #549 for more details. 
fixes #261


ia0 (approved):
Thanks! I'll add @invernizzi to review the `*.scss` and `*.html` files and possibly do an end-to-end test too.

ia0:
Hey @invernizzi , could you take a look at the `*.scss` and `*.html` file? It looks good to me, but I don't know what idiomatic SCSS and HTML should look like. You can also do an end-to-end test with the following commands:

```shell
cargo xtask applet rust hello runner host --no-default-features --web --log=info
cargo xtask applet rust button_abort runner host --no-default-features --web --log=info
cargo xtask applet rust panic runner host --no-default-features --web --log=info
```

ia0:
Actually, let me merge this because I'm going to do some changes that conflict, and the PR is good enough to merge, we can always address comments in follow-up PRs (like for the verbose logging).

ia0-review (approved):

PR #563, opened 2024-08-04
Add SideTableEntry

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -47,7 +47,11 @@ mod tests {
     use super::*;
     use crate::Error;
 
+    const PROPTEST_CASES: u32 = 1;
+
     proptest! {
+        #![proptest_config(ProptestConfig::with_cases(PROPTEST_CASES))]
+
         #[test]
         fn signed_field_round_trip(expected in -8i32..8) {
             let mask = 0b1111000;
@@ -55,19 +59,13 @@ mod tests {
             prop_assert_eq!(actual, expected);
         }
 
-       #[test]
-        fn signed_field_error_positive(expected in prop_oneof![
-            8..100,
-            100..,
-        ]) {
+        #[test]
+        fn signed_field_error_positive(expected in 8..100) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
         }
 
         #[test]
-        fn signed_field_error_negative(expected in prop_oneof![
-            -100..-9,
-            ..-100,
-        ]) {
+        fn signed_field_error_negative(expected in -100..-8) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
         }
 
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -15,15 +15,14 @@
 use crate::error::*;
 
 pub fn into_signed_field(mask: u32, value: i32) -> Result<u32, Error> {
-    let value = value.wrapping_add(offset(mask)) as u32;
-    into_field(mask, value)
+    into_field(mask, value.wrapping_add(offset(mask)) as u32)
 }
 
 pub fn from_signed_field(mask: u32, field: u32) -> i32 {
     from_field(mask, field) as i32 - offset(mask)
 }
 
-pub fn offset(mask: u32) -> i32 {
+fn offset(mask: u32) -> i32 {
     1 << (mask.count_ones() - 1)
 }
 
@@ -56,13 +55,21 @@ mod tests {
             prop_assert_eq!(actual, expected);
         }
 
-        #[test]
-        fn signed_field_error_positive(expected in 8i32..100) {
+       #[test]
+        fn signed_field_error_positive(expected in prop_oneof![
+            Just(8),
+            9..100,
+            100..,
+        ]) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
         }
 
         #[test]
-        fn signed_field_error_negative(expected in -100..-8i32) {
+        fn signed_field_error_negative(expected in prop_oneof![
+            Just(-9),
+            -100..-10,
+            ..-100,
+        ]) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
         }
 
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -1,0 +1,81 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::error::*;
+
+pub fn into_signed_field(mask: u32, value: i32) -> Result<u32, Error> {
+    let value = value.wrapping_add(offset(mask)) as u32;
+    into_field(mask, value)
+}
+
+pub fn from_signed_field(mask: u32, field: u32) -> i32 {
+    from_field(mask, field) as i32 - offset(mask)
+}
+
+pub fn offset(mask: u32) -> i32 {
+    1 << (mask.count_ones() - 1)
+}
+
+pub fn into_field(mask: u32, value: u32) -> Result<u32, Error> {
+    let field = (value << mask.trailing_zeros()) & mask;
+    if from_field(mask, field) != value {
+        return Err(unsupported(if_debug!(Unsupported::SideTable)));
+    }
+    Ok(field)
+}
+
+pub fn from_field(mask: u32, field: u32) -> u32 {
+    (field & mask) >> mask.trailing_zeros()
+}
+
+#[cfg(test)]
+mod tests {
+    use alloc::format;
+
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::Error;
+
+    proptest! {
+        #[test]
+        fn signed_field_round_trip(expected in -8i32..8) {
+            let mask = 0b1111000;
+            let actual = from_signed_field(mask, into_signed_field(mask, expected).unwrap());
+            prop_assert_eq!(actual, expected);
+        }
+
+        #[test]
+        fn signed_field_error_positive(expected in 8i32..100) {
+            prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
+        }
+
+        #[test]
+        fn signed_field_error_negative(expected in -100..-8i32) {
+            prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
+        }
+
+        #[test]
+        fn unsigned_field_round_trip(expected in 0u32..8) {
+            let mask = 0b111000;
+            let actual = from_field(mask, into_field(mask, expected).unwrap());
+            prop_assert_eq!(actual, expected);
+        }
+
+        #[test]
+        fn unsigned_field_error(expected in 8u32..100) {
+            prop_assert!(matches!(into_field(0b111000, expected), Err(Error::Unsupported(_))));
+        }
+    }
+}
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -57,8 +57,7 @@ mod tests {
 
        #[test]
         fn signed_field_error_positive(expected in prop_oneof![
-            Just(8),
-            9..100,
+            8..100,
             100..,
         ]) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
@@ -66,8 +65,7 @@ mod tests {
 
         #[test]
         fn signed_field_error_negative(expected in prop_oneof![
-            Just(-9),
-            -100..-10,
+            -100..-9,
             ..-100,
         ]) {
             prop_assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -54,15 +54,17 @@ mod tests {
 
     #[test]
     fn signed_field_error_positive() {
-        for expected in 8 .. 100 {
-            assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
+        let mask = 0b1111000;
+        for value in [8, i32::MAX] {
+            assert!(matches!(into_signed_field(mask, value), Err(Error::Unsupported(_))));
         }
     }
 
     #[test]
     fn signed_field_error_negative() {
-        for expected in -100 .. -8 {
-            assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
+        let mask = 0b1111000;
+        for value in [-9, i32::MIN] {
+            assert!(matches!(into_signed_field(mask, value), Err(Error::Unsupported(_))));
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -30,6 +30,57 @@ pub fn validate(binary: &[u8]) -> Result<(), Error> {
...
+    /// Get the number of values that will be copied if the branch is taken.
+    fn get_val_cnt(&self) -> u32 {
+        (self.data & Self::VAL_CNT_MASK) >> 22
+    }
+
+    /// Get the number of values that will be popped if the branch is taken.
+    fn get_pop_cnt(&self) -> u32 {
+        (self.data & Self::POP_CNT_MASK) >> 27
+    }
+}
```
ia0, line 82: I believe this logic could be simplified by storing only positive values and avoiding sign-extension.

```rust
pub struct SideTableEntry(u32);
pub struct SideTableEntryView {
    pub delta_ip: i32,
    pub delta_stp_mask: i32,
    pub val_cnt: u32,
    pub pop_cnt: u32,
}

impl SideTableEntry {
    const DELTA_IP_MASK: u32 = 0x0000ffff;
    const DELTA_STP_MASK: u32 = 0x003f0000;
    const VAL_CNT_MASK: u32 = 0x07c00000;
    const POP_CNT_MASK: u32 = 0xf8000000;

    fn new(view: SideTableEntryView) -> Result<Self, Result> {
        let mut fields = 0;
        fields |= Self::into_signed_field(Self::DELTA_IP_MASK, view.delta_ip)?;
        fields |= Self::into_signed_field(Self::DELTA_STP_MASK, view.delta_stp)?;
        fields |= Self::into_field(Self::VAL_CNT_MASK, view.val_cnt)?;
        fields |= Self::into_field(Self::POP_CNT_MASK, view.pop_cnt)?;
        SideTableEntry(fields)
    }

    fn view(self) -> SideTableEntryView {
        let delta_ip = Self::from_signed_field(Self::DELTA_IP_MASK, self.0);
        // ...
        SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt }
    }

    fn into_signed_field(mask: u32, value: i32) -> Result<u32, Error> {
        let value = value.wrapping_add(Self::offset(mask));
        Self::into_field(mask, value)
    }

    fn from_signed_field(mask: u32, field: u32) -> i32 {
        Self::from_field(mask, field) as i32 - Self::offset(mask)
    }

    fn offset(mask: u32) -> i32 {
        1 << (mask.count_ones() - 1)
    }

    fn into_field(mask: u32, value: u32) -> Result<u32, Error> {
        let field = (value << mask.trailing_zeros()) & mask;
        if from_field(mask, field) != value {
            return Err(unsupported(if_debug!(Unsupported::SideTable)));
        }
        Ok(field)
    }

    fn from_field(mask: u32, field: u32) -> u32 {
        (field & mask) >> mask.trailing_zeros()
    }
}
```

ia0 (commented):
Thanks! I left some comments that should hopefully simplify the logic. You could make the helper from/into functions into a `bitfield` module with unit-tests.

crates/interpreter/src/valid.rs
```diff
@@ -30,6 +30,78 @@ pub fn validate(binary: &[u8]) -> Result<(), Error> {
...
+        let delta_stp = bit_field::from_signed_field(Self::DELTA_STP_MASK, self.0);
+        let val_cnt = bit_field::from_field(Self::VAL_CNT_MASK, self.0);
+        let pop_cnt = bit_field::from_field(Self::POP_CNT_MASK, self.0);
+
+        SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt }
+    }
+}
+
+mod bit_field {
+    #[allow(unused_imports)]
```
zhouwfang, line 76: The reason of adding it is that there is an unused error for the `use super::Unsupported` below, which is in fact used in `into_field()`. Not sure how to get around it.

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+    use super::bit_field::*;
+    use crate::Error;
+
+    proptest! {
+        #[test]
+        fn signed_field_round_trip(value in -8i32..8) {
+            let mask = 0b1111000;
+            let field = into_signed_field(mask, value).unwrap();
+            let recovered = from_signed_field(mask, field);
+            prop_assert_eq!(value, recovered);
```
ia0, line 917: nit: let's use `actual` and `expected` vocabulary
```suggestion
        fn signed_field_round_trip(expected in -8i32..8) {
            let mask = 0b1111000;
            let actual = from_signed_field(mask, into_signed_field(mask, expected).unwrap());
            prop_assert_eq!(actual, expected);
```

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        }
+
+        #[test]
+        fn unsigned_field_round_trip(value in 0u32..8) {
```
ia0, line 921: Let's do the same here.

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+            let mask = 0b111000;
+            let field = into_field(mask, value).unwrap();
+            let recovered = from_field(mask, field);
+            prop_assert_eq!(value, recovered);
+        }
+    }
+
+    #[test]
+    #[cfg(debug_assertions)]
```
ia0, line 930: You shouldn't need this.

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let mask = 0b111000;
+        let value = 8;
+
+        let result = into_field(mask, value);
+        assert!(result.is_err());
+        assert!(matches!(result.unwrap_err(), Error::Unsupported(_)));
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
```
ia0, line 942: This test will never run. Also it's wrong. The `unwrap` should fail in release mode. We should always fail with a wrong value.

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+    #[cfg(debug_assertions)]
+    fn into_field_error_debug() {
+        // This mask has 3 bits set, so it can only represent values 0-7.
+        let mask = 0b111000;
+        let value = 8;
+
+        let result = into_field(mask, value);
+        assert!(result.is_err());
+        assert!(matches!(result.unwrap_err(), Error::Unsupported(_)));
+    }
```
ia0, line 939: can be simplified to
```suggestion
    fn into_field_error() {
        // The value 8 cannot be represented on 3 bits.
        assert!(matches!(into_field(0b1110000, 8), Err(Error::Unsupported(_))));
    }
```

crates/interpreter/src/valid.rs
```diff
@@ -826,3 +898,67 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let value = 8;
+
+        let field = into_field(mask, value).unwrap();
+        assert_ne!(value, field);
+    }
+
+    #[test]
+    fn into_signed_field_overflow() {
+        let mask = 0b1111000;
+        let value = i32::MAX;
```
ia0, line 955: I would actually prefer to split this test in 2:
- `into_signed_field_error_positive()` testing `into_signed_field(0b1111000, 8)`
- `into_signed_field_error_negative()` testing `into_signed_field(0b1111000, -9)`

That said, I believe we should always use proptest instead, even for negative testing. Which means we should replace those unit-tests to those prop-tests:
- `unsigned_field_error(value in 8u32..)`
- `signed_field_error_positive(value in 8i32..)`
- `signed_field_error_negative(value in ..-8i32)`

ia0 (commented):
Thanks! I like the fact of using `proptest` for testing.

crates/interpreter/src/valid.rs
zhouwfang: Sorry I wrote this test intentionally to explore `debug_assertions` but forgot to remove it...

crates/interpreter/src/bit_field.rs
```diff
@@ -0,0 +1,81 @@
...
+    proptest! {
+        #[test]
+        fn signed_field_round_trip(expected in -8i32..8) {
+            let mask = 0b1111000;
+            let actual = from_signed_field(mask, into_signed_field(mask, expected).unwrap());
+            prop_assert_eq!(actual, expected);
+        }
+
+        #[test]
+        fn signed_field_error_positive(expected in 8i32..100) {
```
zhouwfang, line 60: You may wonder why I wrote `8i32..100` instead of `8i32..` (and similarly below). I find that writing `0i32..` would actually make the test pass! The reason may be that `proptest` randomly samples values and the interval `[0, 7)` somehow was not sampled. So I use `100` to represent infinity, so that every value in `[8, 100)` are all checked.

ia0: I see. I somehow thought proptest would be "smart" and try edge cases then random uniform. But apparently it only does random uniform. And thinking about it more, I think we should also randomize the mask if we want to do it properly.

So I see 2 options:
- We go full proptest and also randomize the mask (see below for hints)
- We go full unittest and brute-force success using the masks you chose and use edge cases for errors

To randomize a mask, you can use the `number-encoding` crate with `prop_compose!()` macro. Something along those lines:

```rust
prop_compose! {
    // We take as input all subsets of 2 distinct elements between 0 and 32 (inclusive), except for
    // the last subset, which is {31, 32}.
    fn arb_mask()(n in 0 .. combination(33, 2) - 1) -> u32 {
        let mut b = [0, 0];
        combinadics::decode_mut(n, 2, &mut b);
        // We use the smallest element of the set as the mask length minus one, since it can range
        // from 0 to 30 (we skip {31, 32}).
        let len = b[0] + 1;
        // We use the largest element of the set as one past the highest bit of the mask, since it
        // can range from 1 to 32.
        let off = b[1] - len;
        (1 << len) - 1 << off
    }
}
```

You can then add the following helper functions:

```rust
fn max_unsigned(mask: u32) -> u32 {
    (1 << mask.count_ones()) - 1
}
fn max_signed(mask: u32) -> i32 {
    (1 << mask.count_ones() - 1) - 1
}
fn min_signed(mask: u32) -> i32 {
    max_signed(mask) - max_unsigned(mask) as i32
}
```

Then proptest success with `(mask, expected) in arb_mask().prop_flat_map(|mask| (Just(mask), 0..=max_unsigned(mask)))` for unsigned and `min_signed(mask)..=max_signed(mask)` for signed.

And proptest failures with `mask in arb_mask()` and testing `max_unsigned(mask) + 1`, `min_signed(mask) - 1`, and `max_signed(mask) + 1`.

ia0 (commented):
Thanks! It looks good. Only the tests can be improved.

crates/interpreter/src/bit_field.rs
zhouwfang: Indeed, it is a great idea to randomize the mask in the meanwhile. The approach I take is different: making use of `prop_oneof!`. Note that the number of test runs is [256 by default](https://docs.rs/proptest/latest/proptest/test_runner/struct.Config.html). The probability that one of the three cases is chosen is `1/3`. So the probability that the edge cases are not chosen is extremely close to `0`.

ia0: Is the branch up-to-date? I don't see using arbitrary masks, which would really be what justifies property testing, otherwise a simple unittest with a for-loop would be faster and simpler.

Using `prop_oneof` is a good idea to skew the odds, but given proptest is not smart (just random), we will waste compute testing the same thing over and over. When the search space is small, we should simply use an exhaustive approach with unittest. When it is bigger but not unreasonably big, we should use proptesting (because the search is stupid). When it is excessively big and the search needs to be smart, then fuzzing is needed.

Here, there are 527 masks (good for proptesting). And there are 12'884'901'820 pairs of mask and value (almost needs fuzzing at that point) ignoring signedness (multiply by 2 otherwise).

So I'm less and less convinced about using proptest for this usecase. If we want to test properly we would need fuzzing. Otherwise if we just want to test a single mask, then unittest is enough. Fuzzing is actually rather easy to do in Rust if you want to try it.

zhouwfang: PTAL at the current version. Basically I revert the tests to the earliest version in this thread with one modification: I set `PROPTEST_CASES` to be 1, which makes the tests only executed once.

The main reason for me to use `proptest` is to avoid writing for loop within a unit test, which should be handled by a testing framework ideally. The reason why I prefer the current version to the randomizing masks approach is that it is easier to read, and the latter may require additional tests for `arb_mask()` to be rigorous.

zhouwfang: > > PTAL at the current version.
> 
> This looks much worse. Now instead of unreliably wasting resources, it unreliably doesn't test what it should.
> 
> > I set `PROPTEST_CASES` to be 1, which makes the tests only executed once.
> 
> How is that a good thing? We want to execute each value once, not the whole proptest once. Because now there's only 1/16 chance of testing any of the values.
>

Sorry I'm confused. Just wanted to make sure we are on the same page. Regardless of `PROPTEST_CASES`, I thought `signed_field_error_positive(expected in 8..100)` means all values of `expected` from 8 to 99 are checked, right? (I did not use `prop_oneof!`.)`PROPTEST_CASES` is about the number of times `signed_field_error_positive` is executed (if it is correct). Let me know what went wrong if any. Thanks.

zhouwfang: And... A drawback of manually writing a for loop with `cargo test` is that it does not tell us the minimum failing value in the error message, besides we'd have to write it manually.

ia0: > Sorry I'm confused.

That's exactly the problem of using frameworks needlessly. It adds more problems than it solves.

> Regardless of `PROPTEST_CASES`, I thought `signed_field_error_positive(expected in 8..100)` means all values of `expected` from 8 to 99 are checked, right?

No, it means it will test 256 (or whatever is `PROPTEST_CASES`) random values (maybe multiple times the same) in the `8..100` range. In particular, if you change the range to `7..100`, you will see that it won't always fail. (Once it fails it will continue failing, but it's just because proptest creates a local regression corpus. If you delete it or don't save it, you're back to square one.)

> `PROPTEST_CASES` is about the number of times `signed_field_error_positive` is executed

No, `PROPTEST_CASES` is how many random values are taken from the strategy.

ia0: > And... A drawback of manually writing a for loop with `cargo test` is that it does not tell us the minimum failing value in the error message, besides we'd have to write it manually.

That's not really an issue. There's 2 ways to deal with that:
- Either don't care because tests are not supposed to fail. If it fails, then it's fine to lose some time debugging with a single printf.
- Or preemptively print the value of the iteration in the assert message. If the message is not too long that's fine.

crates/interpreter/src/bit_field.rs
```diff
@@ -0,0 +1,84 @@
...
+        for expected in -8i32 .. 8 {
+            let actual = from_signed_field(mask, into_signed_field(mask, expected).unwrap());
+            assert_eq!(actual, expected);
+        }
+    }
+
+    #[test]
+    fn signed_field_error_positive() {
+        for expected in 8 .. 100 {
+            assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
```
ia0, line 58:
```suggestion
        let mask = 0b1111000;
        for value in [8, i32::MAX] {
            assert!(matches!(into_signed_field(mask, value), Err(Error::Unsupported(_))));
```

crates/interpreter/src/bit_field.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    #[test]
+    fn signed_field_error_negative() {
+        for expected in -100 .. -8 {
+            assert!(matches!(into_signed_field(0b1111000, expected), Err(Error::Unsupported(_))));
```
ia0, line 65:
```suggestion
        let mask = 0b1111000;
        for value in [-8, i32::MIN] {
            assert!(matches!(into_signed_field(mask, value), Err(Error::Unsupported(_))));
```

crates/interpreter/src/bit_field.rs
```diff
@@ -0,0 +1,84 @@
...
+        for expected in 0 .. 8 {
+            let actual = from_field(mask, into_field(mask, expected).unwrap());
+            assert_eq!(actual, expected);
+        }
+    }
+
+    #[test]
+    fn unsigned_field_error() {
+        for expected in 8 .. 10 {
+            assert!(matches!(into_field(0b111000, expected), Err(Error::Unsupported(_))));
```
ia0, line 81:
```suggestion
        let mask = 0b111000;
        for value in [8, u32::MAX] {
            assert!(matches!(into_field(mask, value), Err(Error::Unsupported(_))));
```

ia0 (commented):
Thanks! LGTM

ia0 (approved):

Issue #448, opened 2024-04-29
Create a tool to automate updating changelog files


ia0:
See [this doc](https://github.com/google/wasefire/blob/main/docs/contributing/changelog.md#logging-a-change) for a description of the problem. The goal of this issue is to write a tool that would update the `CHANGELOG.md` file of a crate, and if necessary the `Cargo.toml` too as well as all the dependents `Cargo.toml` and `CHANGELOG.md` files if any.

It should probably look something like:

```shell
cargo xtask change crates/prelude minor 'Add foo::bar module to support FooBar'
cargo xtask change crates/board major 'Change Foo::bar() to return an error'
```

invernizzi:
Would https://github.com/google-github-actions/release-please-action solve this issue?

ia0:
> Would https://github.com/google-github-actions/release-please-action solve this issue?

Not really. I haven't yet seen a tool that supports more than simple repositories. In particular, most tools (like `release-please`) rely on conventional commits which has deep issues for mono-repos (this repository contains 72 crates of which 14 are published). In particular, conventional commits don't support PRs that touch multiple crates in different ways but still with the same purpose. There is [this issue](https://github.com/conventional-commits/conventionalcommits.org/issues/269) trying to add mono-repo support but it has been stale for 2 years and doesn't even fix the problem. It only suggests to support multiple scopes (which also has a [stale issue](https://github.com/conventional-commits/conventionalcommits.org/issues/302)). But multiple scopes is not enough because changes may be of different nature to different crates. This is particularly true in Rust with proc-macro, where a change to the proc-macro crate may be minor while its usage in another crate may be major (typically `api-macro` and `api` in the case of this repository).[^1]

However, it's a good point that a part of this task could be to look for existing tools to get inspiration or even find something working that doesn't need more setup and maintenance than writing something custom (I don't expect the tool we need to be more than 100 lines of code, in particular given most of the code will be shared with other tooling or reuse existing crates like `semver` or maybe `git2`, while I expect most of those tools to need more than 100 lines of configuration because they don't scale with the number of crates).

[^1]: That's actually something that is not automatable. The tool could either try to detect those cases or simply do something wrong which is going to be caught during review.

PR #569, opened 2024-08-05
Loosen dependency crate version requirements

exzachlyvv:


crates/api-desc/Cargo.toml
```diff
@@ -14,9 +14,9 @@ categories = ["embedded", "no-std", "wasm"]
 [dependencies]
 data-encoding = { version = "2.6.0", default-features = false }
 data-encoding-macro = { version = "0.1.15", default-features = false }
-proc-macro2 = { version = "1.0.83", default-features = false }
-quote = { version = "1.0.36", default-features = false }
-syn = { version = "2.0.66", default-features = false, features = ["printing"] }
+proc-macro2 = { version = "1.0", default-features = false }
+quote = { version = "1.0", default-features = false }
+syn = { version = "2.0", default-features = false, features = ["printing"] }
```
ia0, line 19: Let's use `M.m.p` even if `p` is 0 for consistency. That's what `cargo add` does and it's preserved by `cargo update --breaking`. If we want to drop the patch version (which we probably shouldn't for security reasons), we should do it across the whole project.

ia0 (changes requested):
Thanks!

Let's also do the same change to `crates/wire-derive` so that we can share compilation artifacts.

ia0 (dismissed):
Thanks!

exzachlyvv:
I realize I missed updating `[dependencies.syn]` in wire-derive. Want me to get that in this PR quick?

exzachlyvv:
Seeing `Error: CHANGELOG.md for crates/api-desc is not up-to-date`

I'll update

exzachlyvv:
> I realize I missed updating [dependencies.syn] in wire-derive. Want me to get that in this PR quick?

Done.

> Seeing Error: CHANGELOG.md for crates/api-desc is not up-to-date

Done.

ia0 (commented):
You're up for a ride. You'll also have to update the crate version in the Cargo.toml and then recursively into the dependents. There's #448 to help with this ideally.

ia0:
You can just run ./scripts/ci-changelog.sh locally. But it won't help finding the dependents. There's either git grep or running the full ci.sh.

exzachlyvv:
> You're up for a ride. ...

Hahaha I see now :) what is the recommended approach?

ia0:
> Hahaha I see now :) what is the recommended approach?

I personally just use something along the lines of `sed -i '/^wasefire-stuff =/{s/oldver/newver/}' $(git ls-files '*Cargo.toml')` then manually edit those using `[dependencies.wasefire-stuff]` that I find with git grep. Then I go over each changelog file returned by ci-changelog until it passes. You can also check https://github.com/google/wasefire/blob/main/docs/contributing/changelog.md and possibly improve it. But ideally we should fix #448.

exzachlyvv:
Took a simpler approach, update: `Increment to skip CHANGELOG.md test ...`

Will that cause issues? I guess we'll want to publish this commit to crates.io so should we increment versioning anyways?

ia0:
> Took a simpler approach, update: `Increment to skip CHANGELOG.md test ...`

That's actually something I want the tool of #448 to detect and warn against. Any change to what would be published since the last published version should bump the version accordingly, in particular if the version is not yet a prerelease, it should become a prerelease, to avoid having 2 possible states for the same release version (one in github and one in crates.io). If the github version is a release version then it should exactly match what is on crates.io (mostly for auditing purposes).

> Will that cause issues?

Apart from the auditing issue above, the publish script will also not publish a crate that is already a release version.

> I guess we'll want to publish this commit to crates.io so should we increment versioning anyways?

Yes exactly. But the increment is done at each change, not at the end when we forgot what happened.

I'll add a commit to this PR to show the idea, and merge it (I just want to check cargo audit first).

ia0:
> > Took a simpler approach, update: `Increment to skip CHANGELOG.md test ...`
> 
> That's actually something I want the tool of #448 to detect and warn against. Any change to what would be published since the last published version should bump the version accordingly, in particular if the version is not yet a prerelease, it should become a prerelease, to avoid having 2 possible states for the same release version (one in github and one in crates.io). If the github version is a release version then it should exactly match what is on crates.io (mostly for auditing purposes).
> 
> > Will that cause issues?
> 
> Apart from the auditing issue above, the publish script will also not publish a crate that is already a release version.
> 
> > I guess we'll want to publish this commit to crates.io so should we increment versioning anyways?
> 
> Yes exactly. But the increment is done at each change, not at the end when we forgot what happened.
> 
> I'll add a commit to this PR to show the idea, and merge it (I just want to check cargo audit first in regard to #570).

crates/board/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Patch
 
+- Update dependencies
```
ia0, line 11: Just for example, here there is already a pre-release, so we just add the change to the appropriate sub-section (and bump the pre-release version in case we add a sub-section with the highest priority).

crates/cli-tools/CHANGELOG.md
```diff
@@ -12,4 +12,4 @@
 
 ## 0.1.0
 
-<!-- Increment to skip CHANGELOG.md test: 0 -->
+<!-- Increment to skip CHANGELOG.md test: 1 -->
```
ia0, line 15: And here, since it's already a pre-release with a similar change, we don't duplicate it and instead just bump the skip counter to acknowledge that the changelog is up-to-date.

ia0 (approved):

ia0-review (approved):

exzachlyvv:
Nice, thanks!

PR #572, opened 2024-08-06
Abstract writing to flash on Nordic

ia0:
This follows #568 towards dynamic loading and unloading of applets. The process of writing to flash can be shared between platform update and applet install.

### DIFF
--- a/crates/runner-nordic/src/storage.rs
+++ b/crates/runner-nordic/src/storage.rs
@@ -14,6 +14,7 @@
 
 use alloc::borrow::Cow;
 use alloc::vec;
+use alloc::vec::Vec;
 use core::ptr::addr_of_mut;
 
 use embedded_storage::nor_flash::{
@@ -21,9 +22,11 @@ use embedded_storage::nor_flash::{
 };
 use nrf52840_hal::nvmc::Nvmc;
 use nrf52840_hal::pac::NVMC;
+use wasefire_error::{Code, Error};
 use wasefire_store::{self as store, StorageError, StorageIndex, StorageResult};
 use wasefire_sync::{AtomicBool, Ordering, TakeCell};
 
+const WORD_SIZE: usize = <Nvmc<NVMC>>::WRITE_SIZE;
 const PAGE_SIZE: usize = <Nvmc<NVMC>>::ERASE_SIZE;
 
 static DRIVER: TakeCell<NVMC> = TakeCell::new(None);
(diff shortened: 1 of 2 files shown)


crates/runner-nordic/src/storage.rs
```diff
@@ -90,20 +93,107 @@ impl Storage {
...
+    // None if not running.
+    dry_run: Option<bool>,
+    // offset + buffer.len() <= storage.len() && offset % WORD_SIZE == 0
+    offset: usize,
+    // buffer.len() < WORD_SIZE
+    buffer: Vec<u8>,
+}
+
+impl StorageWriter {
+    pub fn new(storage: Storage) -> Result<Self, (Storage, Error)> {
```
ia0-review, line 116: This never fails. It will also simplify the storage init in update.

crates/runner-nordic/src/storage.rs
```diff
@@ -90,20 +93,107 @@ impl Storage {
...
 
-    pub const ERASE_SIZE: usize = PAGE_SIZE;
-    pub fn erase(&self, offset: usize, length: usize) -> StorageResult<()> {
-        let mut helper = Helper::new(self);
-        helper.nvmc().erase(offset as u32, (offset + length) as u32).map_err(convert)
+pub struct StorageWriter {
+    storage: Storage,
+    // None if not running.
+    dry_run: Option<bool>,
+    // offset + buffer.len() <= storage.len() && offset % WORD_SIZE == 0
```
ia0-review, line 109: We never check and use `offset + buffer.len() <= storage.len()`, we just rely on the NVMC implementation. We can just remove this part of the comment.

crates/runner-nordic/src/storage.rs
```diff
@@ -90,20 +93,107 @@ impl Storage {
...
-    pub const WRITE_SIZE: usize = <Nvmc<NVMC>>::WRITE_SIZE;
-    pub fn write(&self, offset: usize, data: &[u8]) -> StorageResult<()> {
-        let mut helper = Helper::new(self);
-        helper.nvmc().write(offset as u32, data).map_err(convert)
+    pub fn dry_run(&self) -> Result<bool, Error> {
+        self.dry_run.ok_or(Error::user(Code::InvalidState))
+    }
+
+    pub fn start(&mut self, dry_run: bool) -> Result<(), Error> {
+        self.dry_run = Some(dry_run);
```
ia0-review, line 130: We should check if `dry_run.is_some()` in which case we should reset the other fields.

crates/runner-nordic/src/storage.rs
```diff
@@ -90,20 +93,107 @@ impl Storage {
...
+        }
+        assert!(self.buffer.is_empty());
+        let length = chunk.len() / WORD_SIZE * WORD_SIZE;
+        self.aligned_write(&chunk[.. length])?;
+        self.buffer.extend_from_slice(&chunk[length ..]);
+        Ok(())
+    }
+
+    fn write_buffer(&mut self) -> Result<(), Error> {
+        let data = core::mem::take(&mut self.buffer);
```
ia0-review, line 169: We could `assert_eq!(self.buffer.len(), WORD_SIZE)`.

ia0-review (changes requested):

ia0-review (approved):

PR #574, opened 2024-08-07
Use wasefire-error in wasefire-store

ia0:
This removes a bit of conversion logic.

### DIFF
--- a/crates/error/src/lib.rs
+++ b/crates/error/src/lib.rs
@@ -18,6 +18,8 @@
 
 // TODO(https://github.com/rust-lang/rust/issues/122105): Remove when fixed.
 extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
 
 use num_enum::{IntoPrimitive, TryFromPrimitive, TryFromPrimitiveError};
 
@@ -66,6 +68,20 @@ impl Error {
     /// Creates a world error.
     pub fn world(code: impl CodeParam) -> Self {
         Error::new(Space::World, code)
+    }
+
+    /// Pops the error one level up.
+    ///
+    /// User errors become internal errors. Internal errors become world errors.
+    pub fn pop(self) -> Self {
+        let mut space = self.space();
+        match Space::try_from_primitive(space) {
+            Ok(Space::User) => space = Space::Internal as u8,
+            Ok(Space::Internal) => space = Space::World as u8,
+            _ => (),
+        }
+        let code = self.code();
+        Error::new_const(space, code)
     }
 
     /// Decodes a signed integer as a result (where errors are negative values).
@@ -222,6 +238,13 @@ impl defmt::Format for Error {
 }
 
 impl core::error::Error for Error {}
+
+#[cfg(feature = "std")]
+impl From<std::io::Error> for Error {
+    fn from(_: std::io::Error) -> Self {
+        Error::world(Code::Generic)
+    }
+}
 
 #[cfg(test)]
 mod tests {
--- a/crates/store/src/buffer.rs
+++ b/crates/store/src/buffer.rs
@@ -117,8 +117,8 @@ impl BufferStorage {
     /// Arms an interruption after a given delay.
     ///
     /// Before each subsequent mutable operation (write or erase), the delay is decremented if
-    /// positive. If the delay is elapsed, the operation is saved and an error is returned.
-    /// Subsequent operations will panic until either of:
+    /// positive. If the delay is elapsed, the operation is saved and the [`INTERRUPTION`] error is
+    /// returned. Subsequent operations will panic until either of:
     /// - The interrupted operation is [corrupted](BufferStorage::corrupt_operation).
     /// - The interruption is [reset](BufferStorage::reset_interruption).
     ///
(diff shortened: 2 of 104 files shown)


crates/error/src/lib.rs
```diff
@@ -68,6 +70,20 @@ impl Error {
         Error::new(Space::World, code)
     }
 
+    /// Pops the error one level up.
+    ///
+    /// User errors become internal errors. Internal errors become world errors.
+    pub fn pop(self) -> Self {
```
ia0-review, line 76: Also document this in the changelog.

crates/store/src/driver.rs
```diff
@@ -388,7 +390,7 @@ impl StoreDriverOn {
                 }
                 (store_result.err(), StoreDriver::On(self))
             }
-            Err(StoreError::StorageError) => {
+            Err(crate::buffer::INTERRUPTION_ERROR) => {
```
ia0-review, line 393: Let's rename this to `INTERRUPTION`.

crates/store/src/driver.rs
```diff
@@ -364,12 +366,12 @@ impl StoreDriverOn {
...
         mut self, operation: StoreOperation, interruption: StoreInterruption,
-    ) -> Result<(Option<StoreError>, StoreDriver), (Store<BufferStorage>, StoreInvariant)> {
+    ) -> Result<(Option<Error>, StoreDriver), (Store<BufferStorage>, StoreInvariant)> {
         self.store.storage_mut().arm_interruption(interruption.delay);
         let (deleted, store_result) = self.store.apply(&operation);
         Ok(match store_result {
-            Err(StoreError::NoLifetime) => return Err((self.store, StoreInvariant::NoLifetime)),
-            Ok(()) | Err(StoreError::NoCapacity) | Err(StoreError::InvalidArgument) => {
+            Err(crate::store::NO_LIFETIME) => return Err((self.store, StoreInvariant::NoLifetime)),
+            Ok(()) | Err(crate::store::NO_CAPACITY) | Err(crate::store::INVALID_ARGUMENT) => {
```
ia0-review, line 374: Let's expose all error constants at the root and use them from there.

crates/store/src/store.rs
```diff
@@ -146,17 +101,17 @@ impl StoreHandle {
     ///
     /// # Errors
     ///
-    /// Returns [`StoreError::InvalidArgument`] if the entry has been deleted or compacted.
-    pub fn get_length<S: Storage>(&self, store: &Store<S>) -> StoreResult<usize> {
+    /// Returns an error if the entry has been deleted or compacted.
```
ia0-review, line 104: Let's keep the constant name here since it will be exposed. Same throughout the file.

crates/store/src/store.rs
```diff
@@ -23,68 +23,23 @@ use core::convert::TryFrom;
...
-/// The None case is considered invalid and returns [`StoreError::InvalidStorage`].
-fn or_invalid<T>(x: Option<T>) -> StoreResult<T> {
-    x.ok_or(StoreError::InvalidStorage)
-}
+pub(crate) const INVALID_ARGUMENT: Error =
+    Error::new_const(Space::User as u8, Code::InvalidArgument as u16);
+pub(crate) const INVALID_STORAGE: Error =
+    Error::new_const(Space::World as u8, Code::InvalidState as u16);
+pub(crate) const NO_CAPACITY: Error = Error::new_const(Space::World as u8, Code::NotEnough as u16);
+pub(crate) const NO_LIFETIME: Error =
```
ia0-review, line 41: Let's reuse the documentation when making those public.

crates/store/src/store.rs
```diff
@@ -1114,7 +1069,7 @@ impl<S: Storage> Store<S> {
     }
 
     /// Erases a page if not already erased.
-    fn storage_erase_page(&mut self, page: Nat) -> StoreResult<()> {
+    fn storage_erase_page(&mut self, page: Nat) -> Result<(), Error> {
         if !is_erased(&self.read_page(page)) {
             self.storage.erase_page(page as usize)?;
```
ia0-review, line 1074: We should pop here. Search all `self.storage.`.

ia0-review (changes requested):

crates/store/src/buffer.rs
```diff
@@ -197,7 +199,12 @@ impl BufferStorage {
     pub fn set_page_erases(&mut self, page: usize, cycle: usize) {
         self.page_erases[page] = cycle;
     }
+}
 
+/// Error indicating that an operation was interrupted.
+pub const INTERRUPTION: Error = Error::new_const(Space::World as u8, 0xffff);
```
ia0-review, line 205: We could actually make this `Space::Internal`. Not sure though what's best.

ia0-review (dismissed):

crates/store/src/buffer.rs
ia0: Not really, because it would change space between when it's produced and when it's compared. What we could do though is expose the popped error and generate the internal one. I'll test it.

ia0: It works. I agree it's better.

ia0-review (approved):

Issue #577, opened 2024-08-09
Support dynamic loading and unloading of applets


zhouwfang:
This issue is the intersection of #56 and #38. This issue does not try to support multiple applets (or if so, with a maximum of one applet). This would be done in a separate issue.

PR #579, opened 2024-08-09
Implement From<interpreter::Error> for wasefire_error::Error

ia0:
Part of #577 because starting an applet (dynamically) should not crash the platform in case of errors but instead propagate the error to the host.

### DIFF
--- a/crates/interpreter/CHANGELOG.md
+++ b/crates/interpreter/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.3.1-git
+
+### Major
+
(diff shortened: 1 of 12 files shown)


crates/interpreter/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.3.1-git
+
+### Major
```
ia0-review, line 5:
```suggestion
### Minor
```

ia0-review (changes requested):

ia0-review (dismissed):

ia0-review (approved):

PR #584, opened 2024-08-09
Add "xtask changelog ci" to validate CHANGELOG.md files

exzachlyvv:
End goal ([bug](https://github.com/google/wasefire/issues/448)): Create script to automate updating CHANGELOG.md

This task will be done in 2 parts:
1. `xtask changelog ci`: validates all CHANGELOG.md files <-- this PR
2. `xtask changelog {major,minor,patch}`: automatically updates changelogs with a new description.

### DIFF
--- a/crates/api-macro/CHANGELOG.md
+++ b/crates/api-macro/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.6.2-git
 
-## Patch
+### Patch
 
 - Update dependencies
 
--- a/crates/xtask/src/changelog.rs
+++ b/crates/xtask/src/changelog.rs
@@ -44,7 +44,7 @@ struct Changelog {
 
 impl Changelog {
     fn read_file(path: &str) -> Result<Changelog> {
-        Self::parse(String::from_utf8(fs::read(&path)?)?.as_str())
+        Self::parse(String::from_utf8(fs::read(path)?)?.as_str())
     }
 
     /// Converts raw file contents into a Changelog data structure.
@@ -171,7 +171,7 @@ impl Release {
     }
 }
 
-/// Validates CHANGELOG.md files for all Wasefire crates.
+/// Validates all changelog files.
 pub fn execute_ci() -> Result<()> {
     let paths = cmd::output(Command::new("git").args(["ls-files", "*/CHANGELOG.md"]))?;
 
@@ -186,13 +186,11 @@ pub fn execute_ci() -> Result<()> {
     Ok(())
 }
 
-/// Updates a CHANGELOG.md file and CHANGELOG.md files of dependencies.
-pub fn execute_change(
-    crate_path: &str, _release_type: &ReleaseType, _description: &str,
-) -> Result<()> {
-    ensure!(fs::exists(crate_path), "Crate does not exist: {}", crate_path);
+/// Updates the changelog of a crate and its dependents.
+pub fn execute_change(path: &str, _scope: &ReleaseType, _description: &str) -> Result<()> {
+    ensure!(fs::exists(path), "Crate does not exist: {path}");
 
-    let _changelog = Changelog::read_file(format!("{crate_path}/CHANGELOG.md").as_str())?;
+    let _changelog = Changelog::read_file(format!("{path}/CHANGELOG.md").as_str())?;
 
     todo!("Implement changelog updates");
 }
--- a/crates/xtask/src/changelog.rs
+++ b/crates/xtask/src/changelog.rs
@@ -12,11 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use anyhow::{anyhow, ensure, Context, Result};
-use clap::ValueEnum;
-use semver::Version;
+use core::str;
+use std::collections::BTreeMap;
+use std::fmt::Write;
+use std::io::BufRead;
+use std::process::Command;
 
-#[derive(Debug, Clone, clap::ValueEnum)]
+use anyhow::{anyhow, bail, ensure, Context, Result};
+use semver::Version;
+use wasefire_cli_tools::{cmd, fs};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
 pub enum ReleaseType {
     Major,
     Minor,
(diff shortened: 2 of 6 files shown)


crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+#[derive(Debug, Default, PartialEq)]
+struct Changelog {
+    releases: Vec<ChangelogRelease>,
+
+    increment_count: u32,
+}
+
+impl Changelog {
+    fn push_release(&mut self, release: ChangelogRelease) {
+        self.releases.push(release);
```
ia0, line 34: Maybe we should check here that:
- releases are in decreasing precedence order (i.e. push on empty should be 0.1.0, otherwise should be higher precedence that the top)
- can't push on top of a pre-release (i.e. if there's a pre-release it's the top)

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    fn set_increment_count(&mut self, num: u32) {
+        self.increment_count = num;
```
ia0, line 38: Same here, maybe we need to have the field optional? We want to make sure there is exactly one such field and it's the last line. I'll check the rest of the code, but probably we don't want those 2 setter functions and instead directly build the `Changelog` from its fields: `Changelog { releases, skip_counter }`.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+/// Each release of a Changelog, containing a version string and optional major, minor, patch change
+/// descriptions.
+#[derive(Debug, PartialEq)]
+struct ChangelogRelease {
+    version: String,
+
+    major: Option<Vec<String>>,
```
ia0, line 48: We don't need the option because an empty section is meaningless.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    fn set_increment_count(&mut self, num: u32) {
+        self.increment_count = num;
+    }
+}
+
+/// Each release of a Changelog, containing a version string and optional major, minor, patch change
+/// descriptions.
+#[derive(Debug, PartialEq)]
+struct ChangelogRelease {
+    version: String,
```
ia0, line 46: We should probably use `semver::Version`.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+        self.releases.push(release);
+    }
+
+    fn set_increment_count(&mut self, num: u32) {
+        self.increment_count = num;
+    }
+}
+
+/// Each release of a Changelog, containing a version string and optional major, minor, patch change
+/// descriptions.
```
ia0, line 43: Let's try to keep the title short. But I would also not be against having no doc. This is self-describing. (Same for the `Changelog` type actually).

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    fn set_major(&mut self, contents: Vec<String>) {
+        self.major = Some(contents);
+    }
+
+    fn set_minor(&mut self, contents: Vec<String>) {
+        self.minor = Some(contents);
+    }
+
+    fn set_patch(&mut self, contents: Vec<String>) {
+        self.patch = Some(contents);
```
ia0, line 67: Same here, we probably don't need setters. Those functions are too low-level and thus we already have visibility on the fields directly.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+pub fn execute(crate_path: String, _version: ChangelogReleaseType, _message: String) -> Result<()> {
```
ia0, line 71: We don't need to take ownership of arguments, except `ReleaseType` which should be `Copy` anyway.

For paths, we can either take `&str`, or even better `&Path` (or `&impl AsRef<Path>`). But given the rest of xtask uses `String` instead of `PathBuf` for file flags, then we should use `&str` here.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    ensure!(
+        std::fs::exists(&crate_path).expect("Cannot open path right now."),
```
ia0, line 73: We should have a helper function `cli_tools::fs::exists()`. And we can use `?` to propagate the error. The caller can use `.context()` if they want.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+        crate_path
+    );
+
+    let _changelog = read_changelog_file(crate_path)?;
+    // TODO: act upon changelog...
+
+    Ok(())
+}
+
+fn read_changelog_file(crate_path: String) -> Result<Changelog> {
```
ia0, line 84: Those functions could be methods of `Changelog`:
- `Changelog::read()`
- `Changelog::parse()` (or we could implement `FromStr`, and similarly implement `Display` for rendering)

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+    let changelog_path = format!("{crate_path}/CHANGELOG.md");
+
+    let changelog_contents = std::fs::read_to_string(&changelog_path)
```
ia0, line 87: You can use `cli_tools::fs::read()?` which will automatically add the path to the error message.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+        .with_context(|| format!("reading {changelog_path}"))?;
+
+    parse_changelog(changelog_contents)
+}
+
+/// Converts raw file contents into a Changelog data structure.
+fn parse_changelog(contents: String) -> Result<Changelog> {
+    let mut changelog = Changelog::default();
```
ia0, line 95: In Rust you don't need to create a dummy object and fill it. You can build it directly at the end from its fields.

```rust
let mut releases = Vec::new();
let mut skip_counter = None;
let mut state = xxx;
for line in contents.lines() {
    if let Some(version) = line.strip_prefix("## ") {
        // check and update state, possibly pushing parsed release
        continue;
    }
    // etc (including the skip counter)
}
let skip_counter = match skip_counter {
    None => bail!("missing skip counter"),
    Some(x) => x,
};
Ok(Changelog { releases, skip_counter })

File: crates/xtask/src/changelog.rs [hunk 12]
```diff
@@ -0,0 +1,377 @@
...
+
+    // Trim off the trailing comment: <!-- Increment to skip ... -->
+    let contents: Vec<&str> =
+        contents.split("\n\n<!-- Increment to skip CHANGELOG.md test:").collect();
```
ia0, line 99: We should probably use a state machine rather than string manipulations. Or at least we could use `contents.lines().peekable()` which should have enough state that we don't need anything more to know where we are.

For example we know the last version is `0.1.0` and has no sections and is directly followed by the skip counter.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+                    }
+                ],
+                increment_count: 0,
+            }
+        );
+    }
+
+    #[test]
+    fn parse_changelog_with_missing_release_type_success() {
+        let changelog = r###"# Changelog
```
ia0, line 238: Do we really need `r###"`? Doesn't `r"` suffices?

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        changelog.push_release(release);
+    }
+
+    Ok(changelog)
+}
+
+#[cfg(test)]
+mod tests {
```
ia0, line 154: Ideally, we should use fuzzing to make sure parsing and rendering round-trips. Positive tests are not really needed because the repository is already full of changelog files, so CI will always execute parsing on those files. And negative tests are hard to write, which is why fuzzing would be simpler.

crates/xtask/src/main.rs
```diff
@@ -94,6 +95,18 @@ enum MainCommand {
...
+    Changelog {
+        /// Crate to update
+        crate_path: String,
+
+        // Either "major", "minor", or "patch". TODO(zvv): turn into enum.
+        version: changelog::ChangelogReleaseType,
+
+        /// Message added to the CHANGELOG.md
+        message: String,
+    },
```
ia0, line 109: I'm actually not sure we should start with this feature. I think it would be better to start with the CI check:
- Changelog file is well formatted (as defined in the doc, there are many checks).
- Changelog matches the Cargo.toml (could be done in different PR).

ia0 (commented):
Thanks! I've created a `dev/changelog` branch for this work (making it the base branch). We can merge it at 2 different points in time:
- Once when we have the CI check (e.g. `xtask changelog ci`).
- Once when we have `xtask changelog {major,minor,patch}` commands.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,377 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+use anyhow::{ensure, Context, Result};
+use clap::ValueEnum;
+
+#[derive(clap::ValueEnum, Clone, Debug)]
```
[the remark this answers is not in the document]
exzachlyvv: TODO: zach figure out auto format for imports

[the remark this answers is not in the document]
ia0: If by imports you mean `use` items, then you just need to make sure they are all packed together (possibly with empty lines) and rustfmt will do the rest.

crates/xtask/src/changelog.rs
exzachlyvv: Methods on Changelog for now

crates/xtask/src/changelog.rs
exzachlyvv: Are you suggesting to test at a higher level? i.e. assert Changelog::parse({input}).render() == {input}

And use the existing CHANGELOG.md files?

crates/xtask/src/main.rs
exzachlyvv: Makes sense! sounds good to me, I'll go that direction.

crates/xtask/src/changelog.rs
ia0: Yes, except that parsing may fail, which counts as a successful test. For the existing changelog, no need to do anything special, except possibly use them to bootstrap a corpus.

exzachlyvv: Gotcha, if you are okay with it, I'll keep some small tests around. I find them helpful when iterating on logic :)

crates/xtask/src/main.rs
exzachlyvv: I've made updates to create 2 subcommand: `ci` and `change`

* `ci` validates all crates/* that have a changelog.md file
* `change` is unimplementing

exzachlyvv:
Last part is still WIP - refactor string parsing to be stateful instead of rough sting manipulation.

exzachlyvv:
This PR is ready for another pass. There is 1 validation missing: "The first version matches the Cargo.toml version." Will do that in follow up PR.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+                let mut release = Release::from(version_string)?;
+
+                // Parse each release type (major, minor, patch)
+                while next_line.is_some_and(|line| line.starts_with("### ")) {
+                    // Advance
+                    current_line = lines.next().unwrap();
+                    next_line = lines.peek();
+
+                    let release_type_string = current_line.trim_start_matches("### ");
+                    let release_type = ReleaseType::from_str(release_type_string, true)
```
ia0, line 74: Interesting. Here I would also be more strict and check casing too.

```rust
let release_type = match release_type_string {
    "Major" => ReleaseType::Major,
    ...
    _ => bail!("Failed to parse version header {release_type_string:?}"),
};

File: crates/xtask/src/changelog.rs [hunk 18]
```diff
@@ -0,0 +1,549 @@
...
+                        .is_some_and(|line| line.starts_with("- ") || line.starts_with("  "))
+                    {
+                        // Advance
+                        current_line = lines.next().unwrap();
+                        next_line = lines.peek();
+
+                        // New description
+                        if current_line.starts_with("- ") {
+                            release_descriptions
+                                .push(current_line.trim_start_matches("- ").to_string());
```
ia0, line 92: I would avoid using `trim`. Let's use `strip_prefix().unwrap()` if needed, otherwise ideally we would `match strip_prefix()` and handle both cases, but I can see how it's not always convenient.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+                        } else if current_line.starts_with("  ") {
+                            // Append to previous description
+                            let mut last = release_descriptions
+                                .pop()
+                                .expect("No last description to append to.");
+                            last.push_str(format!("\n{current_line}").as_str());
+                            release_descriptions.push(last);
```
ia0, line 99: You should be able to do this in place.

```rust
write!(release_descriptions.last_mut().context("Invalid continuation")?, "\n{current_line}");
```

Maybe not for this PR, but I think we should store logical lines and pretty-print with line folding/wrapping.

```
- This very long
  line
```

would thus become `This very long line` which would essentially be the user input to `cargo xtask changelog change`.

We should also check that descriptions don't end with a dot (can also be another PR).

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+                        }
+                    }
+
+                    match release_type {
+                        ReleaseType::Major => release.major = release_descriptions,
+                        ReleaseType::Minor => release.minor = release_descriptions,
+                        ReleaseType::Patch => release.patch = release_descriptions,
```
ia0, line 106: Actually, I think it might be beneficial to use a map for the descriptions: `BTreeMap<ReleaseType, Vec<String>>` with the invariant that values are not empty and there is at least one key.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+                    }
+                }
+
+                // Validate descending versions.
+                if let Some(previous_version) = releases.last() {
+                    if release.version > previous_version.version {
```
ia0, line 112: You can use `ensure!(release.version < previous_version.version, "Versions should be order by decreasing precedence")`.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+struct Release {
+    version: Version,
+
+    major: Vec<String>,
+    minor: Vec<String>,
+    patch: Vec<String>,
+}
+
+impl Release {
+    fn from(version: &str) -> Result<Self> {
```
ia0, line 166: nit
```suggestion
    fn new(version: &str) -> Result<Self> {
```

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+        Ok(Release { version: semver, major: Vec::new(), minor: Vec::new(), patch: Vec::new() })
+    }
+}
+
+/// Validates CHANGELOG.md files for all Wasefire crates.
+pub fn execute_ci() -> Result<()> {
+    let dir_entries = wasefire_cli_tools::fs::read_dir("crates")?;
+
+    let existing_changelog_paths = dir_entries.filter_map(|entry| {
+        if let Ok(crate_path) = entry {
```
ia0, line 178: Did you encounter errors to not propagate them? like `let entry = entry?;`

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+                }
+            }
+        }
+
+        None
+    });
+
+    for changelog_path in existing_changelog_paths {
+        // Validation done during parsing.
+        Changelog::read_file(&changelog_path)
```
ia0, line 193: We should ideally also pretty-print and check that it round-trips. But this could be in another PR.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+
+        Ok(Release { version: semver, major: Vec::new(), minor: Vec::new(), patch: Vec::new() })
+    }
+}
+
+/// Validates CHANGELOG.md files for all Wasefire crates.
+pub fn execute_ci() -> Result<()> {
+    let dir_entries = wasefire_cli_tools::fs::read_dir("crates")?;
+
+    let existing_changelog_paths = dir_entries.filter_map(|entry| {
```
ia0, line 177: Since you have a for loop just below, I would inline the filtering logic there using `continue`.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn from(version: &str) -> Result<Self> {
+        let semver = Version::parse(version)?;
+
+        Ok(Release { version: semver, major: Vec::new(), minor: Vec::new(), patch: Vec::new() })
+    }
+}
+
+/// Validates CHANGELOG.md files for all Wasefire crates.
+pub fn execute_ci() -> Result<()> {
+    let dir_entries = wasefire_cli_tools::fs::read_dir("crates")?;
```
ia0, line 175: I wonder if it's not more robust to use something like:

```rust
let paths = cmd::output(Command::new("git").args(["ls-files", "*/CHANGELOG.md"]))?;
for path in paths.stdout.lines() {
```

like in `crates/xtask/src/textreview.rs`.

ia0 (commented):
Thanks! Looks pretty good to me.

crates/api-macro/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.6.2-git
 
-## Patch
+### Patch
```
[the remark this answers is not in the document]
ia0: Fixing this in #592 to avoid mistakes.

crates/xtask/src/changelog.rs
exzachlyvv: I'll do that in another PR. Trying to keep scopes smaller.

crates/xtask/src/changelog.rs
exzachlyvv: much cleaner :)

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+
+        Self::parse(String::from_utf8(changelog_contents)?)
+    }
+
+    /// Converts raw file contents into a Changelog data structure.
+    ///
+    /// Validates requirements in docs/contributing/changelog.md
+    fn parse(contents: String) -> Result<Changelog> {
+        let mut skip_counter = 0;
+        let mut releases: Vec<Release> = Vec::new();
```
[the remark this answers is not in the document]
exzachlyvv: It's needed in this case. Compiler complains without it.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+}
+
+/// Validates CHANGELOG.md files for all Wasefire crates.
+pub fn execute_ci() -> Result<()> {
+    let dir_entries = wasefire_cli_tools::fs::read_dir("crates")?;
+
+    let existing_changelog_paths = dir_entries.filter_map(|entry| {
+        if let Ok(crate_path) = entry {
+            if crate_path.metadata().is_ok_and(|metadata| metadata.is_dir()) {
```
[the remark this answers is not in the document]
exzachlyvv: Non-issue when using the Command::new("git") approach.

crates/xtask/src/changelog.rs
exzachlyvv: Nice, using write! now and also added validation to make sure each description does not end in a "."

exzachlyvv: We can include whole lines and pretty print ourselves in future PR.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+        // First line is H1.
+        if lines.next() != Some("# Changelog") {
+            return Err(anyhow!("H1 with 'Changelog' is required"));
+        }
+
+        while let Some(mut current_line) = lines.next() {
+            let mut next_line = lines.peek();
+
+            // Parse a whole release
+            if let Some(version_string) = current_line.strip_prefix("## ") {
```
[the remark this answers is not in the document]
exzachlyvv: Yep, I'm made updates to improve validation by throwing exception if we see malformed releases.

crates/xtask/src/changelog.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    /// Converts raw file contents into a Changelog data structure.
+    ///
+    /// Validates requirements in docs/contributing/changelog.md
+    fn parse(contents: String) -> Result<Changelog> {
+        let mut skip_counter = 0;
+        let mut releases: Vec<Release> = Vec::new();
+
+        let mut lines = contents.lines().filter(|line| !line.is_empty()).peekable();
```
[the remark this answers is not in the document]
exzachlyvv: I'm going to leave that as a future enhancement.

ia0 (approved):
Thanks! I've updated the branch to master and added a small commit with nits. I didn't review the changelog.rs file. I'll do that in the follow-up PRs. Also, we should have a PR before then that moves the module from xtask to cli-tools.

PR #587, opened 2024-08-15
Return an error instead of trapping for unimplemented API

### LINKED ISSUE #464: Avoid as much trap as possible in the scheduler
Search for all functions returning `Result<_, Trap>` and fix them if needed.

Good candidates are:
- `MemoryApi::alloc()` should return `World:NotEnough` (or something similar).
- `Applet::{enable,disable}()` should probably return `User::InvalidState`.
- `Scheduler::disable_event()` is similar to above.
- `or_trap!()` should be replaced with `or_fail!()`.

Bad candidates are:
- Wrong memory accesses (e.g. `MemoryApi::{get,get_mut}()`) should continue to `Trap`.

The distinction is not always clear. It would be good to come up with a criteria of whether an operation should trap or error.

This is part of #43.

lukeyeh:
Part of #464

### DIFF
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 ### Minor
 
-- Migrate `or_trap!` to `or_fail!` for when there's no board support
+- Return an error when calling an unimplemented API
 - Support `PlatformVendor` protocol
 - Support `platform::serial()` function
 - Migrate `platform::version()` to the new board and applet APIs
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Minor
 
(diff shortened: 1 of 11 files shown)


crates/scheduler/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 ### Minor
 
+- Migrate `or_trap!` to `or_fail!` for when there's no board support
```
ia0, line 18: nit: `or_trap` and `or_fail` are implementation details, the changelog should be understandable from a user perspective (the PR will be automatically added during release for those who want to see the implementation)

```suggestion
- Return an error when calling an unimplemented API
```

ia0 (approved):
Thanks! I'll apply the fix and merge.

ia0 (approved):

ia0-review (approved):

PR #588, opened 2024-08-23
Changed logging to warn level when web client is built in release mode.

chris-dietz:


ia0 (commented):
That's not really what I meant in https://github.com/google/wasefire/pull/549#pullrequestreview-2209973668:

> If trunk doesn't build in release by default, it's definitely needed to do that. And speaking of release, maybe there's too much logging to the console. Instead of use `wasm_logger::Config::default()` we should use `wasm_logger::Config::new(level)` with appropriate level. This probably needs to be done at compile-time using features (one per `log::Level`) and set in the host runner when building trunk using `log::max_level().to_level()` to know which feature to use. If no feature is passed (i.e. when `to_level()` returns `None`), then logging should be completely disabled. But this should be done in a separate PR.

The idea is to have one feature per log level in `web-client`. Then in `runner-host` when calling `trunk` we check `log::max_level().to_level()` and enable the appropriate feature. If no feature is selected, web-client should not log anything. This lets the user control the level of logging instead of having something hard-coded.

crates/runner-host/crates/web-client/Cargo.toml
```diff
@@ -21,3 +22,11 @@ yew-hooks = "0.3.2"
 clippy.unit-arg = "allow"
 rust.unsafe-op-in-unsafe-fn = "warn"
 rust.unused-crate-dependencies = "warn"
+
+[features]
+web_client_log_level_info = []
+web_client_log_level_debug = []
+web_client_log_level_warn = []
+web_client_log_level_error = []
+web_client_log_level_trace = []
```
ia0, line 31: You need to put this table before the `[lints]` one, because the sync script assumes `[lints]` is the last one. We can also use much simple feature names:

```toml
[features]
# At most one log level can be selected (defaults to off).
log_debug = []
log_error = []
...
```

crates/runner-host/src/main.rs
```diff
@@ -109,9 +109,9 @@ async fn main() -> Result<()> {
                 }
             }
         });
-        let mut trunk = tokio::process::Command::new("../../scripts/wrapper.sh");
-        trunk.args(["trunk", "build", "--release", "crates/web-client/index.html"]);
-        wasefire_cli_tools::cmd::execute(&mut trunk).await?;
+        let mut trunk = std::process::Command::new("../../scripts/wrapper.sh");
```
ia0, line 112: You reverted the `tokio::process` and `--locked` changes while merging.

crates/runner-host/src/main.rs
```diff
@@ -109,9 +109,9 @@ async fn main() -> Result<()> {
...
+        trunk.args(["trunk", "build", "--locked", "--release", "--features=web_client_log_level_warn", "crates/web-client/index.html"]);
```
ia0, line 113: I guess you're still working on computing this automatically.

crates/runner-host/crates/web-client/src/main.rs
```diff
@@ -18,7 +18,27 @@ mod board_components;
 mod console;
 mod hooks;
 
+use cfg_if::cfg_if;
+
 fn main() {
-    wasm_logger::init(wasm_logger::Config::default());
+    wasm_logger::init(wasm_logger::Config::new(get_log_level()));
```
ia0, line 24: This could simply be something like this:
```suggestion
    #[cfg(feature = "log_trace")]
    wasm_logger::init(wasm_logger::Config::new(log::Level::Trace));
    #[cfg(feature = "log_debug")]
    wasm_logger::init(wasm_logger::Config::new(log::Level::Debug));
    #[cfg(feature = "log_info")]
    wasm_logger::init(wasm_logger::Config::new(log::Level::Info));
    #[cfg(feature = "log_warn")]
    wasm_logger::init(wasm_logger::Config::new(log::Level::Warn));
    #[cfg(feature = "log_error")]
    wasm_logger::init(wasm_logger::Config::new(log::Level::Error));
```

Then we don't need the `cfg_if` dependency and when no features are selected, logging is disabled.

PR #589, opened 2024-08-23
[WIP] Implementation of atomic_wait / atomic_notify

resulknad:
We discussed a solution to have multiple threads with a single store while we wait for the [shared-everything-threads](https://github.com/WebAssembly/shared-everything-threads) to land. This is opposed to the `instance-per-thread` solution as implemented in [wasi-threads](https://github.com/WebAssembly/wasi-threads?tab=readme-ov-file#design-choice-instance-per-thread).

This commit is still WIP, but maybe a good point to pause and have a discussion of where we will take this.

Right now it contains:
- Adding support for multi-threaded tests in `spec.rs`
- Thread-safe module instantiation
- Very basic wait / notify implementation

This is somewhat working, it passes the wait notify test cases. Right now we have only ever one thread per module instantiation, which makes things easier to argue about.

I wonder though, in the desired use case, does that assumption still hold?

### DIFF
--- a/crates/interpreter/tests/spec.rs
+++ b/crates/interpreter/tests/spec.rs
@@ -13,9 +13,13 @@
 // limitations under the License.
 
 #![feature(int_roundings)]
+#![feature(get_mut_unchecked)]
 #![allow(unused_crate_dependencies)]
 
+use std::borrow::BorrowMut;
 use std::collections::HashMap;
+use std::sync::Arc;
+use std::{thread, vec};
 
 use lazy_static::lazy_static;
 use wasefire_interpreter::*;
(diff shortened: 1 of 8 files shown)


crates/interpreter/tests/spec.rs
```diff
@@ -24,44 +28,79 @@ use wast::lexer::Lexer;
...
-    for directive in wast.directives {
+    let pool =
+        unsafe { std::slice::from_raw_parts_mut(std::alloc::alloc_zeroed(layout), layout.size()) };
+    let mut env_arc = Arc::new(Env::new(pool));
+    let first_call = root_env.is_some();
+    if root_env.is_some() {
+        env_arc = root_env.unwrap();
+    }
+    let env_arc_cpy = env_arc.clone();
+    let env = unsafe { Arc::<Env<'_>>::get_mut_unchecked(env_arc.borrow_mut()) };
```
ia0, line 41: This is actually unsound. When we spawn multiple threads, they will each get a mutable reference to the environment.

ia0 (commented):
Thanks for the PR! It's much easier to discuss with code.

> Right now we have only ever one thread per module instantiation, which makes things easier to argue about.

Interesting, I was imagining it with a single module instantiation because of the use-case I have in mind (see below). But I understand that to be able to use the existing test suite, we kind of have to stay similar to the instance-per-thread design, since it's written using different modules for different threads.

> I wonder though, in the desired use case, does that assumption still hold?

The use-case I have in mind is something like:
- Each applet has its own store (right now there's only one applet, so one store).
- The store may run multiple threads (in practice, it will probably be at most 2: "main" and "interrupt").
- Threads don't run in parallel, only concurrently (we only target single-core CPUs for now). But if we can avoid using this assumption, we should. In particular, we can assume the interpreter will only switch threads at very specific points (like calling a host function).
- Threads have a priority given by their creation time (most recent has higher priority). The highest priority non-waiting thread runs (if running in parallel, then the top N highest priority non-waiting threads).

We probably would need to write our own tests if we go that way. I don't think we need more than a few of them. This feature is going to be experimental until the shared-everything-threads proposal becomes concrete.

Actually, thinking about it, we might be able to run the test suite with the multiple threads per store. It's just that we won't use it like that. But it should support instantiating different modules and invoking them concurrently.

We can discuss the design in more details in another channel if you want.

PR #595, opened 2024-08-30
Implement Display for Changelog. Hooked up to parsing validation.

exzachlyvv:
Diff will clean up after PR #584 is merged

### DIFF
--- a/crates/xtask/src/changelog.rs
+++ b/crates/xtask/src/changelog.rs
@@ -14,7 +14,7 @@
 
 use core::str;
 use std::collections::BTreeMap;
-use std::fmt::Write;
+use std::fmt::{Display, Write};
 use std::io::BufRead;
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
@@ -23,10 +23,20 @@ use tokio::process::Command;
 use wasefire_cli_tools::{cmd, fs};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
-pub enum ReleaseType {
+pub enum Severity {
     Major,
     Minor,
     Patch,
+}
+
+impl Display for Severity {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Major => write!(f, "Major"),
+            Self::Minor => write!(f, "Minor"),
+            Self::Patch => write!(f, "Patch"),
+        }
+    }
 }
 
 #[derive(Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


crates/xtask/src/changelog.rs
```diff
@@ -29,6 +30,16 @@ pub enum ReleaseType {
     Patch,
 }
 
+impl Display for ReleaseType {
```
ia0, line 33: Unrelated but I think it's worth doing it soon, is to rename `ReleaseType` to `Scope` to be a bit more abstract, because this type is used for 2 things:
- Describe the scope or impact of a change.
- Describe the highest scope of a change between 2 releases.

And the second thing is actually only a consequence of the first, so the name should be more relevant to the first, which is describing the scope/impact of a change.

crates/xtask/src/changelog.rs
```diff
@@ -154,7 +165,28 @@ impl Changelog {
         );
         ensure!(last_release.contents.is_empty(), "The last release must contain no changes");
 
-        Ok(Changelog { releases, skip_counter })
+        let changelog = Changelog { releases, skip_counter };
+        let output_string = format!("{changelog}");
+
+        ensure!(
+            output_string == contents,
+            "Input string does not equal rendered changelog.\n\nIn: {contents}\n\nOut: {contents}"
```
ia0, line 173: It's probably more readable to use the `similar-asserts` crate like [here](https://github.com/google/wasefire/blob/488c3d8e720e47ad6b40f5164d201ddf2b48ed55/crates/wire-derive/src/tests.rs#L36). Use the same version as the wire-derive crate does. Ideally we should have a test for that, but it's not there yet.

crates/xtask/src/changelog.rs
```diff
@@ -154,7 +165,28 @@ impl Changelog {
...
+        );
+
+        Ok(changelog)
+    }
+}
+
+impl Display for Changelog {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let releases_string = self.releases.iter().map(|release| format!("{release}")).join("\n\n");
```
ia0, line 182: This is not efficient to build a string for each version and then join them. We want to output directly to the formatter.
```suggestion
        for release in &self.releases {
            writeln!(f, "{release}\n")?;
        }
```

Also this permits getting rid of `itertools` which is usually not useful.

crates/xtask/src/changelog.rs
```diff
@@ -154,7 +165,28 @@ impl Changelog {
...
+
+        write!(
```
ia0, line 184: Same, we can split this in 2 around the for-loop above.

```rust
writeln!(f, "# Changelog\n")?;
for release in &self.releases { ... }
writeln!(f, "<!-- Increment ...: {skip_counter} -->")
```

crates/xtask/src/changelog.rs
```diff
@@ -171,6 +203,21 @@ impl Release {
...
 
+impl Display for Release {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "## {}", self.version)?;
+
+        for (release_type, descriptions) in &self.contents {
+            let description_string =
+                descriptions.iter().map(|description| format!("- {description}")).join("\n");
+
+            write!(f, "\n\n### {release_type}\n\n{description_string}")?;
```
ia0, line 214: Same here.
```suggestion
            writeln!(f, "### {scope}\n")?;
            for description in descriptions {
                writeln!(f, "- {description}")?;
            }
```

ia0 (commented):
Thanks! Looks good to me modulo comments.

crates/xtask/src/changelog.rs
ia0: Actually, we could also use the name `Severity` since it's used [here](https://doc.rust-lang.org/cargo/reference/semver.html).

ia0:
By the way, there will be merge conflicts due to #601 when we'll merge into main. We'll have to decide how to do that since GitHub doesn't support reviewing merge commits.

crates/xtask/src/changelog.rs
exzachlyvv: Ack. Should we keep Cargo.lock unchanged until we can clean up merging.

ia0:
You can now merge or rebase this on dev/changelog. You might want to also move changelog.rs to cli-tools too.

exzachlyvv:
Cleaned up diff and squashed commits. Let's move to cli-tools in separate PR. Trying to keep scopes small.

ia0 (approved):
Thanks! I simplified a match in the review commit.

PR #598, opened 2024-09-02
Combine Label values in Frame

zhouwfang:
This is the first step towards adding a value stack in `Thread`. The next step is to merge `locals` and `labels_values` in `Frame`. After that, we can create a value stack in `Thread` by combining the values from the `Frame`'s.

#46

### DIFF
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -1016,10 +1016,7 @@ impl<'m> Thread<'m> {
     }
 
     fn pop_value(&mut self) -> Val {
-        // There may be multiple drops in wasm.
-        if self.label().values_cnt >= 1 {
-            self.label().values_cnt -= 1;
-        }
+        self.label().values_cnt -= 1;
         self.values().pop().unwrap()
     }
 
@@ -1033,14 +1030,14 @@ impl<'m> Thread<'m> {
     }
 
     fn push_label(&mut self, type_: FuncType<'m>, kind: LabelKind<'m>) {
-        let values = self.pop_values(type_.params.len());
+        let params_len = type_.params.len();
         let arity = match kind {
             LabelKind::Block | LabelKind::If => type_.results.len(),
-            LabelKind::Loop(_) => type_.params.len(),
+            LabelKind::Loop(_) => params_len,
         };
-        let label = Label { arity, kind, values_cnt: values.len() };
+        self.label().values_cnt -= params_len;
+        let label = Label { arity, kind, values_cnt: params_len };
         self.labels().push(label);
-        self.values().extend_from_slice(&*values);
     }
 
     fn pop_label(mut self, inst: &mut Instance<'m>, l: LabelIdx) -> ThreadResult<'m> {
@@ -1048,17 +1045,18 @@ impl<'m> Thread<'m> {
         if i == 0 {
             return self.exit_frame();
         }
-        let values = core::mem::take(self.values());
         let frame = self.frame();
+        // TODO: Store prefix sums of the labels' values_cnt in Frame to quickly compute
+        // label_values_cnt_before.
         let label_values_cnt_before: usize =
-            frame.labels[0 .. i].iter().map(|l| l.values_cnt).sum();
-        let label_values_cnt_after: usize =
-            frame.labels[i + 1 ..].iter().map(|l| l.values_cnt).sum();
+            frame.labels[0 .. i].iter().map(|label| label.values_cnt).sum();
         let Label { arity, kind, values_cnt } = frame.labels.drain(i ..).next().unwrap();
-        self.values().extend_from_slice(&values[0 .. label_values_cnt_before]);
+        // TODO: Is the following line necessary?
+        self.values().drain(label_values_cnt_before + values_cnt ..);
         self.values()
-            .extend_from_slice(&values[values.len() - label_values_cnt_after - values_cnt ..]);
-        self.label().values_cnt = values_cnt;
+            .drain(label_values_cnt_before .. label_values_cnt_before + values_cnt - arity);
+        // The (i-1)-th label
+        self.label().values_cnt += arity;
         match kind {
             LabelKind::Loop(pos) => unsafe { self.parser.restore(pos) },
             LabelKind::Block | LabelKind::If => self.skip_to_end(inst, l),
@@ -1069,8 +1067,8 @@ impl<'m> Thread<'m> {
     fn exit_label(mut self) -> ThreadResult<'m> {
         let frame = self.frame();
         let label = frame.labels.pop().unwrap();
-        let popped_values: Vec<Val> =
-            frame.labels_values.drain(frame.labels_values.len() - label.values_cnt ..).collect();
+        let popped_values =
+            frame.labels_values.split_off(frame.labels_values.len() - label.values_cnt);
         if frame.labels.is_empty() {
             let frame = self.frames.pop().unwrap();
             debug_assert_eq!(label.values_cnt, frame.arity);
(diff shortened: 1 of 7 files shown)


crates/interpreter/src/exec.rs
```diff
@@ -1055,15 +1069,18 @@ impl<'m> Thread<'m> {
     fn exit_label(mut self) -> ThreadResult<'m> {
         let frame = self.frame();
         let label = frame.labels.pop().unwrap();
+        let popped_values: Vec<Val> =
+            frame.labels_values.drain(frame.labels_values.len() - label.values_cnt ..).collect();
```
ia0, line 1073:
```suggestion
        let popped_values = frame.labels_values.split_off(frame.labels_values.len() - label.values_cnt);
```

crates/interpreter/src/exec.rs
```diff
@@ -1011,9 +1012,14 @@ impl<'m> Thread<'m> {
 
     fn push_values(&mut self, values: &[Val]) {
         self.values().extend_from_slice(values);
+        self.label().values_cnt += values.len();
     }
 
     fn pop_value(&mut self) -> Val {
+        // There may be multiple drops in wasm.
```
ia0, line 1019: I'm surprised by this logic. Could you provide an example when you hit this issue?

crates/interpreter/src/exec.rs
```diff
@@ -989,7 +989,7 @@ impl<'m> Thread<'m> {
     }
 
     fn values(&mut self) -> &mut Vec<Val> {
-        &mut self.label().values
+        &mut self.frame().labels_values
```
ia0, line 992: We're changing a bit the semantics here. Ideally the return value is never swapped and only pushed/popped.

crates/interpreter/src/exec.rs
```diff
@@ -1032,8 +1038,9 @@ impl<'m> Thread<'m> {
             LabelKind::Block | LabelKind::If => type_.results.len(),
             LabelKind::Loop(_) => type_.params.len(),
         };
-        let label = Label { arity, kind, values };
+        let label = Label { arity, kind, values_cnt: values.len() };
         self.labels().push(label);
+        self.values().extend_from_slice(&*values);
```
ia0, line 1043: This is typically an example where we don't need to pop to push directly after (and the whole point of this effort). It should be possible to remove line 1036 and not add this line.

crates/interpreter/src/exec.rs
```diff
@@ -1043,8 +1050,15 @@ impl<'m> Thread<'m> {
         }
         let values = core::mem::take(self.values());
```
ia0, line 1051: Mmm... so much for "I hope the return value of `values()` is not swapped".

crates/interpreter/src/exec.rs
```diff
@@ -1043,8 +1050,15 @@ impl<'m> Thread<'m> {
...
         let frame = self.frame();
-        let Label { arity, kind, .. } = frame.labels.drain(i ..).next().unwrap();
-        self.values().extend_from_slice(&values[values.len() - arity ..]);
+        let label_values_cnt_before: usize =
+            frame.labels[0 .. i].iter().map(|l| l.values_cnt).sum();
+        let label_values_cnt_after: usize =
+            frame.labels[i + 1 ..].iter().map(|l| l.values_cnt).sum();
+        let Label { arity, kind, values_cnt } = frame.labels.drain(i ..).next().unwrap();
+        self.values().extend_from_slice(&values[0 .. label_values_cnt_before]);
```
ia0, line 1058: This is extremely costly. We don't want to swap `values()`. We want to directly drain the proper range, which should be something like `label_values_cnt_before .. values.len() - label_values_cnt_after - values_cnt`.

crates/interpreter/src/exec.rs
```diff
@@ -1055,15 +1069,18 @@ impl<'m> Thread<'m> {
...
-            debug_assert_eq!(label.values.len(), frame.arity);
+            debug_assert_eq!(label.values_cnt, frame.arity);
             if self.frames.is_empty() {
-                return ThreadResult::Done(label.values);
+                return ThreadResult::Done(popped_values);
             }
             unsafe { self.parser.restore(frame.ret) };
         }
-        self.values().extend_from_slice(&label.values);
+        self.values().extend_from_slice(&popped_values);
```
ia0, line 1082: This looks again like a case that can be optimized by not draining and not extending. It might be a bit more complicated due to the case where we're exiting the last label of a frame and thus exiting the frame.

ia0 (commented):
Yes, I think the overall plan looks good.

crates/interpreter/src/exec.rs
```diff
@@ -1043,8 +1050,15 @@ impl<'m> Thread<'m> {
         }
         let values = core::mem::take(self.values());
         let frame = self.frame();
-        let Label { arity, kind, .. } = frame.labels.drain(i ..).next().unwrap();
```
zhouwfang, line 1046: @ia0 Could you clarify in this line (1046) in the original code the reason to call `drain(i ..)` instead of `drain(l ..)`? `l` is the label index. Thanks.

ia0: Because `l` starts counting from the end of the vector and `drain()` expected an index from the start of the vector. If Rust had some Python syntax with negative indices (which would probably be a bad idea), then it could have been `drain(-l ..)` (or `-l -1` depending of the actual semantics).

zhouwfang: Is it possible to give a conceptual explanation on why it is natural for `l` to be indexed reversely?

ia0: This is WASM specification. They probably did that because it permits copy pasting code without having to change all indices, because they don't depend on their label depth.

crates/interpreter/src/exec.rs
```diff
@@ -1041,10 +1045,18 @@ impl<'m> Thread<'m> {
...
-        self.values().extend_from_slice(&values[values.len() - arity ..]);
+        // TODO: Store prefix sums of the labels' values_cnt in Frame to quickly compute
+        // label_values_cnt_before.
+        let label_values_cnt_before: usize =
+            frame.labels[0 .. i].iter().map(|label| label.values_cnt).sum();
+        let Label { arity, kind, values_cnt } = frame.labels.drain(i ..).next().unwrap();
+        // TODO: Is the following line necessary?
+        self.values().drain(label_values_cnt_before + values_cnt ..);
+        self.values()
+            .drain(label_values_cnt_before .. label_values_cnt_before + values_cnt - arity);
```
zhouwfang, line 1057: @ia0 Would like to get some clarification that will help me fix the remaining 9 failing tests. Thanks!

1) From `values[values.len() - arity ..]` in line 1047 in the original code, it looks like `arity` is stored after the other values. Two questions:
1.1) I don't see how `arity` is stored in `values` in the first place, by reading [push_label](https://github.com/google/wasefire/blob/main/crates/interpreter/src/exec.rs#L1029) and [Frame::new](https://github.com/google/wasefire/blob/main/crates/interpreter/src/exec.rs#L1408).
1.2) If `arity` is stored in `values`, my "intuition" is that it should be stored in front because it could represent the result.

2) See the TODO on line 1054 in the new code. Do we also need to remove the labels after the `i`-th one? Why or why not?

crates/interpreter/src/exec.rs
```diff
@@ -1055,15 +1067,18 @@ impl<'m> Thread<'m> {
...
-            debug_assert_eq!(label.values.len(), frame.arity);
+            debug_assert_eq!(label.values_cnt, frame.arity);
             if self.frames.is_empty() {
-                return ThreadResult::Done(label.values);
+                return ThreadResult::Done(popped_values);
             }
             unsafe { self.parser.restore(frame.ret) };
         }
-        self.values().extend_from_slice(&label.values);
+        self.values().extend(popped_values);
```
ia0, line 1080: This can also be optimized out. It only needs to be done when changing frames (until that part is also handled).

ia0 (commented):
I'll be pushing the changes. It's faster than iterate. Feel free to try to rediscover them on your own.

ia0 (approved):

PR #599, opened 2024-09-05
Add platform protocol transport using tokio traits

ia0:
Part of #577 because the runner host without USB still needs to support the platform protocol.


crates/runner-host/src/main.rs
```diff
@@ -90,12 +113,31 @@ async fn main() -> Result<()> {
...
+        let sender = sender.clone();
+        move |event: Event| drop(sender.try_send(event.into()))
+    };
+    #[cfg(feature = "tcp")]
+    let pipe = wasefire_protocol_tokio::Pipe::new_tcp(flags.tcp_addr, push).await.unwrap();
+    #[cfg(feature = "unix")]
+    let pipe = {
+        let pipe = wasefire_protocol_tokio::Pipe::new_unix(&flags.unix_path, push).await.unwrap();
+        let unix_path = flags.unix_path.clone();
+        CLEANUP.lock().unwrap().push(Box::new(move || {
```
ia0-review, line 128: Maybe we should use a `cleanup` module.

scripts/ci.sh
```diff
@@ -34,7 +34,7 @@ x ./scripts/ci-taplo.sh
 x ./scripts/ci-applets.sh
 x ./scripts/ci-runners.sh
 x ./scripts/ci-tests.sh
-x ./scripts/hwci.sh host --no-default-features
+x ./scripts/hwci.sh host --no-default-features --features=unix
```
ia0-review, line 37: Don't forget to update the workflow.

ia0-review (changes requested):

ia0-review (approved):

PR #603, opened 2024-09-06
Simplify protocol API macro

ia0:


crates/protocol/src/lib.rs
```diff
@@ -231,4 +229,6 @@ api! {
 
     /// Calls a vendor-specific platform command.
     6 [2 -] PlatformVendor: &'a [u8] => &'a [u8],
+
+    next 7 [2 - ]
```
ia0-review, line 233: Maybe the version should also be the next one.

ia0-review (approved):

PR #605, opened 2024-09-07
Merge locals and label values in Frame

zhouwfang:
Continuation of #598 towards adding a value stack in `Thread`.

#46


crates/interpreter/src/exec.rs
```diff
@@ -1061,14 +1061,15 @@ impl<'m> Thread<'m> {
...
+            frame.values.drain(0 .. frame.locals_cnt);
+            debug_assert_eq!(frame.values.len(), label.values_cnt);
+            debug_assert_eq!(frame.values.len(), frame.arity);
             if self.frames.is_empty() {
-                return ThreadResult::Done(frame.labels_values);
+                return ThreadResult::Done(frame.values);
             }
             unsafe { self.parser.restore(frame.ret) };
-            self.values().extend(frame.labels_values);
+            self.values().extend(frame.values);
```
zhouwfang, line 1072: @ia0 You mentioned earlier that this line could be optimized. Here `self.values()` refers to the values of the frame before the popped `frame`, and it does not know about `frame.values`. Could you explain more? Thanks.

crates/interpreter/src/exec.rs
```diff
@@ -1061,14 +1061,15 @@ impl<'m> Thread<'m> {
         let frame = self.frame();
         let label = frame.labels.pop().unwrap();
         if frame.labels.is_empty() {
-            let frame = self.frames.pop().unwrap();
-            debug_assert_eq!(frame.labels_values.len(), label.values_cnt);
-            debug_assert_eq!(frame.labels_values.len(), frame.arity);
+            let mut frame = self.frames.pop().unwrap();
+            frame.values.drain(0 .. frame.locals_cnt);
```
ia0, line 1065: I'm not sure what's the best here. We might want to try both options (and look what others are doing). It looks like we can either:
- keep the locals out of the stack (thus saving this drain)
- keep the locals inside the stack (thus saving popping them when calling a function)

The main difference is that draining is proportional in the number of remaining elements, while popping is proportional to what is popped. But I expect both of those numbers to be rather small, so maybe it doesn't matter. Just something to keep in mind once we have something working and start doing benchmarks.

crates/interpreter/src/exec.rs
ia0: This will get optimized once we use a single value stack for the whole thread (not just per function frame). But this will only be true if we go with the "keep locals outside the stack" option (discussed above). Otherwise we'll have to drain.

ia0 (approved):
LGTM, although it's not clear yet whether we want to keep this in the end.

crates/interpreter/src/exec.rs
zhouwfang: Thanks for the explanation. I'll add a TODO for this tradeoff in the next PR.

PR #608, opened 2024-09-16
Add value stack in Thread

zhouwfang:
Reverted #605 and keep values separate from locals for now.

#46


crates/interpreter/src/exec.rs
```diff
@@ -989,7 +992,11 @@ impl<'m> Thread<'m> {
     }
 
     fn values(&mut self) -> &mut Vec<Val> {
-        &mut self.frame().labels_values
+        &mut self.values
+    }
+
+    fn last_frame_values_cnt(&mut self) -> usize {
+        self.frame().labels.iter().map(|label| label.values_cnt).sum()
     }
```
zhouwfang, line 1000: @ia0 The reason why this is not computed on the fly is that I suspect it could be obtained from the side table. Please let me know if this is not the case.

crates/interpreter/src/exec.rs
```diff
@@ -1029,6 +1036,15 @@ impl<'m> Thread<'m> {
         values
     }
 
+    fn only_pop_values(&mut self, n: usize) -> Vec<Val> {
```
zhouwfang, line 1039: Just found a minor improvement that I could make later: to unify this function and the `pop_values(&mut self, n: usize)` above, create a `pop_values(&mut self, n: usize, pop_once: fn(&mut self) -> Val)` by passing the function used in `values.push()` as `pop_once`.

zhouwfang: Done by generics.

crates/interpreter/src/exec.rs
ia0: Yes, I also believe this could be known statically and doesn't need to be optimized until we get the side table computed. In particular, I expect the `values_cnt` fields to disappear. So all code mentioning `values_cnt` fields should also disappear.

crates/interpreter/src/exec.rs
```diff
@@ -728,7 +731,7 @@ enum ThreadResult<'m> {
 
 impl<'m> Thread<'m> {
     fn new(parser: Parser<'m>, frames: Vec<Frame<'m>>) -> Thread<'m> {
```
ia0, line 733: I'm not sure why I made this take a stack of frames, it seems a thread is always created with exactly one frame. We could change the type of this function to take `frame: Frame<'m>` instead of `Vec<Frame<'m>>`. This could be done in another PR against the `main` branch.

crates/interpreter/src/exec.rs
```diff
@@ -1021,12 +1028,11 @@ impl<'m> Thread<'m> {
     }
 
     fn pop_values(&mut self, n: usize) -> Vec<Val> {
-        let mut values = Vec::new();
-        for _ in 0 .. n {
-            values.push(self.pop_value());
-        }
-        values.reverse();
-        values
+        pop_values(n, || self.pop_value())
```
ia0: I'm not sure why I wrote this function like that, but it seems `Vec::split_off` would work (it probably didn't exist back then).

crates/interpreter/src/exec.rs
```diff
@@ -1058,25 +1064,27 @@ impl<'m> Thread<'m> {
     }
 
     fn exit_label(mut self) -> ThreadResult<'m> {
+        let values_cnt = self.last_frame_values_cnt();
```
ia0: I simplified this a bit. We don't need to compute this because when we enter the `if`, we know there was exactly one label.

crates/interpreter/src/exec.rs
```diff
@@ -1021,12 +1028,13 @@ impl<'m> Thread<'m> {
...
-        for _ in 0 .. n {
-            values.push(self.pop_value());
-        }
-        values.reverse();
-        values
+        self.label().values_cnt -= n;
+        self.only_pop_values(n)
+    }
+
+    fn only_pop_values(&mut self, n: usize) -> Vec<Val> {
```
ia0, line 1035: I expect this function to disappear once `values_cnt` disappears, since it becomes equivalent to `pop_values`.

ia0 (approved):

crates/interpreter/src/exec.rs
zhouwfang: Done in #612.

PR #609, opened 2024-09-20
Support platform update in platform protocol

ia0:
This is part of #577.


crates/cli-tools/src/action.rs
```diff
@@ -140,21 +141,92 @@ impl PlatformList {
...
+        connection.call::<S>(Request::Start { dry_run: *dry_run }).await?.get();
+        for chunk in payload.chunks(*chunk_size) {
+            connection.call::<S>(Request::Write { chunk }).await?.get();
+            progress.inc(chunk.len() as u64);
+        }
+        progress.finish();
+        if *dry_run {
+            connection.call::<S>(Request::Finish).await?.get();
+        } else {
+            final_call::<S>(connection, Request::Finish, |x| Ok(*x.get())).await?;
```
ia0-review, line 202: We actually don't want `Ok`. We should panic if reachable.

crates/cli-tools/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 //! This library is also used for the internal maintenance CLI of Wasefire called xtask.
 
 #![feature(async_fn_track_caller)]
+#![feature(never_type)]
```
ia0-review, line 20: Looks like a left-over, but should be useful again when fixing the other comment.

crates/cli/src/main.rs
```diff
@@ -64,8 +64,17 @@ enum Action {
     },
     PlatformList(action::PlatformList),
 
-    /// Updates a connected platform.
-    PlatformUpdate,
+    PlatformUpdateMetadata {
```
ia0-review, line 67: This should have documentation.

crates/protocol/src/connection.rs
```diff
@@ -47,8 +48,11 @@ pub trait ConnectionExt: Connection {
         &mut self, request: S::Request<'_>,
     ) -> impl Future<Output = anyhow::Result<Yoke<S::Response<'static>>>> {
         async {
-            self.send(&S::request(request)).await.context("sending request")?;
-            self.receive::<S>().await.context("receiving response")
+            let result: anyhow::Result<_> = try {
+                self.send(&S::request(request)).await.context("sending request")?;
+                self.receive::<S>().await.context("receiving response")?
+            };
+            result.with_context(|| format!("calling {}", S::NAME))
```
ia0-review, line 55: Let's inline the name in the 2 contexts above.

crates/protocol/src/lib.rs
```diff
@@ -30,6 +30,7 @@
 #![feature(doc_auto_cfg)]
 #![feature(macro_metavar_expr)]
 #![feature(never_type)]
+#![feature(try_blocks)]
```
ia0-review, line 33: revert

crates/protocol/src/lib.rs
```diff
@@ -224,11 +229,17 @@ api! {
     /// Starts a direct tunnel with an applet.
     4 [0 -] AppletTunnel: applet::Tunnel<'a> => (),
 
-    /// Returns platform information (e.g. serial and version).
+    /// Returns platform information.
```
ia0-review, line 232: revert

ia0-review (changes requested):

PR #611, opened 2024-09-20
Support waiting on a call with optional response

ia0:
Part of #577. This is going to be used by `AppletExitStatus` too.


crates/cli-tools/src/action.rs
```diff
@@ -90,27 +91,72 @@ pub struct AppletRpc {
...
-    retries: usize,
+    #[clap(flatten)]
+    wait: Wait,
 }
 
 impl AppletRpc {
     pub async fn run(self, connection: &mut dyn Connection) -> Result<()> {
-        let AppletRpc { applet, rpc, retries } = self;
+        let AppletRpc { applet, rpc, mut wait } = self;
+        wait.ensure_wait();
```
ia0-review, line 101: Let's move this line closer to the `bail!()`.

ia0-review (changes requested):

ia0-review (approved):

ia0-review (dismissed):

ia0-review (approved):

PR #616, opened 2024-09-26
Make validation return side tables

zhouwfang:
Make validation return side tables, and store them in module.

#46


crates/interpreter/src/valid.rs
```diff
@@ -17,62 +17,22 @@ use alloc::vec;
...
-
-    fn view(self) -> SideTableEntryView {
-        let delta_ip = from_signed_field(Self::DELTA_IP_MASK, self.0);
-        let delta_stp = from_signed_field(Self::DELTA_STP_MASK, self.0);
-        let val_cnt = from_field(Self::VAL_CNT_MASK, self.0);
-        let pop_cnt = from_field(Self::POP_CNT_MASK, self.0);
-        SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt }
-    }
-}
+type CheckedResult = MResult<(), Check>;
```
ia0, line 32: Why rename this?

crates/interpreter/src/valid.rs
```diff
@@ -17,62 +17,22 @@ use alloc::vec;
...
+type CheckedResultAsSideTables = MResult<Vec<SideTableEntry>, Check>;
```
ia0, line 33: This is used only once and a very long name. Let's just inline its definition which is much more readable than the name.

crates/interpreter/src/valid.rs
```diff
@@ -195,7 +155,7 @@ impl<'m> Context<'m> {
             check(parser.is_empty())?;
         }
         self.check_section(parser, SectionId::Custom)?;
-        check(parser.is_empty())
+        check_and_return(parser.is_empty(), vec![])
```
ia0, line 158: You don't need this `check_and_return` function. The following is much more readable by having small things do simple but precise things, and relying on composition to build more complex things (like unix philosophy).

```suggestion
        check(parser.is_empty())?;
        Ok(vec![]) // TODO(dev/fast-interp): implement
```

crates/interpreter/src/module.rs
```diff
@@ -27,12 +28,14 @@ use crate::*;
 pub struct Module<'m> {
     binary: &'m [u8],
     types: Vec<FuncType<'m>>,
+    // TODO(dev/fast-interp): Change to `&'m [SideTableEntry]` when making it persistent in flash.
+    side_tables: Vec<SideTableEntry>,
```
ia0, line 32: There's at least 2 options here: `[SideTableEntry]` or `[[SideTableEntry]]`. The first one assumes that each function would have an index to its first entry (and the side tables of all functions are concatenated). That's probably the simplest once we persist to flash. The second option is one side table `[SideTableEntry]` per function, hence the second layer. The side table of a function would be at the same outer index as the function index.

crates/interpreter/src/valid.rs
zhouwfang: To be consistent with `CheckedResultAsSideTables`. Reverted as the latter is inlined.

crates/interpreter/src/module.rs
zhouwfang: Good catch. I intended to make it a 2D vector as `side_tables` was plural, but forgot to do so.

crates/interpreter/src/module.rs
```diff
@@ -65,8 +69,10 @@ impl ImportDesc {
 impl<'m> Module<'m> {
     /// Validates a WASM module in binary format.
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
-        validate(binary)?;
-        Ok(unsafe { Self::new_unchecked(binary) })
+        let side_tables = validate(binary)?;
+        let mut module = unsafe { Self::new_unchecked(binary) };
```
ia0, line 73: Ideally `new_unchecked` should take the side tables, but this will break the scheduler. But this issue is temporary until we persist the side tables.

ia0 (approved):

PR #617, opened 2024-09-26
Support applet install and uninstall

### LINKED ISSUE #577: Support dynamic loading and unloading of applets

ia0:
Fixes #577


crates/cli-tools/CHANGELOG.md
```diff
@@ -8,9 +8,15 @@
...
+- Add `action::PlatformLock` for locking a platform protocol
+- Expose `action::Transfer` for transfers from host to device
+- Add `action::Wait::set_period()` and `impl Default for action::Wait`
+- Add `action::AppletExitStatus` to get the applet exit status
+- Add `action::Applet{Install,Uninstall}` for simple applet management
+- Add `action::PlatformApiVersion` to get a platform API version
 - Change the flags of `action::AppletRpc` to use `action::Wait`
 - Add `action::Wait` for commands returning an optional response
-- Increase the default connection timeout from 1 to 5 seconds
+- Change the default connection timeout from 1 seconds to infinite (0 seconds)
```
ia0-review, line 19: typo

crates/cli-tools/CHANGELOG.md
```diff
@@ -8,9 +8,15 @@
 
 ### Minor
 
+- Add `action::PlatformLock` for locking a platform protocol
+- Expose `action::Transfer` for transfers from host to device
+- Add `action::Wait::set_period()` and `impl Default for action::Wait`
```
ia0-review, line 13: already contained below

ia0-review (changes requested):
Only reviewed up to and including `crates/cli`. Will do another iteration after fixes.

ia0-review (approved):

ia0:
There are 2 issues:
- Footprint logic must be updated to take into account that wasm applets are not embedded anymore.
- Seems like using unix socket on Github doesn't work (maybe use something else than `/tmp/wasefire`).

ia0:
There's a few issues that might be worse fixing after this PR:
- `protocol-usb` is used with both `device` and `host`. The feature documentation in `Cargo.toml` should be fixed and a test with both features should be added to `test.sh`.
- `ensure_applet` assumes `thumbv7em-none-eabi` which is not always true. We use `i686-unknown-linux-gnu` in `scheduler` and `runner-host`. This should be cleaned up to use an appropriate library.
- https://github.com/a1ien/rusb/issues/211 may either be related to the above 2 points or something else (but it seems the runner image is the same as yesterday where the test was passing, so it must be a change in this PR and it must be specific to `runner-host` since `scheduler` is fine, so probably the first point rather than the second point)

ia0-review (approved):

ia0:
Ok the issues have been understood: 
- #619 fixes the first issue 
- `cargo check` doesn't care what `libapplet.a` is as long as it exists, so we should just touch it
- the issue was that we compiled `rusb` in a device, this is fixed by #619 (initially fixed by #618 but the wrong fix)

ia0:
Sending #620 for the remaining issue, essentially fixing the known issues introduced by this PR.

PR #626, opened 2024-10-01
Compute delta_ip and delta_stp in side table

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -18,7 +18,7 @@ use crate::error::*;
 #[allow(dead_code)] // TODO(dev/fast-interp)
 #[derive(Default, Copy, Clone, Debug)]
 #[repr(transparent)]
-pub struct SideTableEntry(u32);
+pub struct SideTableEntry(u64);
 
 pub struct SideTableEntryView {
     /// The amount to adjust the instruction pointer by if the branch is taken.
@@ -26,17 +26,17 @@ pub struct SideTableEntryView {
     /// The amount to adjust the side-table pointer by if the branch is taken.
     pub delta_stp: i32,
     /// The number of values that will be copied if the branch is taken.
-    pub val_cnt: u32,
+    pub val_cnt: u64,
     /// The number of values that will be popped if the branch is taken.
-    pub pop_cnt: u32,
+    pub pop_cnt: u64,
 }
 
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
-    const DELTA_IP_MASK: u32 = 0x0000ffff;
-    const DELTA_STP_MASK: u32 = 0x003f0000;
-    const VAL_CNT_MASK: u32 = 0x07c00000;
-    const POP_CNT_MASK: u32 = 0xf8000000;
+    const DELTA_IP_MASK: u64 = 0xffff;
+    const DELTA_STP_MASK: u64 = 0xffff << 16;
+    const VAL_CNT_MASK: u64 = 0xffff << 32;
+    const POP_CNT_MASK: u64 = 0xffff << 48;
 
     pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
         let mut fields = 0;
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -33,10 +33,10 @@ pub struct SideTableEntryView {
 
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
-    const DELTA_IP_MASK: u32 = 0x0000ffff;
-    const DELTA_STP_MASK: u32 = 0x003f0000;
-    const VAL_CNT_MASK: u32 = 0x07c00000;
-    const POP_CNT_MASK: u32 = 0xf8000000;
+    const DELTA_IP_MASK: u32 = 0x7ff;
+    const DELTA_STP_MASK: u32 = 0x7ffff800; // 0x7fff << 11
+    const VAL_CNT_MASK: u32 = 0x78000000; // 0x7 << 26
+    const POP_CNT_MASK: u32 = 0x60000000; // 0x7 << 29
 
     pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
         let mut fields = 0;
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -33,10 +33,10 @@ pub struct SideTableEntryView {
 
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
-    const DELTA_IP_MASK: u32 = 0x7ff;
-    const DELTA_STP_MASK: u32 = 0x7ffff800; // 0x7fff << 11
-    const VAL_CNT_MASK: u32 = 0x78000000; // 0x7 << 26
-    const POP_CNT_MASK: u32 = 0x60000000; // 0x7 << 29
+    const DELTA_IP_MASK: u32 = 0x0000ffff;
+    const DELTA_STP_MASK: u32 = 0x7fff0000;
+    const VAL_CNT_MASK: u32 = 80000000;
+    const POP_CNT_MASK: u32 = 80000000;
 
     pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
         let mut fields = 0;
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -34,9 +34,9 @@ pub struct SideTableEntryView {
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
     const DELTA_IP_MASK: u32 = 0x0000ffff;
-    const DELTA_STP_MASK: u32 = 0x7fff0000;
-    const VAL_CNT_MASK: u32 = 80000000;
-    const POP_CNT_MASK: u32 = 80000000;
+    const DELTA_STP_MASK: u32 = 0x003f0000;
+    const VAL_CNT_MASK: u32 = 0x07c00000;
+    const POP_CNT_MASK: u32 = 0xf8000000;
 
     pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
         let mut fields = 0;
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -26,9 +26,9 @@ pub struct SideTableEntryView {
     /// The amount to adjust the side-table pointer by if the branch is taken.
     pub delta_stp: i32,
     /// The number of values that will be copied if the branch is taken.
-    pub val_cnt: u64,
+    pub val_cnt: u32,
     /// The number of values that will be popped if the branch is taken.
-    pub pop_cnt: u64,
+    pub pop_cnt: u32,
 }
 
 #[allow(dead_code)] // TODO(dev/fast-interp)
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -38,7 +38,7 @@ impl SideTableEntry {
     const VAL_CNT_MASK: u32 = 0x07c00000;
     const POP_CNT_MASK: u32 = 0xf8000000;
 
-    fn new(view: SideTableEntryView) -> Result<Self, Error> {
+    pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
         let mut fields = 0;
         fields |= into_signed_field(Self::DELTA_IP_MASK, view.delta_ip)?;
         fields |= into_signed_field(Self::DELTA_STP_MASK, view.delta_stp)?;
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -420,6 +420,10 @@ impl SideTableBranch<'_> {
     fn delta_ip(end: &[u8], start: &[u8]) -> i32 {
         (end.as_ptr() as isize - start.as_ptr() as isize) as i32
     }
+
+    fn delta_stp(end: usize, start: usize) -> i32 {
+        (end - start) as i32
+    }
 }
 
 #[derive(Debug, Default)]
@@ -437,7 +441,7 @@ struct Label<'m> {
 enum LabelKind<'m> {
     #[default]
     Block,
-    Loop(&'m [u8]),
+    Loop(SideTableBranch<'m>),
     If,
 }
 
@@ -518,7 +522,13 @@ impl<'a, 'm> Expr<'a, 'm> {
                 self.push_label(self.blocktype(&b)?, LabelKind::Block)?;
             }
             Loop(b) => {
-                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.parser.save()))?;
+                self.push_label(
+                    self.blocktype(&b)?,
+                    LabelKind::Loop(SideTableBranch {
+                        parser: self.parser.save(),
+                        side_table: self.side_table.len(),
+                    }),
+                )?;
             }
             If(b) => {
                 self.pop_check(ValType::I32)?;
@@ -801,8 +811,8 @@ impl<'a, 'm> Expr<'a, 'm> {
         assert!(entry.is_none());
         *entry = Some(SideTableEntryView {
             delta_ip: SideTableBranch::delta_ip(current_branch.parser, parser),
+            delta_stp: SideTableBranch::delta_stp(current_branch.side_table, *side_table),
             // TODO(dev/fast-interp): Compute the fields below.
-            delta_stp: 0,
             val_cnt: 0,
             pop_cnt: 0,
         });
@@ -821,11 +831,11 @@ impl<'a, 'm> Expr<'a, 'm> {
                 label.type_.results
             }
             LabelKind::If => label.type_.results,
-            LabelKind::Loop(parser) => {
+            LabelKind::Loop(SideTableBranch { parser, side_table }) => {
                 self.side_table.push(Some(SideTableEntryView {
                     delta_ip: SideTableBranch::delta_ip(branch.parser, parser),
+                    delta_stp: SideTableBranch::delta_stp(branch.side_table, side_table),
                     // TODO(dev/fast-interp): Compute the fields below.
-                    delta_stp: 0,
                     val_cnt: 0,
                     pop_cnt: 0,
                 }));
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -406,24 +406,62 @@ struct Expr<'a, 'm> {
     is_body: bool,
     locals: Vec<ValType>,
     labels: Vec<Label<'m>>,
-    // TODO(dev/fast-interp): Remove `Option` after `side_table` is verified.
-    side_table: Vec<Option<SideTableEntryView>>,
+    side_table: SideTable,
+}
+
+#[derive(Default)]
+struct SideTable {
+    // TODO(dev/fast-interp): Consider removing `Option` if confident.
+    entries: Vec<Option<SideTableEntryView>>,
+}
+
+impl SideTable {
+    fn save(&self) -> usize {
+        self.entries.len()
+    }
+
+    fn branch(&mut self) {
+        self.entries.push(None);
+    }
+
+    fn stitch(&mut self, source: SideTableBranch, target: SideTableBranch) -> CheckResult {
+        let delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
+        let delta_stp = Self::delta(source, target, |x| x.side_table as isize)?;
+        // TODO(dev/fast-interp): Compute the fields below.
+        let val_cnt = 0;
+        let pop_cnt = 0;
+        let entry = &mut self.entries[source.side_table];
+        assert!(entry.is_none());
+        *entry = Some(SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt });
+        Ok(())
+    }
+
+    fn delta(
+        source: SideTableBranch, target: SideTableBranch, field: fn(SideTableBranch) -> isize,
+    ) -> MResult<i32, Check> {
+        let source = field(source);
+        let target = field(target);
+        let Some(delta) = target.checked_sub(source) else {
+            #[cfg(feature = "debug")]
+            eprintln!("side-table subtraction overflow {target} - {source}");
+            return Err(unsupported(if_debug!(Unsupported::SideTable)));
+        };
+        i32::try_from(delta).map_err(|_| {
+            #[cfg(feature = "debug")]
+            eprintln!("side-table conversion overflow {delta}");
+            unsupported(if_debug!(Unsupported::SideTable))
+        })
+    }
+
+    fn persist(self) -> MResult<Vec<SideTableEntry>, Check> {
+        self.entries.into_iter().map(|entry| SideTableEntry::new(entry.unwrap())).collect()
+    }
 }
 
 #[derive(Debug, Copy, Clone)]
 struct SideTableBranch<'m> {
     parser: &'m [u8],
     side_table: usize,
-}
-
-impl SideTableBranch<'_> {
-    fn delta_ip(&self, source: &[u8]) -> i32 {
-        (self.parser.as_ptr() as isize - source.as_ptr() as isize) as i32
-    }
-
-    fn delta_stp(&self, source: usize) -> i32 {
-        (self.side_table - source) as i32
-    }
 }
 
 #[derive(Debug, Default)]
@@ -458,7 +496,7 @@ impl<'a, 'm> Expr<'a, 'm> {
             is_body: false,
             locals: vec![],
             labels: vec![Label::default()],
-            side_table: vec![],
+            side_table: SideTable::default(),
         }
     }
 
@@ -481,7 +519,7 @@ impl<'a, 'm> Expr<'a, 'm> {
         expr.locals = locals;
         expr.label().type_.results = results;
         expr.check()?;
-        expr.side_table.into_iter().map(|entry| SideTableEntry::new(entry.unwrap())).collect()
+        expr.side_table.persist()
     }
 
     fn check(&mut self) -> CheckResult {
@@ -519,15 +557,19 @@ impl<'a, 'm> Expr<'a, 'm> {
             Unreachable => self.stack_polymorphic(),
             Nop => (),
             Block(b) => self.push_label(self.blocktype(&b)?, LabelKind::Block)?,
-            Loop(b) => self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch()))?,
+            Loop(b) => {
+                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch_target()))?
+            }
             If(b) => {
                 self.pop_check(ValType::I32)?;
-                self.push_label(self.blocktype(&b)?, LabelKind::If(self.branch()))?;
-                self.side_table.push(None);
+                let branch = self.branch_source();
+                self.push_label(self.blocktype(&b)?, LabelKind::If(branch))?;
             }
             Else => {
                 match core::mem::replace(&mut self.label().kind, LabelKind::Block) {
-                    LabelKind::If(source) => self.stitch(source, self.branch()),
+                    LabelKind::If(source) => {
+                        self.side_table.stitch(source, self.branch_target())?
+                    }
                     _ => Err(invalid())?,
                 }
                 let FuncType { params, results } = self.label().type_;
@@ -772,14 +814,14 @@ impl<'a, 'm> Expr<'a, 'm> {
     }
 
     fn end_label(&mut self) -> CheckResult {
-        let target = self.branch();
+        let target = self.branch_target();
         for source in core::mem::take(&mut self.label().branches) {
-            self.stitch(source, target);
+            self.side_table.stitch(source, target)?;
         }
         let label = self.label();
         if let LabelKind::If(source) = label.kind {
             check(label.type_.params == label.type_.results)?;
-            self.stitch(source, target);
+            self.side_table.stitch(source, target)?;
         }
         let results = self.label().type_.results;
         self.pops(results)?;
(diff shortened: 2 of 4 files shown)


ia0 (commented):
There are a few issues:
- `side_table_last_entry` is wrong, you need to supported nested blocks. You might need to put an index to the side table in `Label` or `LabelKind`.
- I don't think instruction index is the correct abstraction, I think we want instruction address. So you don't need `instr_idx` at all and can use `parser.save().as_ptr() as i32`.

Then in terms of design, I think we should keep `Context` constant in `Expr` and make `Expr::check_body` return the side-table. We can then set it from `Context::check_module` where `check_body` is called.

zhouwfang:
> There are a few issues:
> 
> * `side_table_last_entry` is wrong, you need to supported nested blocks. You might need to put an index to the side table in `Label` or `LabelKind`.
> * I don't think instruction index is the correct abstraction, I think we want instruction address. So you don't need `instr_idx` at all and can use `parser.save().as_ptr() as i32`.
> 
> Then in terms of design, I think we should keep `Context` constant in `Expr` and make `Expr::check_body` return the side-table. We can then set it from `Context::check_module` where `check_body` is called.

Thanks! The issues with `parser.save().as_ptr() as i32` are

1) `parser.save().as_ptr()` overflows.

2) The [`DELTA_IP_MASK`](https://github.com/zhouwfang/wasefire/blob/ae1e52970926707134bd9651f425de9614584f21/crates/interpreter/src/side_table.rs#L36) is too small to hold it.

crates/interpreter/src/valid.rs
```diff
@@ -503,22 +507,53 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                    delta_stp: 0,
+                    val_cnt: 0,
+                    pop_cnt: 0,
+                });
+            }
+            End => {
+                if self.immutable_label().is_block_under_else {
+                    let side_table_idx = self.immutable_label().side_table_idx.unwrap() + 1;
+                    self.side_table[side_table_idx].delta_ip += self.instr_idx;
+                }
```
ia0, line 554: You can see that this code is unreachable from the macro below. The `End` instruction is handled by `end_label()`.

crates/interpreter/src/valid.rs
```diff
@@ -503,22 +507,53 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                self.side_table.push(SideTableEntryView {
+                    delta_ip: -self.instr_idx,
+                    // TODO(dev/fast-interp): implement below.
+                    delta_stp: 0,
+                    val_cnt: 0,
+                    pop_cnt: 0,
+                });
             }
             Else => {
+                let side_table_idx = self.immutable_label().side_table_idx.unwrap();
```
ia0, line 529: I'm not sure why you need `immutable_label()`. You just get something that is not borrowed out of it. Using `label()` should just work.

ia0 (commented):
I'm not sure it's worth iterating too much. It's simpler to just design correctly from the start, which means understanding the goal. Essentially, we want all occurrences of `skip_to_{else,end}()` in `exec.rs` to be resolved from the side-table (to which we can add `Loop` to handle everything uniformly). This means `If`, `Else`, `Br{,If}` should each have an entry in the side-table and `BrTable` as much as its arguments. This defines when an entry is added. Then we need to fill their content, which is done on `Loop`, `Else`, and `End` (with `Loop` done before the entry is added). This means that for `Loop` we can store the parser position in the `LabelKind` (like done in `exec.rs`). For `Else` and `End`, we need the side-table entries to register themselves in their labels. This means that `Label` should contain an `entries: Vec<(&'m [u8], usize)>` recording for each side-table entry the parser position and the side-table length when it was added. Then in `end_label()` we can simply go over all entries compute the difference between the current parser position and the saved one to compute the delta IP, as well as the difference between the current side-table length and the index for the delta STP. We can then update the side-table entry.

ia0:
> Thanks! The issues with `parser.save().as_ptr() as i32` are
> 
> 1. `parser.save().as_ptr()` overflows.
> 2. The [`DELTA_IP_MASK`](https://github.com/zhouwfang/wasefire/blob/ae1e52970926707134bd9651f425de9614584f21/crates/interpreter/src/side_table.rs#L36) is too small to hold it.

Those are not issues because we only care about deltas between `parser.save()`. In the proposed design, I store directly the `&'m [u8]` delays converting to `i32` until after we compute the delta.

zhouwfang:
> I'm not sure it's worth iterating too much. It's simpler to just design correctly from the start, which means understanding the goal. Essentially, we want all occurrences of `skip_to_{else,end}()` in `exec.rs` to be resolved from the side-table (to which we can add `Loop` to handle everything uniformly). This means `If`, `Else`, `Br{,If}` should each have an entry in the side-table and `BrTable` as much as its arguments. This defines when an entry is added. Then we need to fill their content, which is done on `Loop`, `Else`, and `End` (with `Loop` done before the entry is added). This means that for `Loop` we can store the parser position in the `LabelKind` (like done in `exec.rs`). For `Else` and `End`, we need the side-table entries to register themselves in their labels. This means that `Label` should contain an `entries: Vec<(&'m [u8], usize)>` recording for each side-table entry the parser position and the side-table length when it was added. Then in `end_label()` we can simply go over all entries compute the difference between the current parser position and the saved one to compute the delta IP, as well as the difference between the current side-table length and the index for the delta STP. We can then update the side-table entry.

Thanks! I implemented this design except `Loop` for the moment. I'll run some tests in the meanwhile.

crates/interpreter/src/valid.rs
```diff
@@ -742,15 +779,23 @@ impl<'a, 'm> Expr<'a, 'm> {
...
 
     fn end_label(&mut self) -> CheckResult {
+        for (label_parser_pos, idx) in &self.labels.last().unwrap().side_table_log_entries {
+            let parser_pos = self.parser.save().as_ptr() as i32;
+            self.side_table[*idx].delta_ip = parser_pos - label_parser_pos.as_ptr() as i32;
+        }
+
         let label = self.label();
         if label.kind == LabelKind::If {
             check(label.type_.params == label.type_.results)?;
```
zhouwfang, line 801: @ia0 Could you explain the purpose of this if block? Thanks.

crates/interpreter/src/valid.rs
```diff
@@ -742,15 +779,23 @@ impl<'a, 'm> Expr<'a, 'm> {
...
-        let label = Label { type_, kind, polymorphic: false, stack };
+        let label = Label { type_, kind, polymorphic: false, stack, side_table_log_entries };
         self.labels.push(label);
         Ok(())
     }
 
     fn end_label(&mut self) -> CheckResult {
+        for (label_parser_pos, idx) in &self.labels.last().unwrap().side_table_log_entries {
+            let parser_pos = self.parser.save().as_ptr() as i32;
+            self.side_table[*idx].delta_ip = parser_pos - label_parser_pos.as_ptr() as i32;
```
ia0, line 796: Let's do the arithmetic as `isize` just for cleanness.

ia0, line 796: Let's have a `fn stitch(&mut self, branch: &SideTableBranch)` helper for this because we also need to call it from `Else`.

crates/interpreter/src/valid.rs
```diff
@@ -742,15 +779,23 @@ impl<'a, 'm> Expr<'a, 'm> {
...
         self.pops(type_.params)?;
         let stack = type_.params.iter().cloned().map(OpdType::from).collect();
-        let label = Label { type_, kind, polymorphic: false, stack };
+        let label = Label { type_, kind, polymorphic: false, stack, side_table_log_entries };
         self.labels.push(label);
         Ok(())
     }
 
     fn end_label(&mut self) -> CheckResult {
+        for (label_parser_pos, idx) in &self.labels.last().unwrap().side_table_log_entries {
```
ia0, line 794: Let's just take the branches since we're gonna trash the label anyway.

```rust
let branches = core::mem::take(&mut self.label().branches);
for branch in branches {
```

Maybe we can inline, but I'm not sure the borrow checker will see it's fine.

crates/interpreter/src/valid.rs
```diff
@@ -775,6 +820,35 @@ impl<'a, 'm> Expr<'a, 'm> {
         })
     }
 
+    fn br_label_side_table_log_entries(
```
ia0, line 823: This could become `push_entry`. And we can call it from `br_label` and `Else`.

crates/interpreter/src/valid.rs
```diff
@@ -503,27 +502,62 @@ impl<'a, 'm> Expr<'a, 'm> {
...
                 let FuncType { params, results } = label.type_;
                 self.pops(results)?;
                 check(self.stack().is_empty())?;
                 self.label().polymorphic = false;
                 self.pushs(params);
+
+                let parser_pos = self.parser.save();
+                let side_table_len = self.side_table.len();
+                self.label().side_table_log_entries.push((parser_pos, side_table_len));
+                self.push_default_side_table_entry();
```
ia0, line 547: This could be factorized in a `push_entry(&mut self, l: LabelIdx)` function.

crates/interpreter/src/valid.rs
```diff
@@ -775,6 +820,35 @@ impl<'a, 'm> Expr<'a, 'm> {
         })
     }
 
+    fn br_label_side_table_log_entries(
+        &mut self, l: LabelIdx,
+    ) -> Result<&mut Vec<(&'m [u8], usize)>, Error> {
+        let l = l as usize;
+        let n = self.labels.len();
+        check(l < n)?;
```
ia0, line 828: We don't need to check this because it's already done by `br_label`.

crates/interpreter/src/valid.rs
```diff
@@ -775,6 +820,35 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let label = &mut self.labels[n - l - 1];
+        self.side_table.push(SideTableEntryView {
+            delta_ip: 0,
+            delta_stp: 0,
+            val_cnt: 0,
+            pop_cnt: 0,
+        });
+        Ok(&mut label.side_table_log_entries)
```
ia0, line 836: Instead of returning this, we can directly push the branch to the label.

crates/interpreter/src/valid.rs
zhouwfang: Removed the check in `br_label` to satisfy borrow checker.

crates/interpreter/src/valid.rs
```diff
@@ -503,29 +511,58 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                )?;
                 let FuncType { params, results } = label.type_;
                 self.pops(results)?;
                 check(self.stack().is_empty())?;
                 self.label().polymorphic = false;
                 self.pushs(params);
+
+                let parser = self.parser.save();
+                let side_table_len = self.side_table.len();
+                self.label().branches.push(SideTableBranch { parser, side_table_len });
```
ia0, line 553: This should just be `push_entry(0)`.

crates/interpreter/src/valid.rs
```diff
@@ -503,29 +511,58 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                self.push_entry_only();
             }
             End => unreachable!(),
             Br(l) => {
-                self.pops(self.br_label(l)?)?;
+                let res = self.br_label(l)?;
+                self.pops(res)?;
```
ia0, line 559: Interesting, is the borrow checker newly complaining about this?

crates/interpreter/src/valid.rs
```diff
@@ -742,17 +779,26 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let branches = core::mem::take(&mut self.label().branches);
+        for branch in branches {
+            self.stitch(&branch);
+        }
+
         let label = self.label();
-        if label.kind == LabelKind::If {
+        if let LabelKind::If(SideTableBranch { parser: _parser, side_table_len: _side_table_len }) =
+            label.kind
+        {
```
ia0, line 801:
```suggestion
        if matches!(label.kind, LabelKind::If(_)) {
```

crates/interpreter/src/valid.rs
```diff
@@ -412,24 +406,32 @@ struct Expr<'a, 'm> {
...
     /// Whether an `else` is possible before `end`.
-    kind: LabelKind,
+    kind: LabelKind<'m>,
     /// Whether the bottom of the stack is polymorphic.
     polymorphic: bool,
     stack: Vec<OpdType>,
+    branches: Vec<SideTableBranch<'m>>,
 }
 
 #[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
```
ia0, line 429: Actually, now that we're making this enum more than just an integer, we should remove `Copy, PartialEq, Eq`. This was just used for `If` as convenience. But we should use `match` and `matches!` now.

crates/interpreter/src/valid.rs
```diff
@@ -412,24 +406,32 @@ struct Expr<'a, 'm> {
     is_body: bool,
     locals: Vec<ValType>,
     labels: Vec<Label<'m>>,
+    side_table: Vec<Option<SideTableEntryView>>,
+}
+
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
```
ia0, line 412: See below, doesn't need to be `Copy, PartialEq, Eq`. And probably also not `Default, Clone`, actually definitely not `Default`.

crates/interpreter/src/valid.rs
ia0: What do you mean? There shouldn't be any borrow checking issue.

crates/interpreter/src/valid.rs
```diff
@@ -412,24 +406,32 @@ struct Expr<'a, 'm> {
...
+struct SideTableBranch<'m> {
+    parser: &'m [u8],
+    side_table_len: usize,
```
ia0, line 415: This is not the `len`, this is the index of the entry, which just happens to be the length before pushing it.

```suggestion
    side_table: usize,
```

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +810,60 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+            delta_ip: 0,
+            delta_stp: 0,
+            val_cnt: 0,
+            pop_cnt: 0,
+        }));
+        Ok(())
+    }
+
+    fn push_entry_only(&mut self) {
+        self.side_table.push(Some(SideTableEntryView {
```
ia0, line 855: We should never push `Some`. There also shouldn't be 2 push_entry functions. This one seems wrong, we also want to push a branch when pushing an entry, otherwise we may break the one-to-one mapping.

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +810,60 @@ impl<'a, 'm> Expr<'a, 'm> {
...
-        check(l < n)?;
         let label = &self.labels[n - l - 1];
         Ok(match label.kind {
-            LabelKind::Block | LabelKind::If => label.type_.results,
+            LabelKind::Block | LabelKind::If(_) => label.type_.results,
             LabelKind::Loop => label.type_.params,
         })
     }
 
+    fn push_entry(&mut self, l: LabelIdx) -> CheckResult {
```
ia0, line 838: Actually, we probably don't need this function, and we can call `br_label(0)` in `Else`. Then we always push branch and entry in `br_label`.

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +810,60 @@ impl<'a, 'm> Expr<'a, 'm> {
...
     }
 
-    fn br_label(&self, l: LabelIdx) -> Result<ResultType<'m>, Error> {
+    fn stitch(&mut self, branch: &SideTableBranch) {
+        let current_parser = self.parser.save().as_ptr() as isize;
+        let SideTableBranch { parser, side_table_len } = branch;
+        let entry = &mut self.side_table[*side_table_len];
+        if entry.is_none() {
+            *entry = Some(SideTableEntryView {
+                delta_ip: (current_parser - parser.as_ptr() as isize) as i32,
```
ia0, line 819: `(current.parser.as_ptr() as isize - branch.parser.as_ptr() as isize) as i32`
We could have a `SideTableBranch::delta_ip()` helper for that.

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +810,60 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                delta_stp: 0,
```
ia0, line 820: `current.side_table - branch.side_table`

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +810,60 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+
+    fn br_label(&mut self, l: LabelIdx) -> Result<ResultType<'m>, Error> {
+        self.push_entry(l)?;
         let l = l as usize;
         let n = self.labels.len();
-        check(l < n)?;
         let label = &self.labels[n - l - 1];
         Ok(match label.kind {
-            LabelKind::Block | LabelKind::If => label.type_.results,
+            LabelKind::Block | LabelKind::If(_) => label.type_.results,
```
ia0, line 833: Ideally, we should only push an entry and branch in this case. For `Loop` below, we directly want to produce the correct entry without a branch.

ia0 (commented):
Thanks! There are still some fundamental issues but the overall shape is getting there.

crates/interpreter/src/valid.rs
zhouwfang: Yes, because `br_label()` now takes `&mut self` instead.

crates/interpreter/src/valid.rs
zhouwfang: `delta_stp` can be much larger than expected, and leads to [the error here](https://github.com/zhouwfang/wasefire/blob/382e5737b2b674a2ded1e864f29f8a4941cd1f9b/crates/interpreter/src/bit_field.rs#L32). We would need to make the masks larger, and I plan to do it in another PR.

crates/interpreter/src/valid.rs
zhouwfang: `Copy` is used in `self.label().branches.push(branch)` for `If`.

crates/interpreter/src/valid.rs
```diff
@@ -503,27 +502,62 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                self.push_label(
+                    self.blocktype(&b)?,
+                    LabelKind::If,
+                    vec![self.side_table_log_entry()],
+                )?;
+                self.push_default_side_table_entry();
             }
             Else => {
+                self.side_table.last_mut().unwrap().delta_ip = self.parser.save().as_ptr() as i32
+                    - self.label().side_table_log_entries.last().unwrap().0.as_ptr() as i32;
```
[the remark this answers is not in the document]
zhouwfang: I do not store branch in `LabelKind::If` to handle the case when there is no `else` uniformly.

crates/interpreter/src/valid.rs
zhouwfang: I added the computation for `delta_stp` in this PR which requires modifications to the masks.

crates/interpreter/src/side_table.rs
```diff
@@ -33,12 +33,12 @@ pub struct SideTableEntryView {
 
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
-    const DELTA_IP_MASK: u32 = 0x0000ffff;
-    const DELTA_STP_MASK: u32 = 0x003f0000;
-    const VAL_CNT_MASK: u32 = 0x07c00000;
-    const POP_CNT_MASK: u32 = 0xf8000000;
+    const DELTA_IP_MASK: u32 = 0x7ff;
+    const DELTA_STP_MASK: u32 = 0x7ffff800; // 0x7fff << 11
```
zhouwfang, line 37: 1) These two masks have been experimented to be the smallest ones to make the test suite pass. The two `.wast` file that requires much larger `DELTA_STP_MASK` are `align.wast` and `br_table.wast` (e.g. [this function](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107)).  We may want to skip some unusual functions in tests to keep using `u32`, but it does not seem straightforward given the current setup.

2) Because of 1), the number of remaining bits for `VAL_CNT_MASK` and `POP_CNT_MASK` will likely be not enough when we compute `val_cnt` and `pop_cnt`. How about using `u64` instead if we have to keep all the test cases?

crates/interpreter/src/side_table.rs
```diff
@@ -34,11 +34,11 @@ pub struct SideTableEntryView {
 #[allow(dead_code)] // TODO(dev/fast-interp)
 impl SideTableEntry {
     const DELTA_IP_MASK: u32 = 0x0000ffff;
-    const DELTA_STP_MASK: u32 = 0x003f0000;
-    const VAL_CNT_MASK: u32 = 0x07c00000;
-    const POP_CNT_MASK: u32 = 0xf8000000;
+    const DELTA_STP_MASK: u32 = 0x7fff0000;
```
zhouwfang, line 37: 1) Based on experiments, `DELTA_IP_MASK` and `DELTA_STP_MASK` are the smallest ones to make the test suite and CI pass. The two `.wast` file that requires much larger `DELTA_STP_MASK` are `align.wast` and `br_table.wast` (e.g. [this function](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107)).  We may want to skip some unusual functions in tests to keep using `u32`, but it does not seem straightforward given the current setup.

2) Because of 1), the number of remaining bits for `VAL_CNT_MASK` and `POP_CNT_MASK` will not be enough when we compute `val_cnt` and `pop_cnt`. How about using `u64` instead if we have to keep all the test cases?

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +804,50 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let label = &mut self.labels[n - l - 1];
         Ok(match label.kind {
-            LabelKind::Block | LabelKind::If => label.type_.results,
-            LabelKind::Loop => label.type_.params,
+            LabelKind::Block => {
+                label.branches.push(branch);
+                self.side_table.push(None);
+                label.type_.results
+            }
+            LabelKind::If => label.type_.results,
```
ia0, line 833: I suspect this is wrong and the behavior should be shared with `Block`.

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +804,50 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+            LabelKind::Loop(SideTableBranch { parser, side_table }) => {
+                self.side_table.push(Some(SideTableEntryView {
+                    delta_ip: SideTableBranch::delta_ip(branch.parser, parser),
+                    delta_stp: SideTableBranch::delta_stp(branch.side_table, side_table),
+                    // TODO(dev/fast-interp): Compute the fields below.
+                    val_cnt: 0,
+                    pop_cnt: 0,
+                }));
```
ia0, line 841: It's probably cleaner to do something like this:
```suggestion
            LabelKind::Loop(target) => {
                self.stitch(source, target);
```

We need to make `stitch` take both source and target instead of using `self.branch()`. This is accidentally better, because then we can factorize the call to `self.branch()` in `fn end()`.

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +804,50 @@ impl<'a, 'm> Expr<'a, 'm> {
         Ok(())
     }
 
-    fn br_label(&self, l: LabelIdx) -> Result<ResultType<'m>, Error> {
+    fn stitch(&mut self, branch: &SideTableBranch) {
```
ia0, line 807: If `SideTableBranch` is `Copy`, we should not take it by reference.
```suggestion
    fn stitch(&mut self, branch: SideTableBranch) {
```

ia0 (commented):
Thanks, it's almost there. The main question is how to deal with pathological examples in the test suite. I think the strategy of using a sentinel value in the side table entries and returning an error at runtime is a good option. If this results in performance problems (because we need to check the sentinel at runtime), we can gate this behavior behind a feature (like `toctou`). The feature would move returning an error from `exec.rs` to `valid.rs` ensuring that there are no sentinel values at runtime.

ia0:
I've merge `main` into `dev/fast-interp`. You might want to pull and merge your `add-delta-ip` branch into `dev/fast-interp`. This adds better support for unsupported operations, such that we can still run parts of a test instead of having to ignore it completely.

crates/interpreter/src/valid.rs
```diff
@@ -503,29 +514,44 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+            }
+            Loop(b) => {
+                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.parser.save()))?;
+            }
             If(b) => {
                 self.pop_check(ValType::I32)?;
+                let branch = self.branch();
                 self.push_label(self.blocktype(&b)?, LabelKind::If)?;
+                self.side_table.push(None);
+                self.label().branches.push(branch);
```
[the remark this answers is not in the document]
zhouwfang: With `LabelKind::If`, we would also need to add line 527 (`self.side_table.push(None);`) here. But the problem seems that this line only applies to the case when there is `Else`. If there were no `else`, this line would seem redundant and wrong, as it should be added in `br_label` instead. WDYT?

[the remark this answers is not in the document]
ia0: If there is no `Else` it is implicitly the `End`, so in some sense there is always an `Else`. I'm commenting in `end_label()` to describe how to handle the case where there is no `Else`.

crates/interpreter/src/valid.rs
```diff
@@ -742,17 +778,21 @@ impl<'a, 'm> Expr<'a, 'm> {
...
     fn end_label(&mut self) -> CheckResult {
+        for branch in core::mem::take(&mut self.label().branches) {
+            self.stitch(&branch);
+        }
+
         let label = self.label();
-        if label.kind == LabelKind::If {
+        if matches!(label.kind, LabelKind::If) {
             check(label.type_.params == label.type_.results)?;
         }
```
ia0, line 797: Here, we want something like this:
```suggestion
        if LabelKind::If(source) = label.kind {
            self.stitch(source, target);
            check(label.type_.params == label.type_.results)?;
        }
```
Where `target` is defined at the beginning of the function with `let target = self.branch();` and reused by all calls to `self.stitch(source, target)`. (The `for` loop will be `for source in take(...)`.)

crates/interpreter/src/valid.rs
```diff
@@ -742,17 +778,21 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let label = Label { type_, kind, polymorphic: false, stack, branches: vec![] };
         self.labels.push(label);
         Ok(())
     }
 
     fn end_label(&mut self) -> CheckResult {
+        for branch in core::mem::take(&mut self.label().branches) {
+            self.stitch(&branch);
+        }
+
```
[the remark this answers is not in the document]
zhouwfang: Just wondering your thought behind this. I tend to use empty lines to create "paragraphs" as independent units/steps within a function.

zhouwfang:
Could you clarify the [hw-host](https://github.com/google/wasefire/actions/runs/11414751782/job/31764041510?pr=626#logs) error? I guess it results from the mask size issue.

zhouwfang:
I posted https://github.com/zhouwfang/wasefire/pull/2 to add `val_cnt` and `pop_cnt` on top of this PR. Could you also take a look at that? Thanks.

crates/interpreter/src/valid.rs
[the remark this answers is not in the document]
ia0: Yes, I also do that, but only if the function is long and shouldn't be split in smaller functions. This function is quite short, adding vertical spacing just reduces readability instead of increasing it. Said otherwise, vertical space is useful if the function doesn't fit on the screen, by adding anchors in the form of empty lines (ideally with a comment which would be the name of the helper function if the function were to be split).

crates/interpreter/src/valid.rs
```diff
@@ -504,28 +519,36 @@ impl<'a, 'm> Expr<'a, 'm> {
             Unreachable => self.stack_polymorphic(),
             Nop => (),
             Block(b) => self.push_label(self.blocktype(&b)?, LabelKind::Block)?,
-            Loop(b) => self.push_label(self.blocktype(&b)?, LabelKind::Loop)?,
+            Loop(b) => {
+                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch()))?;
+            }
```
ia0, line 524:
```suggestion
            Loop(b) => self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch()))?,
```

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +792,48 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+                label.type_.results
+            }
+            LabelKind::Loop(SideTableBranch { parser, side_table }) => {
+                self.side_table.push(Some(SideTableEntryView {
+                    delta_ip: target.delta_ip(parser),
+                    delta_stp: target.delta_stp(side_table),
+                    // TODO(dev/fast-interp): Compute the fields below.
+                    val_cnt: 0,
+                    pop_cnt: 0,
+                }));
```
ia0, line 827: I think it's better to reuse code (even if the compiler might not optimize to the same code).
```suggestion
            LabelKind::Loop(target) => {
                self.side_table.push(None);
                self.stitch(source, target);
```

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +792,48 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+            pop_cnt: 0,
+        });
+    }
+
+    fn br_label(&mut self, l: LabelIdx) -> Result<ResultType<'m>, Error> {
         let l = l as usize;
         let n = self.labels.len();
         check(l < n)?;
-        let label = &self.labels[n - l - 1];
+        let target = self.branch();
```
ia0, line 812:
```suggestion
        let source = self.branch();
```

crates/interpreter/src/valid.rs
```diff
@@ -764,17 +792,48 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        let label = &mut self.labels[n - l - 1];
         Ok(match label.kind {
-            LabelKind::Block | LabelKind::If => label.type_.results,
-            LabelKind::Loop => label.type_.params,
+            LabelKind::Block | LabelKind::If(_) => {
+                label.branches.push(target);
+                self.side_table.push(None);
```
ia0, line 817: We can actually move this `push` right after the `self.branch()` since it's called by both branches of the match. It's also an important invariant that `push(None)` is always called after `self.branch()` when producing a "source".

crates/interpreter/src/valid.rs
```diff
@@ -412,24 +406,43 @@ struct Expr<'a, 'm> {
...
+}
+
+impl SideTableBranch<'_> {
+    fn delta_ip(&self, source: &[u8]) -> i32 {
+        (self.parser.as_ptr() as isize - source.as_ptr() as isize) as i32
+    }
+
+    fn delta_stp(&self, source: usize) -> i32 {
+        (self.side_table - source) as i32
+    }
```
ia0, line 426: Actually, let's inline those 2 functions in `stitch` to incentivize using `stitch`. Ideally, all stitching should occur there.

ia0 (commented):
Thanks! I think this should work modulo off-by-one errors. While updating `exec.rs` we might want to change the exact location of the branches or how to interpret the side table.

ia0:
> Could you clarify the [hw-host error](https://github.com/google/wasefire/actions/runs/11414751782/job/31764041510?pr=626#logs)? I guess it results from the side table bit mask size issue.

Yes, this is indeed the side table masks being too short. So we'll probably have to use a `u64` for the bit fields. You can also apply the following patch:

```patch
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -29,6 +29,8 @@ fn offset(mask: u32) -> i32 {
 pub fn into_field(mask: u32, value: u32) -> Result<u32, Error> {
     let field = (value << mask.trailing_zeros()) & mask;
     if from_field(mask, field) != value {
+        eprintln!("Bit field value {value:08x} doesn't fit in mask {mask:08x}.");
         return Err(unsupported(if_debug!(Unsupported::SideTable)));
     }
     Ok(field)
```

This is similar to what we do in the `memory_too_small` function.

crates/interpreter/src/valid.rs
zhouwfang: Borrow checker error: the `label` above results from a mutable borrow, and `self.stitch()` is a second one.

crates/interpreter/src/valid.rs
zhouwfang: (I didn't do this refactor due to the borrow checker error mentioned in another comment.)

crates/interpreter/src/valid.rs
zhouwfang: (I didn't do this refactor due to the borrow checker error mentioned in another comment.)

ia0 (approved):
Thanks! I've refactored the side table out to help the borrow checker understand it's alright. I also split `branch()` into `branch_source()` and `branch_target()` to be more explicit why we're taking a branch and avoid accidents of forgetting to allocate a side-table entry when taking a source branch.

PR #635, opened 2024-10-07
Fix logging levels in scheduler

ia0:


ia0-review (approved):

ia0:
I'm surprised by the footprint impact and I'm not sure I can explain it. Even if the default logging level of defmt is `error` (which I think it is), I'm just removing 3 (or 4 with the panic) such logging calls. 32k seems too much.

PR #636, opened 2024-10-09
Move changelog.rs into cli-tools. Remove clap dependency.

exzachlyvv:


crates/cli-tools/Cargo.toml
```diff
@@ -22,6 +22,7 @@ humantime = { version = "2.1.0", default-features = false, optional = true }
 indicatif = { version = "0.17.8", default-features = false, optional = true }
 log = { version = "0.4.21", default-features = false }
 rusb = { version = "0.9.4", default-features = false, optional = true }
+semver = "1.0.23"
```
ia0, line 25: I guess the changelog script checks that we always use `default-features = false`. (Add any feature you need explicitly.)
```suggestion
semver = { version = "1.0.23", default-features = false }
```

crates/cli-tools/src/changelog.rs
```diff
@@ -16,13 +16,14 @@ use core::str;
 use std::collections::BTreeMap;
 use std::fmt::{Display, Write};
 use std::io::BufRead;
+use std::str::FromStr;
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use semver::Version;
 use tokio::process::Command;
-use wasefire_cli_tools::{cmd, fs};
+use crate::{cmd, fs};
```
ia0, line 24: I would expect formatting to insert a newline here.

crates/cli-tools/src/changelog.rs
```diff
@@ -16,13 +16,14 @@ use core::str;
...
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
ia0, line 26: Why not use `ValueEnum` anymore?

crates/cli-tools/src/lib.rs
```diff
@@ -34,6 +34,7 @@ macro_rules! debug {
 
 #[cfg(feature = "action")]
 pub mod action;
+pub mod changelog;
```
ia0, line 37: Let's also gate this under a `changelog` feature.

ia0 (commented):
Thanks! Looks good.

crates/cli-tools/src/changelog.rs
exzachlyvv: I was thinking so we didn't need to add clap as a dependency (FromString logic simple enough).

However, since we are hiding changelog module behind a feature anways, might as well include clap. I've reverted back to ValueEnum, no FromStr impl.

ia0: I see. But we already have it as an optional dependency for the `action` feature. So we can also enable it with `changelog`.

crates/cli-tools/src/changelog.rs
```diff
@@ -18,9 +18,11 @@ use std::fmt::{Display, Write};
 use std::io::BufRead;
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
+use cargo_metadata as _;
```
exzachlyvv, line 21: Added to provide Semver error impl.

We will need soon anyways to validate latest version in changelog file matches cargo.toml

crates/cli-tools/src/changelog.rs
exzachlyvv: Yep, agreed. That's what we will do now.

(Before changelog was behind a feature, I was concerned it wasn't worth always adding this dependency.)

ia0 (approved):

PR #640, opened 2024-10-11
Add validation that cargo.toml version matches most recent CHANGELOG.md version.

exzachlyvv:


crates/cli-tools/src/changelog.rs
```diff
@@ -173,6 +173,21 @@ impl Changelog {
 
         Ok(changelog)
     }
+
+    fn validate_cargo_toml(&self, path: &str) -> Result<()> {
+        let most_recent_release =
+            self.releases.first().context("At least 1 release is required")?;
```
ia0, line 179: Can this happen? We only produce valid changelogs right? So it's an invariant that `releases` is not empty.

crates/cli-tools/src/changelog.rs
```diff
@@ -228,7 +243,11 @@ pub async fn execute_ci() -> Result<()> {
         let path = path?;
 
         // Validation done during parsing.
-        Changelog::read_file(&path).await?;
+        let changelog = Changelog::read_file(&path).await?;
+
+        let crate_dir = path.strip_suffix("/CHANGELOG.md").context("Malformed file path")?;
```
ia0, line 248: Same, I think we can `unwrap()` here. This is not a user error, this is an internal bug.

ia0 (commented):
Thanks! LGTM

crates/cli-tools/src/changelog.rs
exzachlyvv: Correct, that cannot happen. I'll use `unwrap()` instead.

ia0 (approved):

PR #646, opened 2024-10-18
Implement changelog update functionality.

exzachlyvv:
2 notes:
* This PR correctly writes back to Cargo.toml file however the original format is not preserved. Any easy way to do that?
* Not yet handling updates to crates that depend on originated change. That is next.

### DIFF
--- a/crates/cli-tools/src/changelog.rs
+++ b/crates/cli-tools/src/changelog.rs
@@ -260,10 +260,6 @@ struct Release {
 }
 
 impl Release {
-    fn new(version: &str) -> Result<Self> {
-        Ok(Release { version: Version::parse(version)?, contents: BTreeMap::new() })
-    }
-
     fn push_content(&mut self, severity: Severity, content: &str) -> Result<()> {
         self.contents.entry(severity).or_default().push(content.to_string());
         Ok(())
@@ -805,7 +801,7 @@ mod tests {
 <!-- Increment to skip CHANGELOG.md test: 0 -->
 ";
 
-        let mut changelog = Changelog::parse(&changelog_str).expect("Failed to parse changelog.");
+        let mut changelog = Changelog::parse(changelog_str).expect("Failed to parse changelog.");
 
         changelog.push_description(Severity::Major, "testing no dash").unwrap();
         changelog.push_description(Severity::Major, "- testing with dash").unwrap();
@@ -835,7 +831,7 @@ mod tests {
 <!-- Increment to skip CHANGELOG.md test: 0 -->
 ";
 
-        let mut changelog = Changelog::parse(&changelog_str).expect("Failed to parse changelog.");
+        let mut changelog = Changelog::parse(changelog_str).expect("Failed to parse changelog.");
 
         let current_release = changelog.get_or_create_release_mut();
 
@@ -857,7 +853,7 @@ mod tests {
 <!-- Increment to skip CHANGELOG.md test: 0 -->
 ";
 
-        let mut changelog = Changelog::parse(&changelog_str).expect("Failed to parse changelog.");
+        let mut changelog = Changelog::parse(changelog_str).expect("Failed to parse changelog.");
 
         let current_release = changelog.get_or_create_release_mut();
 
--- a/crates/cli-tools/src/changelog.rs
+++ b/crates/cli-tools/src/changelog.rs
@@ -24,7 +24,7 @@ use std::io::BufRead;
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use clap::ValueEnum;
-use semver::Version;
+use semver::{Prerelease, Version};
 use tokio::process::Command;
 
 use crate::cargo::metadata;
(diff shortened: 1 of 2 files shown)


crates/cli-tools/src/changelog.rs
```diff
@@ -212,6 +261,20 @@ impl Release {
...
+
+        entry.push(content.to_string());
+
+        Ok(())
+    }
+}
+
+impl From<Version> for Release {
+    fn from(value: Version) -> Self {
+        Release { version: value, contents: BTreeMap::new() }
```
ia0, line 276: nit
```suggestion
    fn from(version: Version) -> Self {
        Release { version, contents: BTreeMap::new() }
```

crates/cli-tools/src/changelog.rs
```diff
@@ -253,12 +316,19 @@ pub async fn execute_ci() -> Result<()> {
 }
 
 /// Updates a CHANGELOG.md file and CHANGELOG.md files of dependencies.
-pub async fn execute_change(path: &str, _severity: &Severity, _description: &str) -> Result<()> {
-    ensure!(fs::exists(path).await, "Crate does not exist: {path}");
+pub async fn execute_change(path: &str, severity: &Severity, description: &str) -> Result<()> {
+    let changelog_file_path = format!("{path}/CHANGELOG.md");
+
+    ensure!(fs::exists(&changelog_file_path).await, "Crate does not exist: {changelog_file_path}");
```
ia0, line 322: This is actually redundant. The `read_file` function below will fail if the file does not exist. And the `changelog_file_path` can actually be inlined and removed.

crates/cli-tools/src/changelog.rs
```diff
@@ -212,6 +261,20 @@ impl Release {
     fn new(version: &str) -> Result<Self> {
         Ok(Release { version: Version::parse(version)?, contents: BTreeMap::new() })
     }
+
+    fn push_content(&mut self, severity: Severity, content: &str) -> Result<()> {
+        let entry = self.contents.entry(severity).or_default();
+
+        entry.push(content.to_string());
+
+        Ok(())
```
ia0, line 270: nit: no need for so much vertical space

```suggestion
        self.contents.entry(severity).or_default().push(content.to_string());
        Ok(())
```

ia0 (commented):
Thanks! In parallel I'm looking to merge the existing code in `main` when I get the time.

crates/cli-tools/src/changelog.rs
```diff
@@ -58,6 +58,37 @@ impl Changelog {
         Self::parse(&String::from_utf8(fs::read(path).await?)?)
     }
 
+    async fn write_file(&self, path: &str) -> Result<()> {
+        fs::write(path, self.to_string().as_bytes()).await
+    }
+
+    fn push_description(&mut self, severity: &Severity, content: &str) -> Result<()> {
```
ia0, line 65: `Severity` is `Copy`, you can take it by value
```suggestion
    fn push_description(&mut self, severity: Severity, content: &str) -> Result<()> {
```

crates/cli-tools/src/changelog.rs
exzachlyvv: Remove the `ensure!` check, however keeping the changelog_file_path variable since it is used twice.

Something I've been thinking about is storing crate path as internal state to Changelog struct. Then don't need to pass it around.

ia0: Yes, it's probably a good idea. And in that case we can inline `parse` in `read_file`, and rename `read_file` and `write_file` to `read` and `write`.

crates/cli-tools/src/changelog.rs
```diff
@@ -187,6 +218,24 @@ impl Changelog {
...
     }
+
+    async fn sync_cargo_toml(&self, path: &str) -> Result<()> {
+        let expected_version = &self.releases.first().unwrap().version;
+        let cargo_toml_path = format!("{path}/Cargo.toml");
+        let mut cargo_toml: toml::Table = fs::read_toml(&cargo_toml_path).await?;
+
+        cargo_toml["package"]["version"] = expected_version.to_string().into();
+
+        fs::write_toml(&cargo_toml_path, &cargo_toml).await?;
```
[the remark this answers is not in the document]
exzachlyvv: Updated to use sed for modification, works well for now.

crates/cli-tools/src/changelog.rs
exzachlyvv: Sounds good, I'll do that in a follow up PR right after this one. I prefer to keep the PR's at atomic as possible to make review easier. Not mixing old with new.

crates/cli-tools/src/changelog.rs
```diff
@@ -212,6 +261,20 @@ impl Release {
     fn new(version: &str) -> Result<Self> {
         Ok(Release { version: Version::parse(version)?, contents: BTreeMap::new() })
     }
+
+    fn push_content(&mut self, severity: Severity, content: &str) -> Result<()> {
```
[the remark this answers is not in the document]
exzachlyvv: Still working on this, not ready for review yet.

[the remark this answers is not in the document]
exzachlyvv: As I think about this a little more, are we sure we want to automate that part? The best spot for this logic is in `get_or_create_release_mut`. In there, we have logic that increments patch number when the current version is not prerelease (`-git`).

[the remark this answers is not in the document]
exzachlyvv: If yes, I suggest we do as follow up PR.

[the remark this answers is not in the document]
ia0: Yes, I think we should automate that part. We could indeed make `get_or_create_release_mut` (we should probably find a better name) take a severity and update the version appropriately. The thing is that both `get_or_create_release_mut` and `push_content` break the invariant of the changelog, so ideally both should do it. So either we use a helper function or we merge `get_or_create_release_mut` and `push_content` inside `push_description` such that we don't introduce complex broken invariants across multiple functions. That's probably the best.

Let's merge like that, we can indeed always address later.

ia0 (approved):

PR #649, opened 2024-10-23
Add val_cnt and pop_cnt in side table

zhouwfang:
There was an earlier review in https://github.com/zhouwfang/wasefire/pull/2.

#46


crates/interpreter/src/valid.rs
```diff
@@ -558,18 +560,22 @@ impl<'a, 'm> Expr<'a, 'm> {
             Nop => (),
             Block(b) => self.push_label(self.blocktype(&b)?, LabelKind::Block)?,
             Loop(b) => {
-                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch_target()))?
+                let branch = self.branch_target();
+                self.push_label(self.blocktype(&b)?, LabelKind::Loop(branch))?
```
zhouwfang, line 564: What is the rust version used by the CI? I use `rustc 1.82.0-nightly (28a58f2fa 2024-07-31)`, and there is a borrow checker error for me.

crates/interpreter/src/valid.rs
```diff
@@ -424,12 +424,12 @@ impl SideTable {
...
-    fn stitch(&mut self, source: SideTableBranch, target: SideTableBranch) -> CheckResult {
+    fn stitch(
+        &mut self, source: SideTableBranch, target: SideTableBranch, val_cnt: u32,
+    ) -> CheckResult {
         let delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
         let delta_stp = Self::delta(source, target, |x| x.side_table as isize)?;
-        // TODO(dev/fast-interp): Compute the fields below.
-        let val_cnt = 0;
-        let pop_cnt = 0;
+        let pop_cnt = Self::delta(source, target, |x| x.stack as isize)? as u32;
```
zhouwfang, line 432: Do we need a `delta()` to return `u32` instead?

zhouwfang:
The [hw-host error](https://github.com/google/wasefire/actions/runs/11473675918/job/31928287988?pr=649) says "Internal:NotImplemented". What does it refer to?

crates/interpreter/src/valid.rs
```diff
@@ -856,8 +883,12 @@ impl<'a, 'm> Expr<'a, 'm> {
         branch
```
ia0, line 883: We probably need to add `self.stack().len()` to `branch.stack`, in addition of setting `val_cnt`.

ia0:
> The [hw-host error](https://github.com/google/wasefire/actions/runs/11473675918/job/31928287988?pr=649) says `Internal:NotImplemented`. What does it refer to?

Ideally we should `#[cfg(feature = "debug")] eprintln!("");` before returning `Unsupported` such that we can easily debug with `runner host --features=wasefire-interpreter/debug`.

crates/interpreter/src/valid.rs
```diff
@@ -558,18 +560,22 @@ impl<'a, 'm> Expr<'a, 'm> {
...
             Else => {
+                let branch = self.branch_target();
                 match core::mem::replace(&mut self.label().kind, LabelKind::Block) {
-                    LabelKind::If(source) => {
-                        self.side_table.stitch(source, self.branch_target())?
-                    }
+                    LabelKind::If(source) => self.side_table.stitch(
+                        source,
+                        branch,
+                        self.labels.last().unwrap().type_.results.len() as u32,
```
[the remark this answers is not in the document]
zhouwfang: My understanding of the reason why they are zero is that jumping from `if` to the beginning of `else` is a "pure jump", and no code is executed other than the condition of `if`. Is this correct/complete?

crates/interpreter/src/valid.rs
```diff
@@ -745,14 +751,22 @@ impl<'a, 'm> Expr<'a, 'm> {
 
     fn push(&mut self, t: OpdType) {
         self.stack().push(t);
+        self.label().labels_values_cnt_so_far += 1;
```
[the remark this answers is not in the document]
zhouwfang: Just wanted to understand it better. You mentioned `Label::prev_stack` should only be updated in `push_label()`. 

- `push_label()` is only called in `Loop`, `If` and `Else` in `instr()`. 

- `push()`, `pushs()` and `pop()` are called in other instructions in `instr()`.

As the validation is linear, many of the other instructions should be handled after `Loop`, `If` and `Else` are handled in `instr()`. If so, why is it correct to only update in `push_label()`?

crates/interpreter/src/valid.rs
zhouwfang: `prev_stack` has been updated so in `push_label`. Why do we need to do this again to `branch.stack`?

ia0 (commented):
Ok, I think it should be mostly correct. It'll be easier to debug once we can run the test suite using the side table.

crates/interpreter/src/valid.rs
```diff
@@ -453,6 +456,21 @@ impl SideTable {
         })
     }
 
+    fn pop_cnt(source: SideTableBranch, target: SideTableBranch) -> MResult<u32, Check> {
+        let source = source.stack as i32;
+        let target = target.stack as i32;
+        let Some(delta) = target.checked_sub(source) else {
```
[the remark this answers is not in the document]
zhouwfang: 1) Could you explain it?  After I switched `source` and `target` here, I got an [error](https://github.com/google/wasefire/actions/runs/11511885439/job/32046041805).

2) To debug locally (I added some debug prints), I got Screenshot 2024-10-24 at 10 24 32 PM in my Docker container. The tricky part is that I always get the same error (even after I use the version that passes the CI), so there seems something persistent in my container.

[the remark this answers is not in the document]
zhouwfang: 1) Let me know if my understand it correctly: `target` refers to the label that appears earlier than the label for `source`, so the value stack has more elements at the label for `source`. 

2) After I switched `source` and `target` here, I got an [error](https://github.com/google/wasefire/actions/runs/11511885439/job/32046041805). To debug locally (I added some debug prints), I got Screenshot 2024-10-24 at 10 24 32 PM in my Docker container. The tricky part is that I always get the same error (even after I use the version that passes the CI), so there seems something persistent in my container. Not sure what went wrong elsewhere.

ia0 (commented):
I tried to look into the tests and actually there are fundamental problems:
- Since I added auto-skipping of unsupported behaviors, we may accidentally pass the test because everything is unsupported due to a bug. I need to fix the tests so that we can distinguish between really unsupported and unsupported by accident. Not completely sure yet how, probably just a counter of skipped tests due to unsupported, but I'd like to find something better.
- We need to be very careful when we call `branch_{source,target}()` because it will save the stack. This matters particularly when we push a label, because we don't want to save the parameters of the label.

Sadly I don't have time at the moment to dig into this. I think the design is not there yet (for example, maybe we actually want to store the "val_cnt/result" in the target branch instead. Maybe it will help with capturing too much of the stack because we know `val_cnt` is included.

If you want to continue working on this until I get time to look into again, you can try to experiment on your own and run the test suite to test the invariants you come up with. The more asserts you can write that don't fail in the test suite, the better your understanding of the code will be.

crates/interpreter/src/valid.rs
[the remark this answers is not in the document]
zhouwfang: Same error with that removed. The output of `rustc -vV` is
```
rustc 1.82.0-nightly (28a58f2fa 2024-07-31)
binary: rustc
commit-hash: 28a58f2fa7f0c46b8fab8237c02471a915924fe5
commit-date: 2024-07-31
host: x86_64-unknown-linux-gnu
release: 1.82.0-nightly
LLVM version: 19.1.0
```

crates/interpreter/src/valid.rs
zhouwfang: You suggested adding `branch.stack += self.stack().len();` in `branch_source()`.

`branch_source()` calls `branch_target()` which set `SideTableBranch::stack` to be `self.immutable_label().prev_stack`. In `push_label()`, `prev_stack` is `prev_label.prev_stack + prev_label.stack.len()`, which seems the same as your suggestion here. Why do we need to do it again?

crates/interpreter/src/valid.rs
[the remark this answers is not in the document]
ia0: Ok so this is not because it's the wrong triple target.

Actually reading the error again it seems obvious. You can't create UNIX sockets in your container.

There is no simple way to confirm (because we don't support dynamic board configuration yet). What you could do to confirm is run `sed -i '/uart/d' Cargo.toml src/board.rs src/main.rs` from `crates/runner-host` and retry. This will disable UART support.

crates/interpreter/src/valid.rs
ia0: I see. It's not the same stack because it's not the same label.

In `push_label()` we record `prev_stack` as the stack size up to (but excluding) the label we're pushing. This is equivalent to the stack size up to (but excluding) the current label (before we push) plus the stack size of that current label, hence: `prev_stack = prev_label.prev_stack + prev_label.stack.len()`.

In `branch_source()` we record `stack` as the stack size at the current program point. This is equivalent to the stack size up to (but excluding) the current label (which we get from `branch_target()`) plus the stack size of the current label, hence: `branch.stack += self.stack().len()`.

crates/interpreter/src/valid.rs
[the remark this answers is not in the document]
zhouwfang: Thanks. The error 2) is indeed gone. But in the following new error, I wonder if you could clarify where the `crates/web-client` directory comes from:
Screenshot 2024-10-25 at 11 51 56 AM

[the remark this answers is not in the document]
zhouwfang: Thanks. The error in 2) is indeed gone. But for the following error, I wonder how to get `web.tar.gz`:
Screenshot 2024-10-25 at 11 51 56 AM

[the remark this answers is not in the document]
ia0: For that, you just need to run `make` from `crates/runner-host/crates/web-client` or `./test.sh` from `crates/runner-host`.

zhouwfang:
With the commit `cc29418` and #658, all the tests pass now.

In the meanwhile, I'm printing the side tables for some test cases in the test suite as a sanity check.

crates/interpreter/src/valid.rs
```diff
@@ -427,9 +427,12 @@ impl SideTable {
     fn stitch(&mut self, source: SideTableBranch, target: SideTableBranch) -> CheckResult {
         let delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
         let delta_stp = Self::delta(source, target, |x| x.side_table as isize)?;
-        // TODO(dev/fast-interp): Compute the fields below.
-        let val_cnt = 0;
-        let pop_cnt = 0;
+        let val_cnt = u32::try_from(source.result).map_err(|_| {
+            #[cfg(feature = "debug")]
+            eprintln!("side-table conversion overflow {0}", source.result);
```
ia0, line 432: nit (we don't get a line number so this helps distinguish error messages)
```suggestion
            eprintln!("side-table val_cnt overflow {}", source.result);
```

crates/interpreter/src/valid.rs
```diff
@@ -427,9 +427,12 @@ impl SideTable {
     fn stitch(&mut self, source: SideTableBranch, target: SideTableBranch) -> CheckResult {
         let delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
         let delta_stp = Self::delta(source, target, |x| x.side_table as isize)?;
-        // TODO(dev/fast-interp): Compute the fields below.
-        let val_cnt = 0;
-        let pop_cnt = 0;
+        let val_cnt = u32::try_from(source.result).map_err(|_| {
```
ia0, line 430: I think it's actually better to have `val_cnt` stored in the branch target instead, since that's the one defining it. This should simplify `br_label()`.

crates/interpreter/src/valid.rs
```diff
@@ -453,6 +456,21 @@ impl SideTable {
         })
     }
 
+    fn pop_cnt(source: SideTableBranch, target: SideTableBranch) -> MResult<u32, Check> {
+        let source = source.stack;
+        let target = target.stack;
+        let Some(delta) = source.checked_sub(target) else {
+            #[cfg(feature = "debug")]
+            eprintln!("side-table negative stack delta {source} < {target}");
```
ia0, line 464: I know I suggested the current version, but when I debugged last time, I think it's clearer to just output the operation, than the unexpected true property.
```suggestion
            eprintln!("side-table negative stack delta {source} - {target}");
```

crates/interpreter/src/valid.rs
```diff
@@ -453,6 +456,21 @@ impl SideTable {
...
+            return Err(unsupported(if_debug!(Unsupported::SideTable)));
+        };
+        u32::try_from(delta).map_err(|_| {
+            #[cfg(feature = "debug")]
+            eprintln!("side-table conversion overflow {delta}");
```
ia0, line 469: same here, we could write that it's for pop_cnt

ia0 (commented):
I don't mind merging this without thorough review such that you can start using the side table in `exec.rs` which is the important part to be able to debug. I think this PR needs a redesign and it's hard to tell unless the side table is actually used during execution. I'll do a proper review once we have everything working end-to-end.

But you can also use the side table in `exec.rs` in this PR if you want. As you prefer.

zhouwfang:
Thanks for the review. I plan to use `val_cnt` and `pop_cnt` in `exec.rs` in this PR.

zhouwfang:
@ia0 Could you merge `main` to `dev/fast-interp` again? Thanks.

zhouwfang:
Note: so far the side tables are correct for the examples we discussed.

crates/interpreter/src/valid.rs
```diff
@@ -560,7 +585,8 @@ impl<'a, 'm> Expr<'a, 'm> {
             Nop => (),
             Block(b) => self.push_label(self.blocktype(&b)?, LabelKind::Block)?,
             Loop(b) => {
-                self.push_label(self.blocktype(&b)?, LabelKind::Loop(self.branch_target()))?
+                let type_ = self.blocktype(&b)?;
+                self.push_label(type_, LabelKind::Loop(self.branch_target(type_.results.len())))?
```
zhouwfang, line 589: For loop, I'm not sure whether I should use `type_.results.len()` (as in other cases) or `type_.params.len()` for `val_cnt`. But I suspect the latter is correct for the following reason: in `exec.rs`, [the last `arity` values are kept](https://github.com/zhouwfang/wasefire/blob/de6c7e889865736cb1f11877d5d4e1e9f50729a8/crates/interpreter/src/exec.rs#L1058-L1059) which should be the same as `val_cnt`. By [the definition of `arity`](https://github.com/zhouwfang/wasefire/blob/de6c7e889865736cb1f11877d5d4e1e9f50729a8/crates/interpreter/src/exec.rs#L1040-L1043) in `exec.rs`, the latter is used.

zhouwfang:
After thinking more, it may be better to merge this PR first because using `val_cnt` and `pop_cnt` in `exec.rs` requires using `delta_ip` and `delta_stp` as a first step. I'd like to have a separate PR for using `delta_ip` and `delta_stp` in `exec.rs`. WDYT?

ia0 (approved):
Sounds good, let's merge this.

crates/interpreter/src/valid.rs
zhouwfang: Let me know if my understanding about the reason why it is different for loop is correct: For loop, we jump to the beginning instead of the end. When we jump to the beginning, we need to look at the params. The result of the loop will be handled by the outer scope of the loop.

ia0: Yes

PR #657, opened 2024-10-25
debug

zhouwfang:


ia0:
Why do you need to create those PRs? Is it to run the CI? You should be able to run `./test.sh` from any directory (`crates/interpreter` in your case) and `./scripts/ci.sh` from the root directory (but probably not needed for you). Otherwise, it should also be possible to do exactly the same from your fork. You can see the workflow runs here: https://github.com/zhouwfang/wasefire/actions/workflows/ci.yml

zhouwfang:
> Why do you need to create those PRs? Is it to run the CI? You should be able to run `./test.sh` from any directory (`crates/interpreter` in your case) and `./scripts/ci.sh` from the root directory (but probably not needed for you). Otherwise, it should also be possible to do exactly the same from your fork. You can see the workflow runs here: https://github.com/zhouwfang/wasefire/actions/workflows/ci.yml

Due to 2) in this comment in the other PR: 
Screenshot 2024-10-24 at 11 50 45 PM

ia0:
I see, so you can do that in your fork. No need to create a PR here. Every push in your fork will run the CI. You just need to find it in the link I gave. It will very often by the most recent one, but you can additionally filter by branch.

For running locally, I'll answer in the other comment.

zhouwfang:
Sorry for the noise.

PR #658, opened 2024-10-25
Add a skip counter for unsupported and check it

ia0:
During development it can happen that a bug returns unsupported instead of failing. We want to detect those cases. Using a skip counter is an approximation. But since for almost all tests it is zero, then it's actually exact.


ia0-review (approved):

zhouwfang:
#46

PR #669, opened 2024-10-31
Zvv/changelog internalize path

exzachlyvv:
Is there a good way to point a PR at another PR for diff?

I only want to send commit 80b4d5a7d8f9b7dc1d0622280e8a367157d8a7f2 for review, but after the other commits are merged.

### DIFF
--- a/crates/cli-tools/src/changelog.rs
+++ b/crates/cli-tools/src/changelog.rs
@@ -49,17 +49,23 @@ impl Display for Severity {
 
 #[derive(Debug, Default, PartialEq, Eq)]
 struct Changelog {
+    crate_path: String,
     releases: Vec<Release>,
     skip_counter: u32,
 }
 
 impl Changelog {
-    async fn read_file(path: &str) -> Result<Changelog> {
-        Self::parse(&String::from_utf8(fs::read(path).await?)?)
+    fn changelog_path(&self) -> String {
+        format!("{}/CHANGELOG.md", self.crate_path)
     }
 
-    async fn write_file(&self, path: &str) -> Result<()> {
-        fs::write(path, self.to_string().as_bytes()).await
+    async fn read(path: &str) -> Result<Changelog> {
+        let changelog_path = format!("{path}/CHANGELOG.md");
+        Self::parse(path, &String::from_utf8(fs::read(changelog_path).await?)?)
+    }
+
+    async fn write(&self) -> Result<()> {
+        fs::write(self.changelog_path(), self.to_string().as_bytes()).await
     }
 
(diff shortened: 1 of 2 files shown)


ia0:
> Is there a good way to point a PR at another PR for diff?

You would essentially need to create the PR in your fork, and use the branch of your other PR as base branch.

crates/cli-tools/src/changelog.rs
```diff
@@ -49,13 +49,51 @@ impl Display for Severity {
 
 #[derive(Debug, Default, PartialEq, Eq)]
 struct Changelog {
+    crate_path: String,
```
ia0, line 52: We should start using `PathBuf` and `impl AsRef<Path>` like `crates/cli/src/main.rs` and `crates/cli-tools/src/actions.rs`. This can be done in a separate PR though.

crates/cli-tools/src/changelog.rs
```diff
@@ -91,7 +97,7 @@ impl Changelog {
     }
 
     /// Parses and validates a changelog.
-    fn parse(input: &str) -> Result<Changelog> {
+    fn parse(crate_path: &str, input: &str) -> Result<Changelog> {
```
ia0, line 100: nit: Since we need ownership, we can take `String` directly.
```suggestion
    fn parse(crate_path: String, input: &str) -> Result<Changelog> {
```

If we believe it's not convenient to call it anymore, we can take `impl Into<String>` instead, and do `let crate_path = crate_path.into()` at the beginning of the function.

crates/cli-tools/src/changelog.rs
```diff
@@ -201,7 +208,7 @@ impl Changelog {
 
 impl Display for Changelog {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let Changelog { releases, skip_counter } = self;
+        let Changelog { crate_path: _, releases, skip_counter } = self;
```
ia0, line 211: nit
```suggestion
        let Changelog { releases, skip_counter, .. } = self;
```

ia0 (commented):
Thanks, LGTM

crates/cli-tools/src/changelog.rs
exzachlyvv: Will do in separate PR

crates/cli-tools/src/changelog.rs
exzachlyvv: Using `impl Into<String>` for now

ia0 (approved):

PR #672, opened 2024-11-07
Take actions by value

ia0:


ia0-review (approved):
What's the motivation behind this change?

ia0:
> What's the motivation behind this change?

We want all actions to behave similarly regarding `self` (either by shared reference or by value) since we want to compose them (e.g. `RustAppletInstall` depends on `RustAppletBuild` and `AppletInstall` and possibly `AppletExitStatus`). Doing this by value is much simpler and most existing actions already do so. This is fixing historical actions.

PR #679, opened 2024-11-08
Changelog tool - Implement robust version incrementing logic

exzachlyvv:


crates/cli-tools/src/changelog.rs
```diff
@@ -68,32 +68,37 @@ impl Changelog {
...
-        // Current version is released, insert a new one.
-        if current_release.version.pre.is_empty() {
-            let mut next_version = Version::new(
-                current_release.version.major,
-                current_release.version.minor,
-                current_release.version.patch + 1,
-            );
+        let current_release = self.get_current_release();
+        let is_new_release_needed =
+            current_release.version.pre.is_empty() || current_release.get_max_severity() > severity;
```
ia0, line 84: This doesn't sound right to me, I would expect only the `pre.is_empty()` check to be needed. In the other case, we don't want a new release but just to call `increment_version()`.

crates/cli-tools/src/changelog.rs
```diff
@@ -68,32 +68,37 @@ impl Changelog {
         fs::write(self.changelog_path(), self.to_string().as_bytes()).await
     }
 
-    fn push_description(&mut self, severity: Severity, content: &str) -> Result<()> {
-        let content = if content.starts_with("- ") { content } else { &format!("- {content}") };
-        let current_release = self.get_or_create_release_mut();
-        current_release.push_content(severity, content)
+    fn get_current_release(&self) -> &Release {
```
ia0, line 71: nit: this name is a bit too long, it could be `current_release` or even `cur_release`

crates/cli-tools/src/changelog.rs
```diff
@@ -266,19 +271,56 @@ struct Release {
...
 
+impl Release {
+    fn get_max_severity(&self) -> Severity {
+        for severity in [Severity::Major, Severity::Minor, Severity::Patch] {
+            if self.contents.contains_key(&severity) {
+                return severity;
+            }
+        }
+
+        Severity::Patch
```
ia0, line 288: 1. Keys are sorted in a BTreeMap.
2. We have an invariant that contents is not empty.
```suggestion
        self.contents.keys().next().unwrap()
```

crates/cli-tools/src/changelog.rs
```diff
@@ -266,19 +271,56 @@ struct Release {
...
+    }
+
+    fn increment_version(&mut self, severity: Severity) {
```
ia0, line 291: I don't think this function will be useful once the test above is fixed. We probably want 2 functions:
- One that creates a pre-release from a release (just incrementing patch and adding `-git`).
- One that makes sure a pre-release is at least a given severity.

That last function could really just be:
- If version.major == 0 then reduce severity by one (major to minor, minor to patch, patch to patch).
- If severity is patch then do nothing and exit.
- If version.patch > 0 then increment version.minor and reset version.patch.
- If severity is minor then do nothing and exit.
- If version.minor > 0 then increment version.major and reset version.minor.
- Do nothing and exit (we know severity is major).

crates/cli-tools/src/changelog.rs
```diff
@@ -872,10 +914,146 @@ mod tests {
...
+
+<!-- Increment to skip CHANGELOG.md test: 0 -->
+";
+
+        let mut changelog =
+            Changelog::parse("", changelog_str).expect("Failed to parse changelog.");
+
+        changelog.push_description(Severity::Major, "asdf").expect("Failed to push description.");
+
+        assert_eq!(changelog.get_current_release().version, Version::parse("0.2.0-git").unwrap());
```
ia0, line 942: We broke the invariant that we have only one pre-release per changelog. It's not enough to check the latest release.

crates/cli-tools/src/changelog.rs
exzachlyvv: Ahh nice!

crates/cli-tools/src/changelog.rs
exzachlyvv: I'm struggling to understand at a higher level. Could you explain how you see those 2 functions being used?

I think some examples of when we want to create a new release vs just increment version would be helpful.

PR #688, opened 2024-11-14
Merge main into dev/fast-interp

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/bit_field.rs
+++ b/crates/interpreter/src/bit_field.rs
@@ -1,0 +1,89 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


ia0 (approved):
clean merge

ia0:
I pushed the merge manually since the github is configured for squashing which would destroy history. Closing as done.

zhouwfang:
By "manually", do you mean without a PR? Just wondering what/why history would be destroyed.

ia0:
Yes, by pushing to the branch. I configured the repository to have linear history and using one commit per PR, so merge commits are impossible to do with a PR. I have to bypass the rules to push a merge commit, but this is fine for development branches. The whole development branch will eventually end up as a single commit in the main branch (as a PR).

The reason I want a merge commit, is that otherwise the history is rewritten, which means the next merge will be much more complicated and have a lot of conflicts. The correct way to use git is to merge, not squash. But linear history and squashing is much easier to understand. Since I'm not planning to have more than one long-term branch (namely `main`), using linear history and squashing has more benefits than issues.

PR #690, opened 2024-11-18
Apply delta_ip and delta_stp from side table

zhouwfang:
Apply `delta_ip` and `delta_stp` from side table in `exec.rs`.

#46


crates/interpreter/src/parser.rs
```diff
@@ -64,6 +64,10 @@ impl<'m, M: Mode> Parser<'m, M> {
         Ok(result)
     }
 
+    pub fn parse_bytes_reversely(&mut self, delta: i32) -> MResult<&'m [u8], M> {
+        unimplemented!()
```
zhouwfang, line 68: The difficulty is that if `delta` is negative, a similar implementation as the `parse_bytes(&mut self, len: usize)` right above would not work because we don't seem to have access to the part of the `self.data` before the current IP. So we may have to save a copy of the entire `self.data` at the beginning, in order to apply negative `delta_ip`. Is that correct?

ia0: You shouldn't need such a function. The whole point of the side-table is that we don't need to parse bytes anymore, we can directly jump where needed. I just checked in Miri with strict provenance and it seems it's not an issue to enlarge a slice (if it came from one that used to be as large). However you'll need to use unsafe:

```rust
fn jump(cur: &'m [u8], off: isize) -> &'m [u8] {
    debug_assert!(off <= cur.len() as isize);
    unsafe { core::slice::from_raw_parts(cur.as_ptr().offset(off), (cur.len() as isize - off) as usize) }
}
```

With the `toctou` feature, we could indeed save a copy of the parser at the function entry and use that to check that `0 <= (entry.len() - cur.len()) as isize + off`.

zhouwfang:
The entire test suite passes, but when `CI/hw-host` runs `target/wasefire/applet.wasm`, it points out a bug, which (the WASM module) is unfortunately too big to debug. Maybe you could find the bug in review. I'll try to debug in the meanwhile.

crates/interpreter/src/exec.rs
```diff
@@ -773,28 +773,47 @@ impl<'m> Thread<'m> {
...
-                return Ok(self.pop_label(inst, ls.get(i).cloned().unwrap_or(ln)));
+                let ls_idx = i;
+                let mut is_last = false;
+                if i >= ls.len() {
+                    is_last = true;
+                }
+                return Ok(self.pop_label(
+                    inst,
+                    ls.get(i).cloned().unwrap_or(ln),
+                    Option::from((ls_idx, is_last)),
```
ia0, line 815: A bit hacky but saves the 5 lines above. (The compiler should factorize the common sub-expression.)
```suggestion
                    ls.get(i).map(|_| i),
```

ia0 (commented):
Thanks! I didn't review everything in details but I think the global idea is there.

zhouwfang:
Forgot to ask: How can I enable `debug` feature when running [hwci.sh](https://github.com/zhouwfang/wasefire/blob/909ec3c10207e2b4dc2816e41225d6aeafc2e291/scripts/hwci.sh)? I'd like to print logs for the `applet.wasm` error.

ia0:
> Forgot to ask: How can I enable `debug` feature when running [hwci.sh](https://github.com/zhouwfang/wasefire/blob/909ec3c10207e2b4dc2816e41225d6aeafc2e291/scripts/hwci.sh)?

You can apply the following patch:

```diff
diff --git a/scripts/hwci.sh b/scripts/hwci.sh
index 4c222fd0..1f4ac986 100755
--- a/scripts/hwci.sh
+++ b/scripts/hwci.sh
@@ -77,7 +77,7 @@ full() {
 }
 
 case $1 in
-  host) full unix host --arg=--protocol=unix ;;
+  host) full unix host --arg=--protocol=unix --features=wasefire-interpreter/debug ;;
   # P1.01, P1.02, and P1.03 must be connected together (gpio_test).
   nordic) full usb nordic ;;
   *) run --protocol=${1:-usb} "$2" ;;
```

And then simply run `./scripts/hwci.sh host` as usual.

crates/interpreter/src/module.rs
```diff
@@ -222,27 +214,45 @@ impl<'m> Module<'m> {
...
-                let delta = saved.len() - parser.save().len();
-                self.cache.put(key, CacheValue { skip: delta });
+    pub(crate) fn jump_from_if(&mut self, parser: &mut Parser<'m>) {
+        let (i, j) = self.side_table_indices;
+        let entry = self.side_tables[i][j].view();
+        let delta = entry.delta_ip as isize - 1;
+        unsafe {
+            parser.restore(Self::jump(parser.save(), delta));
+            let saved = parser.save();
+            if parser.parse_instr().unwrap() == Instr::End {
```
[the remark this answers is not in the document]
zhouwfang: Could you explain more? An example would be appreciated.

I basically followed the treatment here: https://github.com/google/wasefire/blob/main/crates/interpreter/src/parser.rs#L566-L569

crates/interpreter/src/module.rs
```diff
@@ -199,12 +188,15 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
-    pub(crate) fn func(&self, x: FuncIdx) -> Parser<'m> {
+    pub(crate) fn func(&mut self, x: FuncIdx) -> Parser<'m> {
```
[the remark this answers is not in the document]
zhouwfang: Could you explain why `Module` has to be immutable? Because it is user-facing?

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
zhouwfang: Your observation looks very relevant, and I'd really like to understand it better.

The error comes from [this function](https://github.com/zhouwfang/wasefire/blob/54b17f7c291cd48b576e74defb668bab7abf12f6/target/wasefire/applet.wast#L16101), and its side table is
```
2024-12-02T21:57:57.5343889Z i=53
2024-12-02T21:57:57.5343957Z j=0
2024-12-02T21:57:57.5344023Z delta_ip=14
2024-12-02T21:57:57.5344094Z delta_stp=1
2024-12-02T21:57:57.5344155Z val_cnt=0
2024-12-02T21:57:57.5344219Z pop_cnt=0
2024-12-02T21:57:57.5344285Z j=1
2024-12-02T21:57:57.5344353Z delta_ip=8
2024-12-02T21:57:57.5344422Z delta_stp=1
2024-12-02T21:57:57.5344484Z val_cnt=0
2024-12-02T21:57:57.5344547Z pop_cnt=0
```

The error is `panicked at /home/runner/work/wasefire/wasefire/crates/interpreter/src/module.rs:243:40:
index out of bounds: the len is 2 but the index is 2`

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
ia0: That's by the WASM specification. What can be mutated is the `Instance`. The reason of this design is that the same module can be instantiated multiple times.

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
ia0: The side table looks consistent, there's indeed 13 bytes in the first if and 7 in the second one. You could print the `parser.save()` before each instruction (including the instruction debug) and see how the jump is taken. We should ideally land right after the `End`.

Note that if you don't jump at the correct place, you may interpret arbitrary code (including code that was not present, due to WASM not having aligned bytecode). For example, if you jump on the byte preceding the `End` of the first if, you'll see another if because of the `4` of `local.set 4`, and for the second if you'll see `br` because of the 12 (offset of `i32.store`).

zhouwfang:
@ia0 I'd like to know how to get around the following "cannot borrow `*self` as mutable more than once at a time" error. Thanks.

In `Thread::step()`, I basically use the following function to take the jump
```
fn take_jump(&mut self, ls_idx: Option<usize>, side_tables: &[Vec<SideTableEntry>]) {
    let parser =  &mut self.parser;
    self.frame().take_jump(parser, ls_idx, side_tables);
}
```
where `self.frame()` returns `&mut Frame<'m>`.

crates/interpreter/src/exec.rs
zhouwfang: Does this comment still apply?

zhouwfang: I see. Thanks!

crates/interpreter/src/exec.rs
```diff
@@ -773,28 +789,44 @@ impl<'m> Thread<'m> {
...
                 }
+                self.frame().skip_jump();
             }
             BrTable(ls, ln) => {
                 let i = self.pop_value().unwrap_i32() as usize;
-                return Ok(self.pop_label(inst, ls.get(i).cloned().unwrap_or(ln)));
+                return Ok(self.pop_label(
+                    inst,
+                    ls.get(i).cloned().unwrap_or(ln),
+                    ls.get(i).map(|_| i),
```
ia0, line 828: Actually, if we make this `ls.get(i).map_or(0, |_| i + 1)` then we don't need an `Option` and we can use `0` instead of `None` for the other calls. This should also improve performances a bit by corner-casing only the BrTable.

crates/interpreter/src/exec.rs
```diff
@@ -1617,3 +1716,11 @@ fn memory_too_small(x: usize, n: usize, mem: &Memory) {
     #[cfg(feature = "debug")]
     eprintln!("Memory too small: {x} + {n} > {}", mem.len());
 }
+
+fn index_in_side_tables(module: &Module, index: u32) -> Option<usize> {
+    if module.side_table(index as usize).is_empty() {
```
ia0, line 1721: Why does it matter whether the side-table is empty? It's fine for a function to not have any jumps (and thus an empty table).

ia0 (commented):
Thanks! It looks pretty good.

crates/interpreter/src/exec.rs
```diff
@@ -773,28 +789,44 @@ impl<'m> Thread<'m> {
             Unreachable => return Err(trap()),
             Nop => (),
             Block(b) => self.push_label(self.blocktype(inst, &b), LabelKind::Block),
-            Loop(b) => self.push_label(self.blocktype(inst, &b), LabelKind::Loop(saved)),
+            Loop(b) => {
+                let side_table_idx = self.frame().side_table_indices.1;
+                self.push_label(
+                    self.blocktype(inst, &b),
+                    LabelKind::Loop(LoopState { parser_data: saved, side_table_idx }),
```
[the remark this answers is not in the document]
zhouwfang: It is a good idea to debug `val_cnt` and `pop_cnt` incrementally by label. Is there a reason to try loop first?

crates/interpreter/src/module.rs
```diff
@@ -199,7 +190,7 @@ impl<'m> Module<'m> {
...
         let mut parser = self.section(SectionId::Code).unwrap();
         for i in 0 .. parser.parse_vec().into_ok() {
```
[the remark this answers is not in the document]
zhouwfang: Yeah, I thought so too. Added a TODO.

crates/interpreter/src/exec.rs
zhouwfang: It was necessary when the side table computation was done in `Module`. Will try removing it after the lifetime issue is fixed.

zhouwfang:
I'd like to refactor incrementally. After storing `side_table: &'m [SideTableEntry]` in `Frame`, there appear to be some lifetime and " cannot borrow `*self` as mutable because it is also borrowed as immutable" errors. Inserted the errors in the code. PTAL.

crates/interpreter/src/exec.rs
```diff
@@ -192,10 +194,10 @@ impl<'m> Store<'m> {
             let x = parser.parse_funcidx().into_ok();
             let ptr = self.func_ptr(inst_id, x);
             let inst_id = ptr.instance().unwrap_wasm();
-            let mut parser = self.insts[inst_id].module.func(ptr.index());
+            let (mut parser, side_table) = self.insts[inst_id].module.func(ptr.index());
             let mut locals = Vec::new();
             append_locals(&mut parser, &mut locals);
-            let thread = Thread::new(parser, Frame::new(inst_id, 0, &[], locals));
+            let thread = Thread::new(parser, Frame::new(inst_id, 0, &[], locals, side_table));
             let result = thread.run(self)?;
```
zhouwfang, line 201:
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src/exec.rs:201:26
    |
107 | impl<'m> Store<'m> {
    |      -- lifetime `'m` defined here
...
197 |             let (mut parser, side_table) = self.insts[inst_id].module.func(ptr.index());
    |                                            --------------------------------------------
    |                                            |
    |                                            immutable borrow occurs here
    |                                            argument requires that `self.insts` is borrowed for `'m`
...
201 |             let result = thread.run(self)?;
    |                          ^^^^^^^^^^^^^^^^ mutable borrow occurs here

```

crates/interpreter/src/exec.rs
```diff
@@ -220,11 +223,11 @@ impl<'m> Store<'m> {
...
         let x = ptr.index();
         let t = inst.module.func_type(x);
-        let mut parser = inst.module.func(x);
+        let (mut parser, side_table) = inst.module.func(x);
         check_types(&t.params, &args)?;
         let mut locals = args;
         append_locals(&mut parser, &mut locals);
-        let frame = Frame::new(inst_id, t.results.len(), &[], locals);
+        let frame = Frame::new(inst_id, t.results.len(), &[], locals, side_table);
         Thread::new(parser, frame).run(self)
```
zhouwfang, line 231:
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src/exec.rs:231:9
    |
212 |     pub fn invoke<'a>(
    |                   -- lifetime `'a` defined here
...
223 |         let inst = &self.insts[inst_id];
    |                     ---------- immutable borrow occurs here
...
231 |         Thread::new(parser, frame).run(self)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         mutable borrow occurs here
    |         returning this value requires that `self.insts` is borrowed for `'a`
```

crates/interpreter/src/exec.rs
```diff
@@ -1377,12 +1398,12 @@ impl<'m> Thread<'m> {
             }
```
zhouwfang, line 1398:
```
error[E0621]: explicit lifetime required in the type of `store`
    --> src/exec.rs:1401:40
     |
1383 |     fn invoke(mut self, store: &mut Store<'m>, ptr: Ptr) -> Result<ThreadResult<'m>, Error> {
     |                                -------------- help: add explicit lifetime `'m` to the type of `store`: `&'m mut Store<'m>`
...
1401 |         let (mut parser, side_table) = store.insts[inst_id].module.func(ptr.index());
     |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ lifetime `'m` required

```

crates/interpreter/src/exec.rs
```diff
@@ -736,7 +739,8 @@ impl<'m> Thread<'m> {
 
     fn const_expr(store: &mut Store<'m>, inst_id: usize, mut_parser: &mut Parser<'m>) -> Val {
         let parser = mut_parser.clone();
-        let mut thread = Thread::new(parser, Frame::new(inst_id, 1, &[], Vec::new()));
+        static EMPTY_VEC: Vec<SideTableEntry> = Vec::new();
```
zhouwfang, line 742: Happy to know if there is a better way.

crates/interpreter/src/exec.rs
```diff
@@ -114,9 +115,10 @@ impl<'m> Store<'m> {
     /// The memory is not dynamically allocated and must thus be provided. It is not necessary for
     /// the memory length to be a multiple of 64kB. Execution will trap if the module tries to
     /// access part of the memory that does not exist.
-    pub fn instantiate(
-        &mut self, module: Module<'m>, memory: &'m mut [u8],
-    ) -> Result<InstId, Error> {
+    pub fn instantiate<'a>(
+        &'a mut self, module: Module<'m>, memory: &'m mut [u8],
+    ) -> Result<InstId, Error>
+    where 'a: 'm {
```
ia0, line 121: This is the problem. The actual bound should be the opposite, but we shouldn't need to specify it.

crates/interpreter/src/module.rs
```diff
@@ -199,13 +190,14 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
-    pub(crate) fn func(&self, x: FuncIdx) -> Parser<'m> {
+    // TODO(dev/fast-interp): Improve the performance of such accessor functions from O(n) to O(1).
+    pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &[SideTableEntry]) {
```
ia0, line 194: That's probably the problem. The side-table is long-lived.
```suggestion
    pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [SideTableEntry]) {
```

crates/interpreter/src/exec.rs
```diff
@@ -209,7 +211,8 @@ impl<'m> Store<'m> {
     /// may be corrupted.
     pub fn invoke<'a>(
         &'a mut self, inst: InstId, name: &str, args: Vec<Val>,
-    ) -> Result<RunResult<'a, 'm>, Error> {
+    ) -> Result<RunResult<'a, 'm>, Error>
+    where 'a: 'm {
```
ia0, line 215: Same here. We should never have `'a: 'm`. At best we can have `'m: 'a` but it should be implicit. We shouldn't need to write it.

crates/interpreter/src/exec.rs
ia0: Yes, you can just use `&[]` directly.

crates/interpreter/src/exec.rs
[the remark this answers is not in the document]
ia0: This was not about debugging. We use `take_jump` for Block and If in `pop_label` but do it manually with this saved state for Loop. We should try using `take_jump` for Loop too (but in a different PR and measuring the performance impact).

crates/interpreter/src/module.rs
zhouwfang: I had tried this, and the error is 
```
error: lifetime may not live long enough
   --> src/module.rs:200:24
    |
105 | impl<'m> Module<'m> {
    |      -- lifetime `'m` defined here
...
194 |     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [SideTableEntry]) {
    |                        - let's call the lifetime of this reference `'1`
...
200 |                 return (parser, &self.side_tables[i]);
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method was supposed to return data with lifetime `'m` but it is returning data with lifetime `'1`
```
Then I had tried 
```
  pub(crate) fn func<'a>(&'a self, x: FuncIdx) -> (Parser<'m>, &'m [SideTableEntry])
    where 'a: 'm
```
and the other errors in `exec.rs` remain.

crates/interpreter/src/exec.rs
zhouwfang: It and the one below were suggested by the compiler...

crates/interpreter/src/module.rs
ia0: Oh yes, I tried to do it and figured out the problem, it's because `Module::side_tables` is still owned by `Module` instead of being borrowed. Ideally we need to implement the TODO, but that's going to be too much work for this PR. I pushed a quick fix to unblock.

zhouwfang: Thanks!

crates/interpreter/src/module.rs
```diff
@@ -54,7 +55,8 @@ impl<'m> Module<'m> {
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
         let side_tables = validate(binary)?;
         let mut module = unsafe { Self::new_unchecked(binary) };
-        module.side_tables = side_tables;
+        // TODO(dev/fast-interp): We should take a buffer as argument to write to.
```
zhouwfang, line 58: Could you elaborate on why the TODO here is better than the Box solution below? Thanks.

crates/interpreter/src/module.rs
zhouwfang: Could you elaborate on why this is a problem in Rust("Module::side_tables is still owned by Module instead of being borrowed")?

crates/interpreter/src/exec.rs
```diff
@@ -1096,12 +1135,14 @@ impl<'m> Thread<'m> {
...
-    fn skip_to_else(&mut self, inst: &mut Instance<'m>) {
-        inst.module.skip_to_else(&mut self.parser);
-    }
-
-    fn skip_to_end(&mut self, inst: &mut Instance<'m>, l: LabelIdx) {
-        inst.module.skip_to_end(&mut self.parser, l);
+    fn take_jump(
+        &mut self, non_default_label_index: Option<usize>, side_tables: &[Vec<SideTableEntry>],
+    ) {
+        self.frames.last_mut().unwrap().take_jump(
```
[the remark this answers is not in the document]
zhouwfang: Done. `take_jump()` would also need to take the current parser position.

crates/interpreter/src/exec.rs
zhouwfang: Very clever idea!

crates/interpreter/src/exec.rs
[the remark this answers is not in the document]
zhouwfang: I see. Please take a look at https://github.com/zhouwfang/wasefire/pull/4.

crates/interpreter/src/module.rs
ia0: The Box solution is just a workaround for testing because it's leaking memory. Ultimately we want a solution that we can use in production. It's not yet clear how it will look like. It could be `Module::new()` taking a buffer to write to and `new_unchecked()` a buffer to read from, or it could be something else. It depends on the "streaming" design. But the buffer solution could be a temporary solution until the streaming design is clear.

crates/interpreter/src/module.rs
```diff
@@ -91,6 +76,14 @@ impl<'m> Module<'m> {
         module
     }
 
+    pub fn side_tables(&self) -> &[Vec<SideTableEntry>] {
+        self.side_tables
+    }
+
+    pub fn side_table(&self, i: usize) -> &[SideTableEntry] {
+        &self.side_tables[i]
+    }
```
ia0, line 85: It seems safe to delete those functions now, since their type is wrong. We want to return `&'m [...]`.

crates/interpreter/src/exec.rs
```diff
@@ -114,8 +115,8 @@ impl<'m> Store<'m> {
     /// The memory is not dynamically allocated and must thus be provided. It is not necessary for
     /// the memory length to be a multiple of 64kB. Execution will trap if the module tries to
     /// access part of the memory that does not exist.
-    pub fn instantiate(
-        &mut self, module: Module<'m>, memory: &'m mut [u8],
+    pub fn instantiate<'a>(
+        &'a mut self, module: Module<'m>, memory: &'m mut [u8],
```
ia0, line 119: We can have `'a` implicit again. I forgot to do it in the quick fix.

crates/interpreter/src/exec.rs
```diff
@@ -765,28 +766,43 @@ impl<'m> Thread<'m> {
...
             },
             Else => {
-                self.skip_to_end(inst, 0);
+                self.take_jump(0);
                 return Ok(self.exit_label());
             }
-            End => return Ok(self.exit_label()),
-            Br(l) => return Ok(self.pop_label(inst, l)),
+            End => {
+                return Ok(self.exit_label());
```
ia0, line 791: By removing the semi-colon we can get it on the same line as the pattern, as it was before.
```suggestion
                return Ok(self.exit_label())
```

crates/interpreter/src/exec.rs
```diff
@@ -1409,12 +1425,38 @@ struct Frame<'m> {
...
+    fn take_jump(&mut self, parser_pos: &'m [u8], offset: usize) -> Parser<'m> {
+        self.side_table = Self::jump(self.side_table, offset as isize);
+        let entry = self.side_table[0].view();
+        self.side_table = Self::jump(self.side_table, entry.delta_stp as isize);
+        unsafe { Parser::new(Self::jump(parser_pos, entry.delta_ip as isize)) }
+    }
+
+    // TODO(dev/fast-interp): Add debug asserts when `off` is positive and negative, and `toctou`
+    // support.
+    fn jump<T>(cur: &'m [T], off: isize) -> &'m [T] {
```
ia0, line 1453: We should rename this function now that it's also used by `skip_jump` for the side-table slice. We should also move it out of `Frame` as a free function. Maybe something like:

```rust
fn offset_front<T>(slice: &[T], count: isize) -> &[T] {
   // ...
}
```

ia0 (commented):
Looks good to merge modulo the nits.

crates/interpreter/src/module.rs
ia0: That's because in Rust you can't borrow a specific field of a struct through a method of that struct without borrowing the whole struct. There are people looking into fixing this. In our particular case, by making the struct (`Module`) borrow its field in the first place, then methods can simply copy that reference instead of borrowing it (thus borrowing the whole struct for the lifetime of that field).

PR #694, opened 2024-12-02
Use bytemuck for ArrayU32

ia0:
The motivation is to reduce the number of `unsafe` proofs, because they are manual and error-prone. The invariant is weaker (`Pod` rather than exactly something equivalent to `[u32; N]`), but is enough to be safe.


ia0:
Waiting for approval from @cramertj before merging.

cramertj:
LGTM!

PR #696, opened 2024-12-04
Reduce delta_ip by 1 for "if without else" in validation

zhouwfang:
Jumping from "if without else" should be 1 step less than jumping from "if with else". For "if without else", we don't want to jump over the "end" because `exit_label()` need to be [called](https://github.com/zhouwfang/wasefire/blob/de6c7e889865736cb1f11877d5d4e1e9f50729a8/crates/interpreter/src/exec.rs#L788) at the "end". But for "if with else", we want to jump over the "else" because otherwise we would be forced to jump to the "end" at "else" (normally, we would reach "else" after executing the true branch of "if", and we should not execute the "else" branch).

This is equivalent to https://github.com/google/wasefire/blob/main/crates/interpreter/src/parser.rs#L566-L569, which will be deprecated with the side table applied in `exec.rs`.

#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -426,18 +426,8 @@ impl SideTable {
         self.entries.push(None);
     }
 
-    fn stitch(
-        &mut self, source: SideTableBranch, target: SideTableBranch, is_if_without_else: bool,
-    ) -> CheckResult {
-        let mut delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
-        if is_if_without_else {
-            let Some(reduced_delta_ip) = delta_ip.checked_sub(1) else {
-                #[cfg(feature = "debug")]
-                eprintln!("side-table subtraction overflow {delta_ip} - 1");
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -426,8 +426,18 @@ impl SideTable {
...
+        &mut self, source: SideTableBranch, target: SideTableBranch, is_if_without_else: bool,
+    ) -> CheckResult {
+        let mut delta_ip = Self::delta(source, target, |x| x.parser.as_ptr() as isize)?;
+        if is_if_without_else {
+            let Some(reduced_delta_ip) = delta_ip.checked_sub(1) else {
+                #[cfg(feature = "debug")]
+                eprintln!("side-table subtraction overflow {delta_ip} - 1");
+                return Err(unsupported(if_debug!(Unsupported::SideTable)));
+            };
+            delta_ip = reduced_delta_ip;
```
zhouwfang, line 439: Happy to know if there is a more idiomatic way.

ia0 (approved):

PR #700, opened 2024-12-05
Increment delta_stp by 1 for "if with else" in validation

zhouwfang:
When "if" has an "else", we should skip the entry for "else" in the side table. It was addressed in #690 during execution and should be moved to validation.

#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -596,7 +596,9 @@ impl<'a, 'm> Expr<'a, 'm> {
             Else => {
                 match core::mem::replace(&mut self.label().kind, LabelKind::Block) {
                     LabelKind::If(source) => {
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -596,7 +599,7 @@ impl<'a, 'm> Expr<'a, 'm> {
             Else => {
                 match core::mem::replace(&mut self.label().kind, LabelKind::Block) {
                     LabelKind::If(source) => {
-                        self.side_table.stitch(source, self.branch_target(source.result))?
+                        self.side_table.stitch(source, self.branch_target(source.result, true))?
```
ia0, line 602: Similar to the other PR. We can `let mut target = self.branch_target(source.result)` first, increment `target.side_table`, then stitch.

ia0 (approved):

PR #704, opened 2024-12-09
Use edition 2024 for all crates

ia0:


crates/runner-host/crates/web-server/src/lib.rs
```diff
@@ -120,10 +120,9 @@ async fn handle(mut ws: WebSocket, client: ClientInput) {
     });
     cmd_sender.send(Command::Connected).await.unwrap();
     let board_config = Command::BoardConfig {
-        components: vec![
-            Component::Button { id: BUTTON_ID },
-            Component::MonochromeLed { id: LED_ID },
-        ],
+        components: vec![Component::Button { id: BUTTON_ID }, Component::MonochromeLed {
+            id: LED_ID,
+        }],
```
ia0-review, line 125: This looks worse. Seems to be https://github.com/rust-lang/rustfmt/issues/3370#issuecomment-1288226646.

ia0-review (approved):

crates/runner-host/crates/web-server/src/lib.rs
ia0: I agree. I didn't find a workaround. Let's keep it like that for now.

ia0-review (approved):

PR #705, opened 2024-12-11
Use instead of restoring side table for Loop

zhouwfang:
On my Linux Docker container:

1. 

CoreMark results of this PR: 37.991516 (in 18.695s), 36.7287 (in 19.258s)

CoreMark results of https://github.com/google/wasefire/pull/690: 37.101162 (in 19.133s), 37.39716 (in 18.996s)

Look very similar. Not sure about `nordic`.

2. 

CoreMark results of `main`: 28.791477 (in 18.09s), 28.814291 (in 17.734s)

So there does seem some minor performance improvement with `delta_ip` and `delta_stp`.

https://github.com/google/wasefire/issues/46

### DIFF
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -15,6 +15,7 @@
 // TODO: Some toctou could be used instead of panic.
 use alloc::vec;
 use alloc::vec::Vec;
+use core::fmt::Debug;
 
 use crate::error::*;
 use crate::module::*;
@@ -759,14 +760,13 @@ impl<'m> Thread<'m> {
 
     fn step(mut self, store: &mut Store<'m>) -> Result<ThreadResult<'m>, Error> {
         use Instr::*;
-        let saved = self.parser.save();
         let inst_id = self.frame().inst_id;
         let inst = &mut store.insts[inst_id];
         match self.parser.parse_instr().into_ok() {
             Unreachable => return Err(trap()),
             Nop => (),
             Block(b) => self.push_label(self.blocktype(inst, &b), LabelKind::Block),
-            Loop(b) => self.push_label(self.blocktype(inst, &b), LabelKind::Loop(saved)),
+            Loop(b) => self.push_label(self.blocktype(inst, &b), LabelKind::Loop),
             If(b) => match self.pop_value().unwrap_i32() {
                 0 => {
                     self.take_jump(0);
@@ -982,11 +982,11 @@ impl<'m> Thread<'m> {
         self.frames.last_mut().unwrap()
     }
 
-    fn labels(&mut self) -> &mut Vec<Label<'m>> {
+    fn labels(&mut self) -> &mut Vec<Label> {
         &mut self.frame().labels
     }
 
-    fn label(&mut self) -> &mut Label<'m> {
+    fn label(&mut self) -> &mut Label {
         self.labels().last_mut().unwrap()
     }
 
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/exec.rs
```diff
@@ -1064,11 +1057,7 @@ impl<'m> Thread<'m> {
         self.values().drain(values_len - values_cnt .. values_len - arity);
         self.label().values_cnt += arity;
         match kind {
-            LabelKind::Loop(state) => unsafe {
-                self.frame().side_table = state.side_table;
-                self.parser.restore(state.parser_data)
-            },
-            LabelKind::Block | LabelKind::If => self.take_jump(offset),
+            LabelKind::Block | LabelKind::If | LabelKind::Loop => self.take_jump(offset),
```
ia0, line 1060: The match can be removed. And actually as a consequence, it seems the `kind` field can be removed from `LabelKind`.

crates/interpreter/src/valid.rs
```diff
@@ -979,3 +976,12 @@ impl<'a, 'm> Expr<'a, 'm> {
         }
     }
 }
+
+fn offset_parser(branch: SideTableBranch, off: isize) -> &[u8] {
```
ia0, line 980: This function should be shared. It could either be in `module.rs`, `parser.rs`, or a new `util.rs`.

crates/interpreter/src/valid.rs
```diff
@@ -586,7 +586,9 @@ impl<'a, 'm> Expr<'a, 'm> {
             Block(b) => self.push_label(self.blocktype(&b)?, LabelKind::Block)?,
             Loop(b) => {
                 let type_ = self.blocktype(&b)?;
-                self.push_label(type_, LabelKind::Loop(self.branch_target(type_.params.len())))?
+                let mut target = self.branch_target(type_.params.len());
+                target.parser = offset_parser(target, -2);
```
ia0, line 590: This only works with simple block types. The correct solution is to save the parser position before parsing the instruction line 561 and doing `target.parser = saved` here.

ia0:
On nordic I get:
- This pull request (with the suggested changes): 0.13723445 (in 146.037s) and 6660 bytes of RAM
- This development branch: 0.13704263 (in 146.247s) and 7588 of RAM
- The main branch: 0.09060761 (in 221.04s) and 5416 of RAM

So we should merge this PR, not only because it uses less RAM, but also because it simplifies the code.

zhouwfang:
> On nordic I get:
> 
> * This pull request (with the suggested changes): 0.13723445 (in 146.037s) and 6660 bytes of RAM
> * This development branch: 0.13704263 (in 146.247s) and 7588 of RAM
> * The main branch: 0.09060761 (in 221.04s) and 5416 of RAM
> 
> So we should merge this PR, not only because it uses less RAM, but also because it simplifies the code.

How do you measure RAM?

ia0 (approved):

crates/interpreter/src/valid.rs
ia0: I believe this would happen if there are more than 127 types in the module because then if a loop has parameters, then it has to use a type index (the `x:s33` in https://webassembly.github.io/spec/core/binary/instructions.html#control-instructions) which will need 2 bytes.

ia0:
> How do you measure RAM?

This is specific to Nordic: https://github.com/google/wasefire/blob/15823da66f27984a5147620d8fbe1eb7d0c868e5/crates/wasm-bench/src/target/nordic.rs#L70

PR #706, opened 2024-12-18
Fix val_cnt for if

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -594,8 +594,10 @@ impl<'a, 'm> Expr<'a, 'm> {
             }
             If(b) => {
                 self.pop_check(ValType::I32)?;
-                let branch = self.branch_source();
-                self.push_label(self.blocktype(&b)?, LabelKind::If(branch))?;
+                let mut branch = self.branch_source();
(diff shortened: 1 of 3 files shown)


zhouwfang:
@ia0 I posted https://github.com/zhouwfang/wasefire/pull/5 to show the debug asserts added for `val_cnt` and `pop_cnt` with a question. Could you take a look? Thanks.

crates/interpreter/src/valid.rs
```diff
@@ -594,8 +594,10 @@ impl<'a, 'm> Expr<'a, 'm> {
             }
             If(b) => {
                 self.pop_check(ValType::I32)?;
-                let branch = self.branch_source();
-                self.push_label(self.blocktype(&b)?, LabelKind::If(branch))?;
+                let mut branch = self.branch_source();
+                let type_ = self.blocktype(&b)?;
+                branch.result = type_.results.len();
```
ia0, line 599: Instead of doing this here, we can do it in the `Else`.

ia0 (commented):
I'm pushing a change with my suggestions.

ia0 (approved):

PR #707, opened 2024-12-23
Add prev_stack in Frame

zhouwfang:
It improves the performance by removing the costly `last_frame_values_cnt()`.

On linux, `CoreMark result: 39.795715 (in 17.853s)`. cf. CoreMark result was about 37 in 19s in #705.

#46

### DIFF
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -197,7 +197,7 @@ impl<'m> Store<'m> {
             let (mut parser, side_table) = self.insts[inst_id].module.func(ptr.index());
             let mut locals = Vec::new();
             append_locals(&mut parser, &mut locals);
-            let thread = Thread::new(parser, Frame::new(inst_id, 0, &[], locals, side_table, None));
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/exec.rs
```diff
@@ -1412,15 +1416,25 @@ struct Frame<'m> {
...
     side_table: &'m [SideTableEntry],
+    /// Total length of the value stack in the thread prior to this frame.
+    prev_stack: usize,
 }
 
 impl<'m> Frame<'m> {
     fn new(
         inst_id: usize, arity: usize, ret: &'m [u8], locals: Vec<Val>,
-        side_table: &'m [SideTableEntry],
+        side_table: &'m [SideTableEntry], prev_stack: Option<usize>,
```
ia0, line 1426: Why use an `Option`? Callers can simply use `0`.

ia0 (commented):
Thanks! LGTM modulo the `Option<usize>` which seems redundant.

ia0 (approved):

PR #710, opened 2024-12-25
Apply `val_cnt` and `pop_cnt` to `pop_label()`

zhouwfang:
- Apply `val_cnt` and `pop_cnt`.
- Remove `Label` by introducing `Frame::labels_cnt`.

There is not much performance improvement on Linux due to this PR, which replaces 
```
let values_cnt: usize = frame.labels[i ..].iter().map(|label| label.values_cnt).sum();
```
in `pop_label()` by `pop_cnt + val_cnt`.

#46

### DIFF
--- a/crates/interpreter/src/exec.rs
+++ b/crates/interpreter/src/exec.rs
@@ -18,7 +18,7 @@ use alloc::vec::Vec;
 
 use crate::error::*;
 use crate::module::*;
-use crate::side_table::SideTableEntry;
+use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
 use crate::util::*;
@@ -1031,12 +1031,9 @@ impl<'m> Thread<'m> {
             return self.exit_frame();
         }
         frame.labels_cnt = i;
-        let side_table_entry = frame.side_table[offset].view();
-        let values_len = self.values().len();
-        self.values().drain(
-            (values_len as u32 - (side_table_entry.pop_cnt + side_table_entry.val_cnt)) as usize
-                .. values_len - side_table_entry.val_cnt as usize,
-        );
+        let SideTableEntryView { val_cnt, pop_cnt, .. } = frame.side_table[offset].view();
+        let val_pos = self.values().len() - val_cnt as usize;
+        self.values().drain(val_pos - pop_cnt as usize .. val_pos);
         self.take_jump(offset);
         ThreadResult::Continue(self)
     }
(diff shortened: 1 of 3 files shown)


crates/interpreter/src/exec.rs
```diff
@@ -1407,19 +1382,18 @@ struct Frame<'m> {
     arity: usize,
     ret: &'m [u8],
     locals: Vec<Val>,
-    labels: Vec<Label>,
     side_table: &'m [SideTableEntry],
     /// Total length of the value stack in the thread prior to this frame.
     prev_stack: usize,
+    labels_cnt: usize,
```
zhouwfang, line 1388: @ia0 Do you expect this to be removed eventually? Only from the performance perspective, I'm not sure how much gain there would be by removing it. WDYT?

crates/interpreter/src/exec.rs
```diff
@@ -1032,45 +1020,41 @@ impl<'m> Thread<'m> {
...
+        let i = frame.labels_cnt - l as usize - 1;
         if i == 0 {
             return self.exit_frame();
         }
-        let frame = self.frame();
-        let values_cnt: usize = frame.labels[i ..].iter().map(|label| label.values_cnt).sum();
-        let Label { arity, .. } = frame.labels.drain(i ..).next().unwrap();
+        frame.labels_cnt = i;
+        let side_table_entry = frame.side_table[offset].view();
         let values_len = self.values().len();
```
ia0, line 1035: This would simplify next line.
```suggestion
        let values_len = self.values().len() - side_table_entry.val_cnt as usize;
```

crates/interpreter/src/exec.rs
ia0: Yes, I expect this to be eventually removed (as well as `push_label()`). This is not only cleaner, but also useful for when we'll write a compiler to our own bytecode (or an ISA) because it means we don't need to compile instructions like `Block` or `Loop` that don't do anything except indicate a scope.

But this can be done in a separate PR.

ia0 (approved):
Thanks! That's a really nice cleanup. I'll push a small review commit and merge.

ia0 (approved):

ia0:
Before change:

```
Peak RAM usage: 6692
CoreMark result: 0.13914786 (in 144.034s)
```

After change:

```
Peak RAM usage: 6244
CoreMark result: 0.1541212 (in 130.039s)
```

zhouwfang:
@ia0 Could you also get comparison in code size? Thanks!

ia0:
> @ia0 Could you also get the current code size? Thanks!

You just need to run `rust-size ../../target/thumbv7em-none-eabi/release/wasm-bench` (you might need `cargo install cargo-binutils` first) after `./run.sh nordic base`. For this PR, it went from 144KiB to 141KiB.

PR #711, opened 2024-12-27
Generalize side table

zhouwfang:
I'd like to get some feedback about the design: the "section table" is used to improve the performance of [these accessors](https://github.com/zhouwfang/wasefire/blob/715bbaa6e2f79e40464e9b85ad9fc506c2d02d57/crates/interpreter/src/module.rs#L176-L209) in `module.rs`, which are used in execution. For these three accessors that return `Parser`, we can precompute the `ParseRange` during validation, and create a parser based on the range during execution.

I'm not sure about [the other accessors](https://github.com/zhouwfang/wasefire/blob/715bbaa6e2f79e40464e9b85ad9fc506c2d02d57/crates/interpreter/src/module.rs#L119-L174) that return types or even `Option<ExportDesc>`. IIUC, we want to store the section table in flash as the side table. Does that mean we would have to design some encoding rules for these return types?

**Update 1**

We will only optimize [`func()`](https://github.com/zhouwfang/wasefire/blob/715bbaa6e2f79e40464e9b85ad9fc506c2d02d57/crates/interpreter/src/module.rs#L188) and [`func_type()`](https://github.com/zhouwfang/wasefire/blob/715bbaa6e2f79e40464e9b85ad9fc506c2d02d57/crates/interpreter/src/module.rs#L119) in `module.rs`, because the other accessors should be fast.

**Update 2**

- Rename the previous `SideTableEntry` as `BranchTableEntry`.
- Create per-function `SideTable` to include branch table and other function metadata.

#46


ia0:
On the principle this looks good. From an organization point of view, I would put everything in the side table (which represents pre-computed information on the wasm bytecode). This side table would contain information about how to take branches/jumps, where to find functions, etc. It needs to be serializable with easy random access.

Regarding the other type of accessors, it's probably enough to store their position in the wasm bytecode and deserialize them on demand. This is still O(1). We can probably avoid storing the "end" parser position by always using the end of the wasm bytecode. For functions, we could also store only one position, which would be right before the `u32` encoded size, meaning we need to `parse_u32` and `split_at` for each access, but we save 50% storage, so it's some tradeoff.

crates/interpreter/src/side_table.rs
```diff
@@ -17,7 +17,7 @@ use crate::error::*;
 
 #[derive(Default, Copy, Clone, Debug)]
 #[repr(transparent)]
-pub struct SideTableEntry(u64);
+pub struct SideTableEntry(u128);
```
zhouwfang, line 20: `u128` may be unnecessarily big, but it may not be harmful to leave it for now as we would need to add offsets to it later. WDYT?

zhouwfang:
@ia0 Could you take another look at the design? Thanks.

ia0 (commented):
The (generalized) side table should logically be a mapping from function index to function metadata, where a function metadata has the following components:
- The type (index) of the function.
- The body (parser) of the function.
- The "side table" of the function. Let's call this branch table.

We want to serialize this logical value somehow. One idea is to have something like this:

```
side-table := index flatten-metadata (16-bits aligned)
index := u16 u16 ... u16 (N - 1 times where N is the number of functions in the module)
flatten-metadata := metadata metadata ... metadata (N times where `flatten-metadata[i + 1]` starts at byte position `index[i]`)
metadata := parser:(u32 u32) type:u16 branch-table (if 32-bits aligned)
          | type:u16 parser:(u32 u32) branch-table (otherwise)
branch-table := branch branch ... branch
branch := delta-ip:u16 delta-stp:u16 val-cnt:u8 pop-cnt:u12
```

zhouwfang:
> metadata := parser:(u32 u32) type:u16 branch-table (if 32-bits aligned)
>           | type:u16 parser:(u32 u32) branch-table (otherwise)

Could you elaborate on the two cases here? Why is the type located differently?

zhouwfang:
@ia0 Please take a look at the generalized `SideTableEntry`. Let's first agree on the design, and I'll add unit tests later. Thanks.

ia0:
> > metadata := parser:(u32 u32) type:u16 branch-table (if 32-bits aligned)
> > | type:u16 parser:(u32 u32) branch-table (otherwise)
> 
> Could you elaborate on the two cases here? By "otherwise", do you mean 16 bits aligned?

Yes. Every "item" is at least 16 bits aligned. `metadata` may either be at least 32-bits aligned in which case the parser comes first, or exactly 16-bits aligned (in which case the type index comes first).

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,73 @@
...
+
 use crate::bit_field::*;
 use crate::error::*;
 
+pub const MEMORY_ALIGN: usize = 16;
+pub const MASK_16_BITS_AS_U32: u32 = 0xffff;
+pub const MASK_16_BITS_AS_U64: u64 = 0xffff;
+pub const MASK_48_BITS: u64 = 0xffff_ffff_ffff;
+
+pub struct SideTable(Vec<u16>);
```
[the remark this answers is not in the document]
zhouwfang: Instead of `parser: Parser<'m>` in `Metadata`, I have `parser_size: usize` and `parser_start: &'m [u8]` which will be used to create `parser: Parser<'m>`, as they seem to be what we would store in flash. WDYT?

zhouwfang:
> > > metadata := parser:(u32 u32) type:u16 branch-table (if 32-bits aligned)
> > > | type:u16 parser:(u32 u32) branch-table (otherwise)
> > 
> > 
> > Could you elaborate on the two cases here? By "otherwise", do you mean 16 bits aligned?
> 
> Yes. Every "item" is at least 16 bits aligned. `metadata` may either be at least 32-bits aligned in which case the parser comes first, or exactly 16-bits aligned (in which case the type index comes first).

Why does the order matter? We can read exactly based on how we write, and the alignment is known by design when we write.

crates/interpreter/src/side_table.rs
```diff
@@ -15,11 +15,28 @@
 use crate::bit_field::*;
 use crate::error::*;
 
+#[allow(dead_code)]
+pub struct SideTable<'m> {
+    indices: &'m [u16], // first index is omitted (it is always 0)
+    metadata_array: &'m [Metadata<'m>],
```
ia0, line 21: This doesn't work and that's why we have `indices`. Metadata has variable size because the branch table has variable size. We need an uninterpreted `&'m [u16]` and interpret it on demand for a given index.

And actually, thinking about it, the code would be more uniform if we store 2 more indices: 0 at the start and `metadata.len()` at the end, such that we can simply `&self.metadata[self.indices[i] .. self.indices[i + 1]]` to get the proper uninterpreted metadata slice.

crates/interpreter/src/side_table.rs
```diff
@@ -30,26 +47,43 @@ pub struct SideTableEntryView {
...
+        let entry = u16_array_to_u64(self.0);
+        let delta_ip = from_signed_field(Self::DELTA_IP_MASK, entry);
+        let delta_stp = from_signed_field(Self::DELTA_STP_MASK, entry);
+        let val_cnt = from_field(Self::VAL_CNT_MASK, entry);
+        let pop_cnt = from_field(Self::POP_CNT_MASK, entry);
+        BranchTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt }
     }
 }
+
+fn u64_to_u16_array(value: u64) -> [u16; 4] {
```
ia0, line 75: We shouldn't implement those ourselves. There's 2 options: either use `bytemuck` or actually inline the bitfield logic. If we split like `(u16, u16, u4, u12)` then we don't really need to go through a `u64`. The `[u16; 4]` is close enough.

crates/interpreter/src/side_table.rs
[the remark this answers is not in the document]
ia0: I don't think we need one more intermediary. We either want the serialized representation or the final usable representation. Maybe the answer is to simply have `struct Metadata<'m>(&'m [u16])` and implement `fn type_idx(&self) -> usize`, `fn parser(&self) -> Parser<'m>`, and `fn branch_table(&self) -> BranchTable<'m>` which would do the "parsing". Having a view is useful only if we don't want to "parse" multiple times, but maybe we won't do that. It depends if we store `Metadata<'m>` in the frame or just `BranchTable<'m>`.

ia0:
> > > > metadata := parser:(u32 u32) type:u16 branch-table (if 32-bits aligned)
> > > > | type:u16 parser:(u32 u32) branch-table (otherwise)
> > > 
> > > 
> > > Could you elaborate on the two cases here? By "otherwise", do you mean 16 bits aligned?
> > 
> > 
> > Yes. Every "item" is at least 16 bits aligned. `metadata` may either be at least 32-bits aligned in which case the parser comes first, or exactly 16-bits aligned (in which case the type index comes first).
> 
> Why does the order matter? We can read exactly based on how we write, and the alignment is known by design when we write.

For the CPUs we use it probably doesn't matter because they are quite simple. But Rust which supports much more, makes a distinction between aligned and unaligned accesses. But maybe the solution is to just use `&[u8]` everywhere and use `{to,from}_ne_bytes()` functions. The code will probably be optimized as expected.

Regarding the order, every metadata has an odd amount of `u16` because parser and branch-table have both an even amount of `u16`. So metadata are alternating between 32-bit aligned or only 16-bit aligned. If we want the u32 to be 32-bit aligned then the order matter (and the order is alternating).

There's the same problem for writing. We need to write with the proper alignment. It is not known, it depends how many branches are in each function before the current function. We could add a u16 padding after some functions to make sure all metadata start 32-bit aligned, so that even indices and odd indices can be used to tell the alignment.

But I think at this point, it's probably better to just not rely on alignment and read all values as `&[u8]`. All flash controllers we use so far support reading at byte granularity directly from linear memory. So it's just a matter of unaligned reads having the same performance as aligned reads for the CPU we consider.

crates/interpreter/src/side_table.rs
zhouwfang: I misunderstood "indices" which is more like offsets.

zhouwfang:
> Regarding the order, every metadata has an odd amount of `u16` because parser and branch-table have both an even amount of `u16`.

A branch table has 48 bits, so metadata should have an even number of `u16`?

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,73 @@
...
+
+    pub fn parser(&self) -> Parser<'m> {
+        unsafe {
+            Parser::new(&self.0[self.parser_size_start() + size_of::<u32>() .. self.parser_end()])
+        }
+    }
+
+    fn branch_table(&self) -> &[BranchTableEntry] {
+        let bytes = &self.0[self.parser_end() ..];
+        let branch_table_len = bytes.len() / 8;
```
zhouwfang, line 58: Rust question: how can I obtain the 8 statically?

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,73 @@
...
+pub type Parser<'m> = parser::Parser<'m, Use>;
+
+#[allow(dead_code)]
+pub struct SideTable<'m> {
+    indices: &'m [u16], // including 0 and the length of metadata_array
+    metadata: &'m Metadata<'m>,
+}
+
+impl<'m> SideTable<'m> {
+    fn metadata(&self, func_idx: usize) -> Metadata<'m> {
```
zhouwfang, line 33: I couldn't return a reference directly because of the error "An attempt to use index on a type which doesn't implement the std::ops::Index trait was performed."

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,74 @@
...
+            &self.0[self.parser_size_start() .. self.parser_size_start() + u32_size];
+        let parser_size = usize::from_ne_bytes(parser_size_bytes.try_into().unwrap());
+        self.parser_size_start() + u32_size + parser_size
+    }
+}
 
 #[derive(Default, Copy, Clone, Debug)]
 #[repr(transparent)]
-pub struct SideTableEntry(u64);
+pub struct BranchTableEntry([u8; 8]);
```
ia0, line 80: Do we really need `[u8; 8]` then? It seems we can already use `[u8; 6]`. And actually, we probably want `[u16; 3]`. We should probably dump the bitfield logic and bytemuck. The delta are really just `x[0] as i16` and `x[1] as i16`, while val and pop are `x[2] & 0xf` and `x[2] >> 4`.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,74 @@
...
 use crate::error::*;
+use crate::parser;
+use crate::toctou::Use;
+
+pub type Parser<'m> = parser::Parser<'m, Use>;
+
+#[allow(dead_code)]
+pub struct SideTable<'m> {
+    indices: &'m [u16], // including 0 and the length of metadata_array
+    metadata: &'m Metadata<'m>,
```
ia0, line 29: This should be `&'m [u16]` (or directly `Metadata<'m>`, it's a bit abusive and confusing though).

Note that indices are not byte (u8) offsets but u16 offsets, so we can range twice more.

crates/interpreter/src/side_table.rs
ia0: With the current definition of `Metadata` you don't need to return a reference. `Metadata` is already a reference.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,74 @@
...
+    fn metadata(&self, func_idx: usize) -> Metadata<'m> {
+        Metadata(
+            &self.metadata.0
+                [self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
+        )
+    }
+}
+
+#[allow(dead_code)]
+struct Metadata<'m>(&'m [u8]);
```
ia0, line 42: This should ideally be `&'m [u16]`.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,74 @@
...
+
+#[allow(dead_code)]
+impl<'m> Metadata<'m> {
+    pub fn type_idx(&self) -> usize {
+        let bytes = &self.0[.. size_of::<usize>()];
```
ia0, line 47: We shouldn't use 4 bytes for type idx, I expect 2 to be enough, so it's really just reading one element of the `[u16]`.

crates/interpreter/src/side_table.rs
ia0, line 42: We can make this type `Copy`.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,74 @@
...
+        usize::from_ne_bytes(bytes.try_into().unwrap())
+    }
+
+    pub fn parser(&self) -> Parser<'m> {
+        unsafe {
+            Parser::new(&self.0[self.parser_size_start() + size_of::<u32>() .. self.parser_end()])
```
ia0, line 53: This is wrong. The parser data is not in the side-table but in the module. The signature should be `fn parser(self, module: &'m [u8]) -> Parser<'m>` or probably better `fn parser(self, code: &'m [u8]) -> Parser<'m>` where `code` is the code section of the module, so we can save some range.

Also, let's just use constants for u16 offsets in the metadata:
- 0 for type idx
- 1 for parser start
- 3 for parser end
- 5 for branch table

crates/interpreter/src/side_table.rs
ia0: Feel free to call it offsets if you prefer. And now that we want those to be u16 offsets/indices, we should document it.

```rust
/// Indices in the `metadata` slice below for each function (including one past the last function).
```

crates/interpreter/src/side_table.rs
ia0: I'm not sure what the computation is here. We shouldn't use parser-end as an index into the metadata. It's an index into the module (or code section).

crates/interpreter/src/exec.rs
```diff
@@ -17,7 +17,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 
 use crate::error::*;
-use crate::module::*;
+use crate::module::{Parser, *};
```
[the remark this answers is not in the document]
zhouwfang: Both `module` and `side_table` define `Parser`, so the compiler is confused about which one to use.

crates/interpreter/src/side_table.rs
zhouwfang: Leave it as `[u16; 4]` for now. I'll working on changing it to `[u16; 3]` tomorrow.

crates/interpreter/src/side_table.rs
zhouwfang: I'm not sure about "Note that ...". Could you elaborate?

crates/interpreter/src/exec.rs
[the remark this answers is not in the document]
ia0: `side_table` should use the `Parser` defined by `module` instead

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,59 @@
...
+use core::slice;
+
+use bytemuck::cast;
+
 use crate::bit_field::*;
 use crate::error::*;
+use crate::parser;
+use crate::toctou::Use;
+
+pub type Parser<'m> = parser::Parser<'m, Use>;
```
ia0, line 24: Let's use the one from `module`.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,59 @@
...
+        self.0[0] as usize
+    }
+
+    pub fn parser(&self, code: &'m [u8]) -> Parser<'m> {
+        unsafe { Parser::new(&code[self.0[1] as usize .. self.0[5] as usize]) }
+    }
+
+    fn branch_table(&self) -> &[BranchTableEntry] {
+        let bytes = &self.0[5 ..];
+        unsafe {
```
ia0, line 57: We can use `bytemuck::cast_slice()` here.

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,59 @@
...
+struct Metadata<'m>(&'m [u16]);
+
+#[allow(dead_code)]
+impl<'m> Metadata<'m> {
+    pub fn type_idx(&self) -> usize {
+        self.0[0] as usize
+    }
+
+    pub fn parser(&self, code: &'m [u8]) -> Parser<'m> {
+        unsafe { Parser::new(&code[self.0[1] as usize .. self.0[5] as usize]) }
```
ia0, line 52: There's a few issues here:
- We need to read 2 u16 to get a u32.
- We need to read 1..3 and 3..5 (instead of 1 and 5)

We can probably use something like `bytemuck::pod_read_unaligned::<u32>(bytemuck::bytes_of(&self.0[1..3]))` to read a u32 from a &[u16] of length 2.

ia0 (commented):
Thanks! I think it looks good now, modulo the redoing of the BranchTableEntry representation (using [u16; 3] and no bitfield).

crates/interpreter/src/side_table.rs
ia0: If we have something like this in flash:
- The module is from 0x1000 to 0x2000.
- The code section is from 0x1800 to 0x1c00.
- The side table is from 0x2000 to 0x2400.
- The indices field is from 0x2000 to 0x2080.
- The metadata field is from 0x2080 to 0x2400.

If we read the 2 consecutive indices at 0x2012 and 0x2014, which have values A and B (both u16), then when indexing into the metadata field, we don't look at 0x2080 + A but at 0x2080 + 2A. This is done automatically by rust when indexing into a [u16]. So the function metadata for function with index 9 (which corresponds to offset 2x9 = 0x12 in the indices field) would be between 0x2080 + 2A and 0x2080 + 2B.

crates/interpreter/src/side_table.rs
zhouwfang: It requires the size of the array is known statically. Unlike `parser()`, we don't know the size here.

The error is: Trait `Pod` is not implemented for `[u16]`

zhouwfang:
I have learnt a lot from this PR. Thanks for the review!

crates/interpreter/src/side_table.rs
```diff
@@ -12,14 +12,61 @@
...
+struct Metadata<'m>(&'m [u16]);
+
+#[allow(dead_code)]
+impl<'m> Metadata<'m> {
+    pub fn type_idx(&self) -> usize {
+        self.0[0] as usize
+    }
+
+    pub fn parser(&self, code: &'m [u8]) -> Parser<'m> {
+        unsafe { Parser::new(&code[self.parser_pos(1, 3) .. self.parser_pos(3, 5)]) }
```
ia0, line 48: nit: The second argument of `parser_pos` is redundant. We can remove it.

crates/interpreter/src/side_table.rs
ia0: I got a different error. The problem was the destination `BranchTableEntry` that didn't claim to support any bit pattern.

crates/interpreter/src/side_table.rs
```diff
@@ -30,34 +77,29 @@ pub struct SideTableEntryView {
...
-    pub fn new(view: SideTableEntryView) -> Result<Self, Error> {
-        let mut fields = 0;
-        fields |= into_signed_field(Self::DELTA_IP_MASK, view.delta_ip)?;
-        fields |= into_signed_field(Self::DELTA_STP_MASK, view.delta_stp)?;
-        fields |= into_field(Self::VAL_CNT_MASK, view.val_cnt)?;
-        fields |= into_field(Self::POP_CNT_MASK, view.pop_cnt)?;
-        Ok(SideTableEntry(fields))
+impl BranchTableEntry {
+    pub fn new(view: BranchTableEntryView) -> Result<Self, Error> {
+        Ok(BranchTableEntry([
```
ia0, line 82: We should have debug assertions here (for the casts below).

ia0 (dismissed):
Thanks! LGTM modulo the nits. I'm creating a review commit and merging.

ia0 (dismissed):

ia0 (approved):

PR #713, opened 2025-01-02
Optimize side table masks

zhouwfang:
After experiments, these are the smallest masks to pass CI. In other words, `SideTableEntry` only requires `u35` at this point. I'll add the function index and type index in a follow-up PR, and `SideTableEntry` as `u64` may be sufficient to include them.

#46


ia0 (approved):
That's most probably too strict on the counts. I think we should eventually end up with:
- 16 bits delta IP
- 16 bits delta STP
- 4 bits val count
- 12 bits pop count

This gives 48 bits aligned on 16 bits.

PR #717, opened 2025-01-03
Add safety comments in the prelude

ia0:


crates/prelude/src/platform/update.rs
```diff
@@ -33,7 +33,7 @@ pub fn metadata() -> Result<Box<[u8]>, Error> {
     let mut len = 0;
     let params = api::metadata::Params { ptr: &mut ptr, len: &mut len };
     convert_unit(unsafe { api::metadata(params) })?;
-    let ptr = unsafe { core::slice::from_raw_parts_mut(ptr, len) };
+    let ptr = core::ptr::slice_from_raw_parts_mut(ptr, len);
     Ok(unsafe { Box::from_raw(ptr) })
```
ia0-review, line 37: This could use a safety comment too.

crates/prelude/src/serial.rs
```diff
@@ -138,13 +138,15 @@ pub fn flush<T: Serial>(serial: &T) -> Result<(), Error> {
...
+    notified: *const Cell<bool>,
 }
 
 impl<'a, T: Serial> Listener<'a, T> {
     /// Starts listening for the provided event until dropped.
     pub fn new(serial: &'a T, event: Event) -> Self {
-        let notified = Box::leak(Box::new(Cell::new(true)));
+        let notified = Box::into_raw(Box::new(Cell::new(true)));
         let func = Self::call;
         let data = notified as *mut _ as *const u8;
```
ia0-review, line 151: This could be simplified.

crates/prelude/src/serial.rs
```diff
@@ -153,10 +155,14 @@ impl<'a, T: Serial> Listener<'a, T> {
 
     /// Returns whether the event triggered since the last call.
     pub fn is_notified(&mut self) -> bool {
-        self.notified.replace(false)
+        // SAFETY: `self.notified` is a `Box<Cell<bool>> we share with the platform (see field
+        // invariant). We crate a shared borrow for the duration of this call.
```
ia0-review, line 159:
```suggestion
        // invariant). We create a shared borrow for the duration of this call.
```

ia0-review (changes requested):
Sounds like there's a bug when running the host runner tests.

ia0:
The applet is trapping because the allocator now returns a null pointer on zero size instead of allocating (which was wasteful but correct). Let's specify somewhere (`api-desc` documentation?) the common safety invariants regarding pointer-related functions (exported `alloc` and all imported functions mentioning pointers including function pointers).

ia0-review (approved):

ia0:
@not-an-aardvark Feel free to review if you got time (you can filter out the `Cargo.lock` files which are just noise, as well as the `Cargo.toml` and `CHANGELOG.md` files). The main file is `crates/api-desc/src/api.md` which describes the general invariant.

ia0-review (approved):

PR #718, opened 2025-01-07
Remove `Option` in side table creation

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -423,7 +423,7 @@ impl SideTable {
     }
 
     fn branch(&mut self) {
-        self.entries.push(SideTableEntry::default());
+        self.entries.push(SideTableEntry::invalid());
     }
 
     fn stitch(&mut self, source: SideTableBranch, target: SideTableBranch) -> CheckResult {
@@ -435,6 +435,7 @@ impl SideTable {
             unsupported(if_debug!(Unsupported::SideTable))
         })?;
         let pop_cnt = Self::pop_cnt(source, target)?;
+        debug_assert!(self.entries[source.side_table].is_invalid());
         self.entries[source.side_table] =
             SideTableEntry::new(SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt })?;
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -436,9 +435,8 @@ impl SideTable {
             unsupported(if_debug!(Unsupported::SideTable))
         })?;
         let pop_cnt = Self::pop_cnt(source, target)?;
-        let entry = &mut self.entries[source.side_table];
-        assert!(entry.is_none());
-        *entry = Some(SideTableEntryView { delta_ip, delta_stp, val_cnt, pop_cnt });
+        self.entries[source.side_table] =
```
ia0, line 438: Let's keep some debug assertions.
```suggestion
        debug_assert!(self.entries[source.side_table].is_invalid());
        self.entries[source.side_table] =
```

Where `fn is_invalid(self) -> bool` checks that the `u64` is zero. Also we should probably add `fn invalid() -> Self` that returns zero (instead of relying on Default, which we should remove if possible).

Zero is by luck a good sentinel value because it represents the maximum negative IP and STP jump.

crates/interpreter/src/valid.rs
```diff
@@ -478,7 +476,7 @@ impl SideTable {
     }
 
     fn persist(self) -> MResult<Vec<SideTableEntry>, Check> {
-        self.entries.into_iter().map(|entry| SideTableEntry::new(entry.unwrap())).collect()
+        Ok(self.entries)
```
ia0, line 479: Same here, we want to make sure we catch mistakes in debug mode.
```suggestion
        debug_assert!(self.entries.iter().all(|x| !x.is_invalid()));
        Ok(self.entries)
```

ia0 (commented):
Sounds good, but instead of removing the checks, I would move them to debug-mode only.

crates/interpreter/src/valid.rs
ia0: Most probably, but it depends how we implement the "streaming" validation. Validation could take some abstract `Persist` structure as argument, which behaves like a `Vec<u8>` to which we can read and write. This way we actually directly write to flash and there's no copy.

ia0 (approved):
Thanks!

PR #720, opened 2025-01-13
Return generalized side table from validation

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -30,7 +30,7 @@ pub struct Module<'m> {
     types: Vec<FuncType<'m>>,
     // TODO(dev/fast-interp): Flatten it to 1D array when making it persistent in
     // flash.
-    side_table: &'m [SideTableEntry],
+    side_table: &'m [MetadataEntry],
 }
 
 impl<'m> Import<'m> {
@@ -191,7 +191,7 @@ impl<'m> Module<'m> {
             let size = parser.parse_u32().into_ok() as usize;
             let parser = parser.split_at(size).into_ok();
             if i == x as usize {
-                return (parser, &self.side_table[i].metadata_entry.branch_table);
+                return (parser, &self.side_table[i].branch_table);
             }
         }
         unreachable!()
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -30,7 +30,7 @@ pub struct Module<'m> {
     types: Vec<FuncType<'m>>,
     // TODO(dev/fast-interp): Flatten it to 1D array when making it persistent in
     // flash.
-    side_tables: &'m [Vec<BranchTableEntry>],
+    side_table: &'m [SideTableEntry],
 }
 
 impl<'m> Import<'m> {
@@ -56,7 +56,7 @@ impl<'m> Module<'m> {
         let side_tables = validate(binary)?;
         let mut module = unsafe { Self::new_unchecked(binary) };
         // TODO(dev/fast-interp): We should take a buffer as argument to write to.
-        module.side_tables = Box::leak(Box::new(side_tables));
+        module.side_table = Box::leak(Box::new(side_tables));
         Ok(module)
     }
 
@@ -184,13 +184,14 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
+    // TODO(dev/fast-interp): Improve the performance of such accessor functions from O(n) to O(1).
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
         let mut parser = self.section(SectionId::Code).unwrap();
         for i in 0 .. parser.parse_vec().into_ok() {
             let size = parser.parse_u32().into_ok() as usize;
             let parser = parser.split_at(size).into_ok();
             if i == x as usize {
-                return (parser, &self.side_tables[i]);
+                return (parser, &self.side_table[i].metadata_entry.branch_table);
             }
         }
         unreachable!()
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -184,7 +184,6 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
-    // TODO(dev/fast-interp): Improve the performance of such accessor functions from O(n) to O(1).
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
         let mut parser = self.section(SectionId::Code).unwrap();
         for i in 0 .. parser.parse_vec().into_ok() {
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -28,8 +28,7 @@ use crate::*;
 pub struct Module<'m> {
     binary: &'m [u8],
     types: Vec<FuncType<'m>>,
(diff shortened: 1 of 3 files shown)


crates/interpreter/src/module.rs
```diff
@@ -184,7 +184,6 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
-    // TODO(dev/fast-interp): Improve the performance of such accessor functions from O(n) to O(1).
```
ia0, line 187: Why remove the TODO before fixing it? The `side_table` field of `Module` should actually be the "full side-table". The `func()` function would just read from it.

crates/interpreter/src/valid.rs
```diff
@@ -129,13 +129,24 @@ impl<'m> Context<'m> {
...
+            let saved = parser.save();
             for x in imported_funcs .. self.funcs.len() {
                 let size = parser.parse_u32()? as usize;
                 let mut parser = parser.split_at(size)?;
                 let t = self.functype(x as FuncIdx).unwrap();
                 let mut locals = t.params.to_vec();
                 parser.parse_locals(&mut locals)?;
-                side_tables.push(Expr::check_body(self, &mut parser, &refs, locals, t.results)?);
+                let parser_start =
+                    unsafe { parser.save().as_ptr().offset_from(saved.as_ptr()) as usize };
```
ia0, line 140: We don't need pointer arithmetic, we can just compute using `usize`, since we'll store the result to flash and need to use integers anyway.

crates/interpreter/src/valid.rs
```diff
@@ -412,31 +423,42 @@ struct Expr<'a, 'm> {
...
 
+#[allow(dead_code)]
+#[derive(Default)]
+struct ParserRange {
+    start: usize,
+    end: usize, // exclusive
+}
+
 #[derive(Default)]
 struct SideTable {
```
ia0, line 434: We should rename this to `FuncMeta` since we now generalized the definition of side table.

crates/interpreter/src/valid.rs
```diff
@@ -412,31 +423,42 @@ struct Expr<'a, 'm> {
     side_table: SideTable,
 }
 
+#[allow(dead_code)]
+#[derive(Default)]
+struct ParserRange {
```
ia0, line 428: Unless we want to implement methods on this, we can just use `Range<usize>` whenever we need a `ParserRange`. This is only used locally and the variable name should always clarify the meaning.

crates/interpreter/src/valid.rs
```diff
@@ -514,7 +536,8 @@ enum LabelKind<'m> {
 impl<'a, 'm> Expr<'a, 'm> {
     fn new(
         context: &'a Context<'m>, parser: &'a mut Parser<'m>,
-        is_const: Result<&'a mut [bool], &'a [bool]>,
+        is_const: Result<&'a mut [bool], &'a [bool]>, type_idx: Option<TypeIdx>,
+        parser_range: ParserRange,
```
ia0, line 540: It seems to me we don't need `type_idx` and `parser_range` for `Expr` to function properly. So instead of rename `SideTable` to `FuncMeta`, we just rename it to `BranchTable` with just the one field it used to have. Then in `check_module()`, when going through the code section, we get the output of `check_body()` to create a `FuncMeta` using the `type_idx` and `parser_range` information we have there.

ia0 (commented):
Thanks! The main issue is that I don't think `Expr` should care about `FuncMeta` (type index and parser range). We should build a `FuncMeta` from the `BranchTable` returned by `check_body()` while validating the code section (where we are within the context of a function, so `FuncMeta` makes sense).

crates/interpreter/src/module.rs
zhouwfang: Just because I have started working on it, and I'm sure it will be done.

crates/interpreter/src/valid.rs
```diff
@@ -476,16 +483,16 @@ impl SideTable {
         })
     }
 
-    fn persist(self) -> MResult<Vec<BranchTableEntry>, Check> {
```
zhouwfang, line 479: The reason why I implemented in the previous way is that by calling it "persist", I thought you wanted to write to flash here.

crates/interpreter/src/side_table.rs
```diff
@@ -98,3 +101,17 @@ impl BranchTableEntry {
...
+    #[allow(dead_code)]
+    pub type_idx: usize,
+    pub metadata_entry: MetadataEntry,
+}
+
+#[derive(Default, Debug)]
+pub struct MetadataEntry {
+    #[allow(dead_code)]
+    pub parser_range: Range<usize>,
+    pub branch_table: Vec<BranchTableEntry>,
```
zhouwfang, line 116: These are named based on the `SideTable` and `Metadata` in this file.

crates/interpreter/src/valid.rs
```diff
@@ -129,13 +128,21 @@ impl<'m> Context<'m> {
         let mut side_tables = vec![];
         if let Some(mut parser) = self.check_section(parser, SectionId::Code)? {
             check(self.funcs.len() == imported_funcs + parser.parse_vec()?)?;
+            let mut offset = 0;
             for x in imported_funcs .. self.funcs.len() {
                 let size = parser.parse_u32()? as usize;
+                offset += size_of::<u32>();
                 let mut parser = parser.split_at(size)?;
-                let t = self.functype(x as FuncIdx).unwrap();
+                let t = self.functype(x as FuncIdx)?;
```
ia0, line 136: This call can't fail which is why it's using `unwrap()`. We check `funcs.len()` above and we also have a check against `types.len()` when pushing to `funcs`.

crates/interpreter/src/valid.rs
ia0: No, I was using the word `persist` as in `serialize`. There's 3 options for when to write:
1. At the very end when we have the full side table.
2. During validation without allocation.
3. The module comes with a side table already, and in validation we only validate it.

I guess the 2nd or 3rd option would be the best, but that's for later.

ia0 (commented):
Thanks! LGTM except for the `offset` part.

crates/interpreter/src/valid.rs
```diff
@@ -129,13 +128,21 @@ impl<'m> Context<'m> {
         let mut side_tables = vec![];
         if let Some(mut parser) = self.check_section(parser, SectionId::Code)? {
             check(self.funcs.len() == imported_funcs + parser.parse_vec()?)?;
+            let mut offset = 0;
             for x in imported_funcs .. self.funcs.len() {
                 let size = parser.parse_u32()? as usize;
+                offset += size_of::<u32>();
```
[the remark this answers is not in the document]
zhouwfang: No, I didn't get a lint, but I agree with you and did misunderstand it.

crates/interpreter/src/valid.rs
zhouwfang: In 2), what do you mean by "without allocation"?

In 3), do you mean preallocate space for side table in module before validation?

zhouwfang:
Could you take a look at [the follow-up PR](https://github.com/zhouwfang/wasefire/pull/7)? Thanks.

ia0 (approved):

ia0:
I'm merging `main` into `dev/fast-interp` after this commit. So when pulling, you'll get new submodules and new rust toolchain. You just need to run `git submodules update`. The toolchain is automatic with rustup.

PR #723, opened 2025-01-16
Improve the performance of function accessors

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -183,7 +183,6 @@ impl<'m> Module<'m> {
         unreachable!()
     }
 
-    // TODO(dev/fast-interp): Improve the performance of such accessor functions from O(n) to O(1).
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
         let mut parser = self.section(SectionId::Code).unwrap();
         for i in 0 .. parser.parse_vec().into_ok() {
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -116,14 +116,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func_type(&self, x: FuncIdx) -> FuncType<'m> {
-        let mut parser = self.section(SectionId::Function).unwrap();
-        for i in 0 .. parser.parse_vec().into_ok() {
-            let y = parser.parse_typeidx().into_ok();
-            if i == x as usize {
-                return self.types[y as usize];
-            }
-        }
-        unreachable!()
+        self.types[self.side_table[x as usize].type_idx]
     }
 
     pub(crate) fn table_type(&self, x: TableIdx) -> TableType {
@@ -185,14 +178,13 @@ impl<'m> Module<'m> {
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
         let mut parser = self.section(SectionId::Code).unwrap();
-        for i in 0 .. parser.parse_vec().into_ok() {
-            let size = parser.parse_u32().into_ok() as usize;
-            let parser = parser.split_at(size).into_ok();
-            if i == x as usize {
-                return (parser, &self.side_table[i].branch_table);
-            }
+        let saved = parser.save();
+        let entry = &self.side_table[x as usize];
+        let range = &entry.parser_range;
+        unsafe {
+            parser.restore(&saved[range.start .. range.end]);
         }
-        unreachable!()
+        (parser, &entry.branch_table)
     }
 
     pub(crate) fn data(&self, x: DataIdx) -> Parser<'m> {
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -177,12 +177,11 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let mut parser = self.section(SectionId::Code).unwrap();
-        let saved = parser.save();
+        let mut parser = unsafe { Parser::new(self.binary) };
         let entry = &self.side_table[x as usize];
         let range = &entry.parser_range;
         unsafe {
-            parser.restore(&saved[range.start .. range.end]);
+            parser.restore(&parser.save()[range.start .. range.end]);
         }
         (parser, &entry.branch_table)
     }
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -58,9 +58,7 @@ impl<'m> Metadata<'m> {
 
 #[derive(Default, Debug)]
 pub struct MetadataEntry {
-    #[allow(dead_code)]
     pub type_idx: usize,
-    #[allow(dead_code)]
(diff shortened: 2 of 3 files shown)


crates/interpreter/src/side_table.rs
```diff
@@ -43,24 +43,28 @@ impl<'m> Metadata<'m> {
...
         bytemuck::pod_read_unaligned::<u32>(bytemuck::cast_slice(&self.0[idx .. idx + 2])) as usize
     }
 }
 
 #[derive(Default, Debug)]
 pub struct MetadataEntry {
-    #[allow(dead_code)]
     pub type_idx: usize,
     #[allow(dead_code)]
+    pub code_start: usize,
```
zhouwfang, line 67: It will be useful to get the argument for the `Metadata::parser()` above.

crates/interpreter/src/valid.rs
```diff
@@ -127,19 +128,20 @@ impl<'m> Context<'m> {
         }
         let mut side_tables = vec![];
         if let Some(mut parser) = self.check_section(parser, SectionId::Code)? {
+            let code_start = parser.save().as_ptr() as usize - parser_origin;
             check(self.funcs.len() == imported_funcs + parser.parse_vec()?)?;
-            let code_start = parser.save().as_ptr() as usize;
```
zhouwfang, line 131: The reason why I moved the `code_start` before `parser.parse_vec()` is to be consistent with the `Module::func()` in `main` [where `parse.parse_vec()` is called right after the code section starts](https://github.com/zhouwfang/wasefire/blob/4c7ce53febeae4510152b8b81e79509fb11919dc/crates/interpreter/src/module.rs#L190).

crates/interpreter/src/valid.rs
```diff
@@ -127,19 +128,20 @@ impl<'m> Context<'m> {
...
             for x in imported_funcs .. self.funcs.len() {
                 let size = parser.parse_u32()? as usize;
                 let mut parser = parser.split_at(size)?;
+                let parser_start = parser.save().as_ptr() as usize - parser_origin;
```
zhouwfang, line 136: The reasons why I use `parser_origin` instead of `code_start` to compute `parser_start` is that `Context::check_section()` and `Module::section` seem slightly different because[ there is an extra `parse_name()` in the former](https://github.com/zhouwfang/wasefire/blob/4c7ce53febeae4510152b8b81e79509fb11919dc/crates/interpreter/src/valid.rs#L174). I don't understand why it is missing in the latter. So I couldn't trust `code_start`.

crates/interpreter/src/valid.rs
```diff
@@ -127,19 +128,20 @@ impl<'m> Context<'m> {
...
+                let parser_end = parser_start + size;
                 let t = self.functype(x as FuncIdx).unwrap();
                 let mut locals = t.params.to_vec();
                 parser.parse_locals(&mut locals)?;
-                let parser_start = parser.save().as_ptr() as usize - code_start;
                 let branch_table = Expr::check_body(self, &mut parser, &refs, locals, t.results)?;
-                let parser_end = parser.save().as_ptr() as usize - code_start;
```
zhouwfang, line 140: The reason why I moved `parser_start` and `parser_end` before `parse_locals()` is that [in execution, after `Module::func()` is called, `append_locals()` will be called](https://github.com/zhouwfang/wasefire/blob/4c7ce53febeae4510152b8b81e79509fb11919dc/crates/interpreter/src/exec.rs#L225-L228), which is similar to `parse_locals()` here.

crates/interpreter/src/module.rs
```diff
@@ -183,17 +176,14 @@ impl<'m> Module<'m> {
...
-                return (parser, &self.side_table[i].branch_table);
-            }
+        let mut parser = unsafe { Parser::new(self.binary) };
+        let entry = &self.side_table[x as usize];
+        let range = &entry.parser_range;
+        unsafe {
+            parser.restore(&parser.save()[range.start .. range.end]);
         }
-        unreachable!()
+        (parser, &entry.branch_table)
```
ia0, line 186: This function is quite convoluted. You can create the parser directly with `Parser::new(&self.binary[entry.range])`.

crates/interpreter/src/side_table.rs
```diff
@@ -43,24 +43,28 @@ impl<'m> Metadata<'m> {
         self.0[0] as usize
     }
 
+    pub fn code_start(&self) -> usize {
```
ia0, line 46: This shouldn't be store in each function, aka `Metadata`, it should be in `SideTable` (before `indices`).

crates/interpreter/src/valid.rs
```diff
@@ -46,6 +46,7 @@ struct Context<'m> {
 
 impl<'m> Context<'m> {
     fn check_module(&mut self, parser: &mut Parser<'m>) -> MResult<Vec<MetadataEntry>, Check> {
+        let parser_origin = parser.save().as_ptr() as usize;
```
ia0, line 49: nit: `module_start` for coherence with the other variables

crates/interpreter/src/valid.rs
ia0: The link you provide is for custom sections. We don't care about those (actually, I was thinking of having the side table as a custom section eventually, but this doesn't matter).

You should be able to use `code_start` instead of `module_start`. However, it's true that when saving `code_start` in the side table, we need to store the offset from the module start, not the pointer we use to compute parser_start and end.

ia0 (commented):
Thanks! Looks generally good, the only big problem is that code start should be store in the side table, not in each metadata.

crates/interpreter/src/module.rs
zhouwfang: It does not work: "Trait `SliceIndex<[u8]>` is not implemented for `&Range<usize>"

zhouwfang:
@ia0 Could you help me get [the metrics](https://github.com/google/wasefire/issues/46#issuecomment-2564087368) after this PR on nordic? Thanks.

crates/interpreter/src/valid.rs
```diff
@@ -47,6 +47,7 @@ struct Context<'m> {
 impl<'m> Context<'m> {
     fn check_module(&mut self, parser: &mut Parser<'m>) -> MResult<Vec<MetadataEntry>, Check> {
         check(parser.parse_bytes(8)? == b"\0asm\x01\0\0\0")?;
+        let module_start = parser.save().as_ptr() as usize;
```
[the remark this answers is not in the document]
ia0: Yes, this is important.

ia0 (approved):
Thanks!

crates/interpreter/src/module.rs
ia0: It had to be cloned.

zhouwfang:
> @ia0 Could you help me get [the metrics](https://github.com/google/wasefire/issues/46#issuecomment-2564087368) after this PR on nordic? Thanks.

@ia0 When you get some time, could you take a look? I'm still working on making the nordic DK work with my laptop.

ia0:
Oops I missed that comment.

Before:
```
Peak RAM usage: 5668
CoreMark result: 0.15474127 (in 129.514s)
```

After:
```
Peak RAM usage: 5668
Peak RAM usage: 5684
CoreMark result: 0.15614265 (in 128.356s)
```

The peak RAM was 5668 during the whole test but only increased at the end. It's anyway rather minor.

PR #725, opened 2025-01-21
Add empty `ValidMode` and `Prepare`

zhouwfang:
The next step is to add `Verify` with optimizations.

#46

### DIFF
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -20,7 +20,7 @@ use crate::parser::{SkipData, SkipElem};
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
-use crate::valid::{Prepare, validate};
+use crate::valid::validate;
 use crate::*;
 
 /// Valid module.
@@ -52,7 +52,7 @@ impl ImportDesc {
 impl<'m> Module<'m> {
     /// Validates a WASM module in binary format.
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
-        let side_table = validate::<Prepare>(binary)?;
+        let side_table = validate(binary)?;
         let mut module = unsafe { Self::new_unchecked(binary) };
         // TODO(dev/fast-interp): We should take a buffer as argument to write to.
         module.side_table = Box::leak(Box::new(side_table));
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,7 +20,7 @@ use crate::parser::{SkipData, SkipElem};
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
-use crate::valid::validate;
+use crate::valid::{validate, Prepare};
 use crate::*;
 
 /// Valid module.
@@ -52,7 +52,7 @@ impl ImportDesc {
 impl<'m> Module<'m> {
     /// Validates a WASM module in binary format.
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
-        let side_table = validate(binary)?;
+        let side_table = validate::<Prepare>(binary)?;
         let mut module = unsafe { Self::new_unchecked(binary) };
         // TODO(dev/fast-interp): We should take a buffer as argument to write to.
         module.side_table = Box::leak(Box::new(side_table));
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -20,7 +20,7 @@ use crate::parser::{SkipData, SkipElem};
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
-use crate::valid::{validate, Prepare};
+use crate::valid::{Prepare, validate};
 use crate::*;
 
 /// Valid module.
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -27,7 +27,7 @@ use crate::util::*;
 use crate::*;
 
 /// Checks whether a WASM module in binary format is valid.
-pub fn validate<M: ValidMode>(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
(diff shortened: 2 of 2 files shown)


zhouwfang:
Not sure how this PR leads to "thread `binary_leb128` has overflowed its stack".

crates/interpreter/src/module.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2025 Google LLC
```
ia0, line 1: We don't need to update the copyright date.

crates/interpreter/src/valid.rs
```diff
@@ -25,23 +25,47 @@ use crate::toctou::*;
 use crate::util::*;
 use crate::*;
 
+pub trait ValidMode {}
+
+pub struct Prepare;
+impl ValidMode for Prepare {}
+
+impl Default for Prepare {
```
ia0, line 33: We should `#[derive(Default)]` on `struct Prepare`.

crates/interpreter/src/valid.rs
```diff
@@ -25,23 +25,47 @@ use crate::toctou::*;
 use crate::util::*;
 use crate::*;
 
+pub trait ValidMode {}
```
ia0, line 28: We can have `trait ValidMode: Default {}` for convenience, then we can just write `ValidMode` instead of `ValidMode + Default`.

crates/interpreter/src/valid.rs
```diff
@@ -25,23 +25,47 @@ use crate::toctou::*;
...
+pub fn validate<M: ValidMode + Default>(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
     Context::default().check_module(&mut Parser::new(binary))
 }
 
 type Parser<'m> = parser::Parser<'m, Check>;
 type CheckResult = MResult<(), Check>;
 
-#[derive(Default)]
-struct Context<'m> {
+struct Context<'m, M: ValidMode = Prepare> {
```
ia0, line 47: Do we need the default? Can't we just have `M: ValidMode`? Ideally we should always be explicit.

crates/interpreter/src/valid.rs
```diff
@@ -25,23 +25,47 @@ use crate::toctou::*;
...
 /// Checks whether a WASM module in binary format is valid.
-pub fn validate(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
+pub fn validate<M: ValidMode + Default>(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
     Context::default().check_module(&mut Parser::new(binary))
 }
 
 type Parser<'m> = parser::Parser<'m, Check>;
 type CheckResult = MResult<(), Check>;
 
-#[derive(Default)]
```
ia0, line 36: Why remove the derive?

crates/interpreter/src/valid.rs
```diff
@@ -25,23 +25,47 @@ use crate::toctou::*;
...
+struct Context<'m, M: ValidMode = Prepare> {
     types: Vec<FuncType<'m>>,
     funcs: Vec<TypeIdx>,
     tables: Vec<TableType>,
     mems: Vec<MemType>,
     globals: Vec<GlobalType>,
     elems: Vec<RefType>,
     datas: Option<usize>,
+    #[allow(dead_code)]
+    mode: M,
```
ia0, line 56: This should be `PhantomData<M>` to make it clear we only care about the type. I hope we don't need to store anything, and if we do, then I would prefer to use associated types (for example for `Label::branches`).

ia0 (commented):
The stack overflow is because `Context::new()` and `Context::default()` are mutually recursive. By using `derive` you would fix this.

crates/interpreter/src/valid.rs
zhouwfang: What do you mean by "associated types"?

crates/interpreter/src/valid.rs
```diff
@@ -322,19 +333,21 @@ impl<'m> parser::ParseElem<'m, Check> for ParseElem<'_, 'm> {
...
-    context: &'a mut Context<'m>,
+struct ParseData<'a, 'm, M: ValidMode> {
+    context: &'a mut Context<'m, M>,
     refs: &'a mut [bool],
     num_global_imports: usize,
 }
 
-impl<'a, 'm> ParseData<'a, 'm> {
-    fn new(context: &'a mut Context<'m>, refs: &'a mut [bool], num_global_imports: usize) -> Self {
+impl<'a, 'm, M: ValidMode> ParseData<'a, 'm, M> {
```
zhouwfang, line 342: Rust question: Is there a good reason why the `<'a, 'm, M>` after `ParseData` couldn't be omitted? It seems redundant to write them since they could be inferred from the `impl<'a, 'm, M: ValidMode>`.

crates/interpreter/src/valid.rs
```diff
@@ -26,25 +27,33 @@ use crate::util::*;
 use crate::*;
 
 /// Checks whether a WASM module in binary format is valid.
-pub fn validate(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
-    Context::default().check_module(&mut Parser::new(binary))
+pub fn validate<M: ValidMode>(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
+    Context::<Prepare>::default().check_module(&mut Parser::new(binary))
```
ia0, line 31: I guess it's better to do the design first, like on #726. For this PR, since there's only `Prepare`, I would suggest not making `validate` generic. Only `Context` and its methods need to be. We would then have 2 public functions (and `ValidMode` can stay private):

```rust
pub fn prepare(binary: &[u8]) -> Result<Vec<u8>, Error>; // returns the sidetable section
pub fn validate(binary: &[u8]) -> Result<(), Error>; // the binary has a sidetable section in this case
```

We could even have:

```rust
pub fn prepare(binary: &mut Vec<u8>) -> Result<(), Error>; // appends the sidetable section
```

crates/interpreter/src/valid.rs
ia0: It's because those are different things. `ParseData` can take arbitrary parameters. If there would be a syntactic sugar it would be the opposite. You have to specify the `ParseData` parameters, and any free variable would be universally quantified on the `impl`. But I'm not sure this would help readability.

ia0 (approved):

crates/interpreter/src/valid.rs
ia0: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#specifying-placeholder-types-in-trait-definitions-with-associated-types

crates/interpreter/src/valid.rs
zhouwfang: Could you give an example where the struct has more parameters than `impl`? Thanks.

ia0: [Here](https://doc.rust-lang.org/nightly/std/pin/struct.Pin.html#impl-Pin%3C%26mut+T%3E)'s an example from the standard library.

zhouwfang: Thanks! "universally quantified on the `impl`" is a very good description.

PR #726, opened 2025-01-22
Add the `Verify` mode

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/parser.rs
+++ b/crates/interpreter/src/parser.rs
@@ -98,10 +98,6 @@ impl<'m, M: Mode> Parser<'m, M> {
         Ok(val)
     }
 
-    pub fn parse_u16(&mut self) -> MResult<u16, M> {
-        Ok(self.parse_leb128(false, 16)? as u16)
-    }
-
     pub fn parse_u32(&mut self) -> MResult<u32, M> {
         Ok(self.parse_leb128(false, 32)? as u32)
     }
--- a/crates/interpreter/src/parser.rs
+++ b/crates/interpreter/src/parser.rs
@@ -98,6 +98,10 @@ impl<'m, M: Mode> Parser<'m, M> {
         Ok(val)
     }
 
+    pub fn parse_u16(&mut self) -> MResult<u16, M> {
+        Ok(self.parse_leb128(false, 16)? as u16)
+    }
+
     pub fn parse_u32(&mut self) -> MResult<u32, M> {
         Ok(self.parse_leb128(false, 32)? as u32)
     }
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -19,13 +19,13 @@ use crate::error::*;
 use crate::module::Parser;
 
 #[allow(dead_code)]
-pub struct SideTable<'m> {
+pub struct SideTableView<'m> {
     indices: &'m [u16], // including 0 and the length of metadata_array
     metadata: &'m [u16],
 }
 
 #[allow(dead_code)]
-impl<'m> SideTable<'m> {
+impl<'m> SideTableView<'m> {
     fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
             &self.metadata[self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -22,14 +22,35 @@ pub struct SideTableView<'m> {
     pub func_idx: usize,
     pub indices: &'m [u16], // including 0 and the length of metadata_array
     pub metadata: &'m [u16],
+    pub branch_table_view: Metadata<'m>,
 }
 
 impl<'m> SideTableView<'m> {
+    pub fn new(parser: &mut crate::valid::Parser<'m>) -> Result<Self, Error> {
+        Ok(SideTableView {
+            func_idx: 0,
+            indices: parse_side_table_field(parser)?,
+            metadata: parse_side_table_field(parser)?,
+            branch_table_view: Default::default(),
+        })
+    }
+
     pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
             &self.metadata[self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
         )
     }
+}
+
+fn parse_u16(data: &[u8]) -> u16 {
+    bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&data[0 .. 2]))
+}
+
+fn parse_side_table_field<'m>(parser: &mut crate::valid::Parser<'m>) -> Result<&'m [u16], Error> {
+    let len = parse_u16(parser.save()) as usize;
+    let parser = parser.split_at(len)?;
+    let bytes = parser.save().get(0 .. len * 2).unwrap();
+    Ok(bytemuck::cast_slice::<_, u16>(bytes))
 }
 
 #[derive(Default, Copy, Clone)]
@@ -42,7 +63,7 @@ impl<'m> Metadata<'m> {
 
     #[allow(dead_code)]
     pub fn parser(&self, module: &'m [u8]) -> Parser<'m> {
-        unsafe { Parser::new(&module[self.parser_range().start .. self.parser_range().end]) }
+        unsafe { Parser::new(&module[self.parser_range()]) }
     }
 
     pub fn branch_table(&self) -> &[BranchTableEntry] {
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -33,9 +33,8 @@ impl<'m> SideTableView<'m> {
     }
 }
 
-#[allow(dead_code)]
-#[derive(Copy, Clone)]
-struct Metadata<'m>(&'m [u16]);
+#[derive(Default, Copy, Clone)]
+pub struct Metadata<'m>(&'m [u16]);
 
 #[allow(dead_code)]
 impl<'m> Metadata<'m> {
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -18,15 +18,14 @@ use core::ops::Range;
 use crate::error::*;
 use crate::module::Parser;
 
-#[allow(dead_code)]
 pub struct SideTableView<'m> {
-    indices: &'m [u16], // including 0 and the length of metadata_array
-    metadata: &'m [u16],
+    pub func_idx: usize,
+    pub indices: &'m [u16], // including 0 and the length of metadata_array
+    pub metadata: &'m [u16],
 }
 
-#[allow(dead_code)]
 impl<'m> SideTableView<'m> {
-    fn metadata(&self, func_idx: usize) -> Metadata<'m> {
+    pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
             &self.metadata[self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
         )
@@ -36,18 +35,22 @@ impl<'m> SideTableView<'m> {
 #[derive(Default, Copy, Clone)]
 pub struct Metadata<'m>(&'m [u16]);
 
-#[allow(dead_code)]
 impl<'m> Metadata<'m> {
     pub fn type_idx(&self) -> usize {
         self.0[0] as usize
     }
 
+    #[allow(dead_code)]
     pub fn parser(&self, module: &'m [u8]) -> Parser<'m> {
-        unsafe { Parser::new(&module[self.read_u32(1) .. self.read_u32(3)]) }
+        unsafe { Parser::new(&module[self.parser_range().start .. self.parser_range().end]) }
     }
 
     pub fn branch_table(&self) -> &[BranchTableEntry] {
         bytemuck::cast_slice(&self.0[5 ..])
+    }
+
+    pub fn parser_range(&self) -> Range<usize> {
+        self.read_u32(1) .. self.read_u32(3)
     }
 
     fn read_u32(&self, idx: usize) -> usize {
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -35,7 +35,7 @@ pub fn prepare(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
 #[allow(unused_variables)]
 // Checks whether a WASM module with the side table in binary format is valid.
 pub fn verify(binary: &[u8]) -> Result<(), Error> {
-    Ok(())
+    todo!()
 }
 
 trait ValidMode: Default {
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -27,7 +27,7 @@ use crate::util::*;
 use crate::*;
 
 /// Checks whether a WASM module in binary format is valid.
-pub fn validate(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
+pub fn prepare(binary: &[u8]) -> Result<Vec<MetadataEntry>, Error> {
     Context::<Prepare>::default().check_module(&mut Parser::new(binary))
 }
 
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -19,7 +19,6 @@ use core::cmp::Ordering;
 use core::marker::PhantomData;
 use core::ops::Range;
 
-use crate::Error::Invalid;
 use crate::error::*;
 use crate::side_table::*;
 use crate::syntax::*;
@@ -79,6 +78,7 @@ impl ValidMode for Prepare {
     type Branches<'m> = Vec<SideTableBranch<'m>>;
     type BranchTable<'m> = Vec<BranchTableEntry>;
     type SideTable<'m> = Vec<MetadataEntry>;
+    // TODO(dev/fast-interp): Change it to Vec<u8>.
     type Result = Vec<MetadataEntry>;
 
     fn parse_side_table<'m>(_: &mut Parser<'m>) -> Result<Self::SideTable<'m>, Error> {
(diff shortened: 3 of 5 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -502,6 +502,7 @@ struct Label<'m> {
     /// Whether the bottom of the stack is polymorphic.
     polymorphic: bool,
     stack: Vec<OpdType>,
+    // In the Verify mode, below becomes `branch: Option<SideTableBranch<'m>>` (optimizing RAM)
```
ia0, line 505: Correct.

crates/interpreter/src/valid.rs
```diff
@@ -861,6 +862,8 @@ impl<'a, 'm> Expr<'a, 'm> {
     fn end_label(&mut self) -> CheckResult {
         let results_len = self.label().type_.results.len();
         let mut target = self.branch_target(results_len);
+        // Instead of calling stitch(), make sure the side table entry is consistent with the one in
+        // flash.
```
ia0, line 866: Almost. We need to check that `self.label().branch.is_none_or(|x| x == target)`.

crates/interpreter/src/valid.rs
```diff
@@ -885,7 +888,12 @@ impl<'a, 'm> Expr<'a, 'm> {
         let n = self.labels.len();
         check(l < n)?;
         let source = self.branch_source();
+        // Apply the side table entry in flash and compute the expected target branch.
```
ia0, line 891: Yes, but we shouldn't use the word flash. We're reading the side table from the side table section the input binary slice.

crates/interpreter/src/valid.rs
```diff
@@ -885,7 +888,12 @@ impl<'a, 'm> Expr<'a, 'm> {
...
+        // source.parser += delta_ip;
+        // source.branch_table += delta_stp;
         let label = &mut self.labels[n - l - 1];
+        // if label.branch.is_none() { label.branch = source; }
+        // else { compare label.branch.unwrap() with source to make sure they are the same }
```
ia0, line 896: Yes on the idea, but the Rust code should use a match, or something like `label.branch.replace(source).is_none_or(|x| x == source)`.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,137 @@ use crate::toctou::*;
...
+
+    #[allow(dead_code)]
+    fn br_label<'m>(expr: &mut Expr<'_, 'm, Self>, l: LabelIdx) -> Result<ResultType<'m>, Error> {
+        let l = l as usize;
+        let n = expr.labels.len();
+        check(l < n)?;
+        let source = expr.branch_source();
+        let expected_target = source;
+        // expected_target.parser += delta_ip;
+        // expected_target.branch_table += delta_stp;
```
zhouwfang, line 132: The remaining question is how to format and read the side table from the custom section of the WASM module.

In terms of the format, how about using [the format](https://github.com/zhouwfang/wasefire/blob/8017774e8c1a5c7b5054c9d42f3006e6b607a793/crates/interpreter/src/side_table.rs#L22-L25) that was intended for flash storage? 

It may not be efficient to keep the whole side table in RAM. How about keeping a parser for it and reading the data on demand?

zhouwfang: More precisely, 
```
[Section ID: 0] 
[Section Size: ...]
[Section Name Length: 18]
[Section Name: "Wasefire-SideTable"] 
[Indices Array Size (LEB128): num_functions + 1] 
[Indices Array: (u16 * (num_functions + 1))]
[Metadata Array Size (LEB128): ...]
[Metadata Array: (u16 * ...)]
```

ia0: > In terms of the format, how about using [the format](https://github.com/zhouwfang/wasefire/blob/8017774e8c1a5c7b5054c9d42f3006e6b607a793/crates/interpreter/src/side_table.rs#L22-L25) that was intended for flash storage?

Yes, that's the idea, even if we didn't have a proper parsing function yet. That's the format I wrote [there](https://github.com/google/wasefire/pull/711#pullrequestreview-2528876028) (updated below):

```
side-table := index flatten-metadata (16-bits aligned)
index := u16 u16 ... u16 (N + 1 times where N is the number of functions in the module)
flatten-metadata := metadata metadata ... metadata (N times where `flatten-metadata[i]` starts at byte position `index[i]` and ends at `index[i + 1]`)
metadata := type:u16 parser:(u32 u32) branch-table
branch-table := branch branch ... branch (length is implicitly defined by `index[i + 1]` see above)
branch := delta-ip:u16 delta-stp:u16 val-cnt:u8 pop-cnt:u12
```

In particular, we shouldn't use LEB128 since `u16` is enough and the benefit from using only one byte up to 127 is not worth it.

ia0 (commented):
I've pushed a commit to better factorize `end_label` and `br_label`. It's not finished because similarly to the `ValidMode::Branches`, we also need `ValidMode::BranchTable`, which is either prepared or verified. In particular it must be read from `ValidMode::patch_branch()` for `Verify`.

crates/interpreter/src/valid.rs
```diff
@@ -427,8 +512,8 @@ struct Expr<'a, 'm, M: ValidMode> {
     is_const: Result<&'a mut [bool], &'a [bool]>,
     is_body: bool,
     locals: Vec<ValType>,
-    labels: Vec<Label<'m>>,
-    branch_table: BranchTable,
+    labels: Vec<Label<'m, M>>,
+    branch_table: M::BranchTable<'m>,
```
zhouwfang, line 516: Due to this change, the compiler could not find the existing methods in `impl BranchTable` (used for `Prepare`) such as `persist()`, `branch()`, `save()`, etc. What is an idiomatic way to link these methods to the concrete associated type? It seems doable to add them to the `ValidMode` trait, but it does not "smell right".

ia0: Yes, to be clean we would need something like this:

```rust
trait ValidMode: Default {
    type Branches<'m>: BranchesApi<'m>;
    type BranchTable<'m>: BranchTableApi<'m>;
    // no methods here
}

trait BranchesApi<'m>: Default + IntoIterator<Item = SideTableBranch<'m>> {
    fn push(&mut self, branch: SideTableBranch<'m>) -> CheckResult;
}

trait BranchTableApi<'m> {
    fn stitch(&mut self, source: SideTableBranch<'m>, target: SideTableBranch<'m>) -> CheckResult;
    fn patch(&self, source: SideTableBranch<'m>) -> Result<SideTableBranch<'m>, Error>;
    fn persist();
    // and other functions you need
}
```

zhouwfang: Could you look at the changes to implement the above? I'll continue to work on the `Verify` mode.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+    /// When verifying, contains at most one _target_ branch. Source branches are eagerly patched to
+    /// their target branch using the branch table.
+    type Branches<'m>: BranchesApi<'m>;
+
+    /// Branch table to prepare or verify.
+    type BranchTable<'m>: BranchTableApi<'m>;
+}
+
+trait BranchesApi<'m>: Default + IntoIterator<Item = SideTableBranch<'m>> {
+    // fn push(&mut self, branch: SideTableBranch<'m>) -> CheckResult;
```
ia0, line 53: I guess this didn't work because there would be a conflict when trying to resolve the method name?

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+    ///
+    /// When verifying, only push if there are no branches already. If there is one, verify that
+    /// it's the same.
+    fn push_branch(&mut self, branch: SideTableBranch<'m>) -> CheckResult;
+}
+
+trait BranchTableApi<'m>: Default {
+    /// Updates the branch table for source according to target, when preparing.
+    ///
+    /// When verifying, makes sure source is target branch.
```
ia0, line 64: Actually to simplify review, and probably also your work, we could simply have those documentation as TODOs in the respective implementations. This way it's local. And we don't need documentation for those private traits, we can directly look at the implementation which will usually be shorter than the documentation.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+    fn stitch_branch(
+        &mut self, source: SideTableBranch<'m>, target: SideTableBranch<'m>,
+    ) -> CheckResult;
+
+    /// Does nothing, when preparing.
+    ///
+    /// When verifying, patches a source branch to its target branch using the branch table.
+    fn patch_branch(&self, source: SideTableBranch<'m>) -> Result<SideTableBranch<'m>, Error>;
+
+    fn persist(self) -> MResult<Vec<BranchTableEntry>, Check>;
```
ia0, line 74: As discussed, we probably need a different API here. This should be part of the `SideTableApi` by accessing the next `BranchTable` that we would pass as argument to `check_body`. So we won't need to call `persist()` anymore.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+
+    fn branch(&mut self);
```
ia0, line 76: Let's call this `allocate_branch()`.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+
+    fn save(&self) -> usize;
```
ia0, line 78: And let's call this either `next_branch()` or `next_index()`.

crates/interpreter/src/valid.rs
zhouwfang: Actually it is fine. Used `push` instead.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,135 @@ use crate::toctou::*;
...
+
+trait ValidMode: Default {
+    /// List of source branches, when preparing.
+    ///
+    /// When verifying, contains at most one _target_ branch. Source branches are eagerly patched to
+    /// their target branch using the branch table.
+    type Branches<'m>: BranchesApi<'m>;
+
+    /// Branch table to prepare or verify.
+    type BranchTable<'m>: BranchTableApi<'m>;
```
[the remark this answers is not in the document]
zhouwfang: I'm not sure this associated type should be in `ValidMode`. See the question below.

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,151 @@ use crate::toctou::*;
...
+
+    fn persist(self) -> MResult<Vec<BranchTableEntry>, Check>;
+
+    fn allocate_branch(&mut self);
+
+    fn next_index(&self) -> usize;
+}
+
+trait SideTableApi<'m> {
+    fn parse_side_table() -> ValidMode::SideTable<'m>;
```
zhouwfang, line 68: The compiler complains here: `ambiguous associated type`. We want to return a side table that is an empty vector for `Prepare` and a `SideTableView` for `Verify`. Should the return type be an associated type of `SideTableApi` instead?

crates/interpreter/src/valid.rs
```diff
@@ -26,16 +26,151 @@ use crate::toctou::*;
...
+
+    // For Verify, need to check if the inputs are consistent with the ones in flash.
+    // NextBranchTable is a new associated type in ValidMode that is a &BranchTable for Verify and a
+    // BranchTable for Prepare.
```
zhouwfang, line 72: IIRC, you mentioned the `NextBranchTable` should be a reference to a new table for `Prepare`, and the lifetime of that reference should be shorter than that of the module. Is that correct? Could you clarify more? Thanks.

zhouwfang:
@ia0 I posted some design comments. Could you take a look? Thanks.

crates/interpreter/src/valid.rs
ia0: I'm not sure NextBranchTable should be something different than BranchTable, at least for now. This means that this `next_function<'a>(side_table: &'a mut Self::SideTable<'m>) -> Self::BranchTable<'a, 'm>` should probably be in `ValidMode` directly. Otherwise, the alternative is to make `SideTableApi` parametric over the branch table, but that might be more complicated.

Also, I'm not sure `&BranchTable` is going to be enough for Verify, because we need to increment some index in `push_branch()`.

crates/interpreter/src/valid.rs
zhouwfang: Okay, "In a trait definition, it refers to the type implementing the trait."

crates/interpreter/src/valid.rs
zhouwfang: Do you mean returning `&'a mut Self::BranchTable<'a, 'm>` instead?

zhouwfang: If so, we would need to store the `&mut BranchTable` in `Expr`, which is now [stored by value](https://github.com/google/wasefire/blob/cac7a852e6bb6f69968469ce681ef3003a04628e/crates/interpreter/src/valid.rs#L568). That would require changing the `Expr::new()` to take a `&mut BranchTable` instead. In addition to `check_body`, `check_const` also calls `Expr::new()` and it doesn't really use `&mut BranchTable`.

Could you comment on this route?

zhouwfang: For `Prepare`, it seems easier to not pass `&mut BranchTable` to `check_body`. `check_body` still returns `BranchTable`, and we add a `wrap_up()` in `SideTableApi` that adds the returned `BranchTable` to `SideTable` for `Prepare` and does nothing for `Verify`. How about this?

ia0: You can check my `tmp/valid-mode` branch: https://github.com/ia0/wasefire/tree/tmp/valid-mode

That's the general overview. Of course some types shouldn't be public, should be renamed, should be moved to different modules. I just tried to do the minimum diff.

zhouwfang: Thanks, I see. It roughly aligns with this route except to store `Option<&mut BranchTable>` in `Expr` so that we could store `None` for `check_const`.

> If so, we would need to store the `&mut BranchTable` in `Expr`, which is now [stored by value](https://github.com/google/wasefire/blob/cac7a852e6bb6f69968469ce681ef3003a04628e/crates/interpreter/src/valid.rs#L568). That would require changing the `Expr::new()` to take a `&mut BranchTable` instead. In addition to `check_body`, `check_const` also calls `Expr::new()` and it doesn't really use `&mut BranchTable` and we can store a reference to a "global" default (it does not look nice)?
> 
> Is there a better way?

crates/interpreter/src/valid.rs
```diff
@@ -19,23 +19,197 @@ use core::cmp::Ordering;
...
+    type BranchTable<'m> = Metadata<'m>;
+    type SideTable<'m> = SideTableView<'m>;
+    type Result = ();
+
+    fn parse_side_table<'m>(parser: &mut Parser<'m>) -> Result<Self::SideTable<'m>, Error> {
+        check(parser.parse_section_id()? == SectionId::Custom)?;
+        let mut section = parser.split_section()?;
+        check(section.parse_name()? == "wasefire-sidetable")?;
+        // TODO: Make a SideTableView from the section content. We probably want SideTableView to
+        // have a state (the current function index). We also need to use u8 instead of u16.
```
zhouwfang, line 153: Could you clarify "use u8 instead of u16" for what?

crates/interpreter/src/valid.rs
```diff
@@ -19,23 +19,197 @@ use core::cmp::Ordering;
...
+        let indices_len = parser.parse_u16()? as usize;
+        let mut parser = parser.split_at(indices_len)?;
+        let indices_bytes = parser.save().get(0 .. indices_len * 2).unwrap();
+        let indices: &'m [u16] = bytemuck::cast_slice::<_, u16>(indices_bytes);
+        let metadata_len = parser.parse_u16()? as usize;
+        let parser = parser.split_at(metadata_len)?;
+        let metadata_bytes = parser.save().get(0 .. metadata_len * 2).unwrap();
+        let metadata: &'m [u16] = bytemuck::cast_slice::<_, u16>(metadata_bytes);
```
zhouwfang, line 161: I'll create a helper function to remove duplicates after making sure it is correct.

crates/interpreter/src/valid.rs
```diff
@@ -19,23 +19,197 @@ use core::cmp::Ordering;
...
+        side_table: &'a mut Self::SideTable<'m>, type_idx: usize, parser_range: Range<usize>,
+    ) -> Result<&'a mut Self::BranchTable<'m>, Error> {
+        // TODO: Increment the current function index and return the Metadata of the (now previous)
+        // function. Could be renamed to BranchTableView.
+        // TODO: Also check the type_idx and parser_range.
+        let metadata = side_table.metadata(side_table.func_idx);
+        side_table.func_idx += 1;
+        check(metadata.type_idx() == type_idx && metadata.parser_range() == parser_range)?;
+        // metadata
+        Err(Invalid)
```
zhouwfang, line 175: It seems we couldn't return a mutable reference to `metadata`. (I had a [design](https://github.com/google/wasefire/pull/726#discussion_r1942421525) for returning by value.)

crates/interpreter/src/valid.rs
ia0: Because it's going to simplify our lives by not having to care about alignment. See https://github.com/google/wasefire/pull/711#issuecomment-2574847750

crates/interpreter/src/side_table.rs
```diff
@@ -18,39 +18,41 @@ use core::ops::Range;
...
-#[allow(dead_code)]
 impl<'m> Metadata<'m> {
     pub fn type_idx(&self) -> usize {
         self.0[0] as usize
     }
 
+    #[allow(dead_code)]
     pub fn parser(&self, module: &'m [u8]) -> Parser<'m> {
-        unsafe { Parser::new(&module[self.read_u32(1) .. self.read_u32(3)]) }
+        unsafe { Parser::new(&module[self.parser_range().start .. self.parser_range().end]) }
```
ia0, line 45:
```suggestion
        unsafe { Parser::new(&module[self.parser_range()]) }
```

crates/interpreter/src/parser.rs
```diff
@@ -98,6 +98,10 @@ impl<'m, M: Mode> Parser<'m, M> {
         Ok(val)
     }
 
+    pub fn parse_u16(&mut self) -> MResult<u16, M> {
```
ia0, line 101: We shouldn't need this function. We will never encode a u16 as leb128.

crates/interpreter/src/valid.rs
ia0: Maybe this code should be in `side_table.rs` as `SideTableView::new(data: &[u8]) -> Result<Self, Error>`.

crates/interpreter/src/valid.rs
ia0: Yes, it's because `Verify::BranchTable<'m>` shouldn't be `Metadata<'m>` but some `BranchTableView<'m>` that contains a `Metadata<'m>` and a current branch index.

crates/interpreter/src/valid.rs
```diff
@@ -145,12 +318,19 @@ impl<'m, M: ValidMode> Context<'m, M> {
...
-                side_tables.push(MetadataEntry {
-                    type_idx: self.funcs[x] as usize,
-                    parser_range: Range { start: parser_start, end: parser_start + size },
-                    branch_table,
-                });
+                let branch_table =
+                    M::next_branch_table(&mut side_table, self.funcs[x] as usize, Range {
+                        start: parser_start,
+                        end: parser_start + size,
+                    });
```
ia0, line 325:
```suggestion
                    })?;
```

In particular, we shouldn't `unwrap` below. Those are user errors, not internal invariants.

crates/interpreter/src/valid.rs
```diff
@@ -898,23 +1021,25 @@ impl<'a, 'm, M: ValidMode> Expr<'a, 'm, M> {
...
         Ok(match label.kind {
             LabelKind::Block | LabelKind::If(_) => {
-                label.branches.push(source);
+                M::Branches::push_branch(&mut label.branches, source)?;
                 label.type_.results
             }
             LabelKind::Loop(target) => {
-                self.branch_table.stitch(source, target)?;
-                label.type_.params
+                let params = label.type_.params;
```
ia0, line 1032: We don't need the intermediate variable anymore.

crates/interpreter/src/valid.rs
```diff
@@ -928,7 +1053,7 @@ impl<'a, 'm, M: ValidMode> Expr<'a, 'm, M> {
     fn branch(&self) -> SideTableBranch<'m> {
         SideTableBranch {
             parser: self.parser.save(),
-            branch_table: self.branch_table.save(),
+            branch_table: self.branch_table.as_ref().map_or(0, |x| x.next_index()),
```
ia0, line 1056: It doesn't work to unwrap? I would expect `branch` to be only called outside const context.

crates/interpreter/src/valid.rs
```diff
@@ -19,23 +19,197 @@ use core::cmp::Ordering;
...
 #[derive(Default)]
-pub struct Prepare;
-impl ValidMode for Prepare {}
+struct Prepare;
+impl ValidMode for Prepare {
+    /// List of source branches.
+    type Branches<'m> = Vec<SideTableBranch<'m>>;
+    type BranchTable<'m> = Vec<BranchTableEntry>;
+    type SideTable<'m> = Vec<MetadataEntry>;
+    type Result = Vec<MetadataEntry>;
```
ia0, line 82: We could probably have this be `Vec<u8>` ultimately (and represent the content of the custom section).

ia0 (commented):
Thanks! I think it's getting there.

crates/interpreter/src/valid.rs
zhouwfang: I'm not sure I understand but I have a different solution. PTAL.

crates/interpreter/src/valid.rs
zhouwfang: No, it could be `None`. I guess it is because `branch_target()` is called for `Verify`.

crates/interpreter/src/valid.rs
zhouwfang: Should we convert `MetadataEntry` to `Vec<u8>` directly?

If you don't mind, I'll do that in the next PR when I add the `merge()`.

zhouwfang: Can we convert it in `merge()` in https://github.com/zhouwfang/wasefire/pull/8? Please also take a look at the design in that PR. Thanks!

ia0 (approved):
Thanks! We can do the rest in followup PRs.

PR #729, opened 2025-02-07
Do not assume support in rng_test

ia0:


examples/rust/rng_test/src/lib.rs
```diff
@@ -24,6 +24,10 @@ fn main() {
 
 fn test_non_constant() {
     debug!("test_non_constant(): This should generate 5 different buffers.");
+    if rng::fill_bytes(&mut []).is_err() {
```
ia0-review, line 27: We should check that the error is actually not supported.

ia0-review (approved):

examples/rust/rng_test/src/lib.rs
ia0: Done, also fixed a few bunch other stuff.

PR #730, opened 2025-02-10
Merge side table into WASM module

zhouwfang:
- Merge side table into WASM module as custom section
- Address review comments left in #726.

I'll debug in the meanwhile.

#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -19,13 +19,13 @@ use core::cmp::Ordering;
 use core::marker::PhantomData;
 use core::ops::Range;
 
+use crate::Error::Invalid;
 use crate::error::*;
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
 use crate::util::*;
 use crate::*;
-use crate::Error::Invalid;
 
 pub fn merge(binary: &[u8], side_table: Vec<MetadataEntry>) -> Result<Vec<u8>, Error> {
     // Convert Vec<MetadataEntry> to Vec<u8>.
@@ -147,7 +147,7 @@ impl ValidMode for Verify {
     /// Contains at most one _target_ branch. Source branches are eagerly patched to
     /// their target branch using the branch table.
     type Branches<'m> = Option<SideTableBranch<'m>>;
-    type BranchTable<'m> = Metadata<'m>;
+    type BranchTable<'m> = BranchTableView<'m>;
     type SideTable<'m> = SideTableView<'m>;
     type Result = ();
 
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/valid.rs
```diff
@@ -26,6 +26,23 @@ use crate::toctou::*;
 use crate::util::*;
 use crate::*;
 
+pub fn merge(binary: &[u8], side_table: Vec<MetadataEntry>) -> Vec<u8> {
```
ia0, line 29: We can take a slice: `&[MetadataEntry]`.

crates/interpreter/src/valid.rs
```diff
@@ -26,6 +26,23 @@ use crate::toctou::*;
...
+    let mut wasm = vec![];
+    wasm.extend_from_slice(&binary[0 .. 8]);
+    wasm.push(0);
+    wasm.extend_from_slice(&side_table.len().to_le_bytes());
```
ia0, line 33: This one needs to be leb128 since that's part of wasm spec.

crates/interpreter/src/valid.rs
```diff
@@ -26,6 +26,23 @@ use crate::toctou::*;
...
+    for entry in side_table {
+        wasm.extend_from_slice(&entry.type_idx.to_le_bytes());
```
ia0, line 35: We need to convert the `usize` to `u16` first.

crates/interpreter/src/valid.rs
```diff
@@ -26,6 +26,23 @@ use crate::toctou::*;
...
+        wasm.extend_from_slice(&entry.parser_range.start.to_le_bytes());
+        wasm.extend_from_slice(&entry.parser_range.end.to_le_bytes());
```
ia0, line 37: Same here, `usize` has not a fixed size, and we expect a `u32`.

crates/interpreter/src/valid.rs
```diff
@@ -26,6 +26,23 @@ use crate::toctou::*;
 use crate::util::*;
 use crate::*;
 
+pub fn merge(binary: &[u8], side_table: Vec<MetadataEntry>) -> Vec<u8> {
+    let mut wasm = vec![];
+    wasm.extend_from_slice(&binary[0 .. 8]);
+    wasm.push(0);
+    wasm.extend_from_slice(&side_table.len().to_le_bytes());
+    for entry in side_table {
```
ia0, line 34: We also need to serialize the indices. Below is just the metadata.

ia0 (commented):
Just did a quick review. I think we can also change `next_branch_table` to return `Self::BranchTable<'a, 'm>` instead of `&'a mut Self::BranchTable<'m>`. This will permit removing the `branch_table_view` field of `SideTableView`. Also I think we can split this PR:
- One for the change above.
- One for using `[u8]` instead of `[u16]`.
- One for serializing/deserializing the side-table.
- One for adding and using merge.

The first 2 can be done in parallel.

zhouwfang:
Done in https://github.com/zhouwfang/wasefire/pull/9 instead.

PR #735, opened 2025-02-10
Make `next_branch_table()` return value

zhouwfang:
Based on [this design](https://github.com/google/wasefire/pull/726#discussion_r1942421525).

#46

### DIFF
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -22,7 +22,6 @@ pub struct SideTableView<'m> {
     pub func_idx: usize,
     pub indices: &'m [u16], // including 0 and the length of metadata_array
     pub metadata: &'m [u16],
-    pub branch_table_view: Metadata<'m>,
 }
 
 impl<'m> SideTableView<'m> {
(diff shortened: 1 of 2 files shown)


ia0 (commented):
That's not what I had in mind. This can be much simpler by having `Prepare::BranchTable<'a, 'm> = &'a Vec<BranchTableEntry>`.

zhouwfang:
> That's not what I had in mind. This can be much simpler by having `Prepare::BranchTable<'a, 'm> = &'a Vec<BranchTableEntry>`.

Hmm, I don't quite understand. Could you clarify how that helps? What type for the branch table would you store in `Expr` in that case?

ia0:
> Hmm, I don't quite understand. Could you clarify how that helps? What type for the branch table would you store in `Expr` in that case?

I've did the change since it's simpler than explain it.

ia0 (approved):

zhouwfang:
> That's not what I had in mind. This can be much simpler by having `Prepare::BranchTable<'a, 'm> = &'a Vec<BranchTableEntry>`.

Thanks, you actually meant `Prepare::BranchTable<'a, 'm> = &'a mut Vec<BranchTableEntry>`.

ia0:
> Thanks, you actually meant `Prepare::BranchTable<'a, 'm> = &'a mut Vec<BranchTableEntry>`.

Oh yes, that was a typo I didn't see.

PR #736, opened 2025-02-10
Use `[u8]` instead of `[u16]` to not rely on alignment

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -26,8 +26,7 @@ pub struct SideTableView<'m> {
 
 impl<'m> SideTableView<'m> {
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
-        let num_functions =
-            bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&binary[0 .. 2])) as usize;
+        let num_functions = u16::from_le_bytes(binary[.. 2].try_into().unwrap()) as usize;
         let indices_end = 2 + (num_functions + 1) * 2;
         Ok(SideTableView {
             func_idx: 0,
@@ -38,10 +37,14 @@ impl<'m> SideTableView<'m> {
 
     pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
-            &self.metadata
-                [self.indices[func_idx * 2] as usize .. self.indices[(func_idx + 1) * 2] as usize],
+            &self.metadata[parse_u16(self.indices, func_idx * 2) as usize
+                .. parse_u16(self.indices, (func_idx + 1) * 2) as usize],
         )
     }
+}
+
+fn parse_u16(data: &[u8], offset: usize) -> u16 {
+    u16::from_le_bytes(data[offset ..][.. 2].try_into().unwrap())
 }
 
 #[derive(Default, Copy, Clone)]
@@ -49,7 +52,7 @@ pub struct Metadata<'m>(&'m [u8]);
 
 impl<'m> Metadata<'m> {
     pub fn type_idx(&self) -> usize {
-        bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&self.0[0 .. 2])) as usize
+        u16::from_le_bytes(self.0[.. 2].try_into().unwrap()) as usize
     }
 
     #[allow(dead_code)]
@@ -66,7 +69,7 @@ impl<'m> Metadata<'m> {
     }
 
     fn read_u32(&self, idx: usize) -> usize {
-        bytemuck::pod_read_unaligned::<u32>(bytemuck::cast_slice(&self.0[idx .. idx + 4])) as usize
+        u32::from_le_bytes(self.0[idx .. idx + 4].try_into().unwrap()) as usize
     }
 }
 
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -59,7 +59,7 @@ impl<'m> Metadata<'m> {
     }
 
     pub fn branch_table(&self) -> &[BranchTableEntry] {
-        let entry_size = 6;
+        let entry_size = size_of::<BranchTableEntry>();
         assert_eq!(
             self.0.len() % entry_size,
             0,
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -20,43 +20,36 @@ use crate::module::Parser;
 
 pub struct SideTableView<'m> {
     pub func_idx: usize,
-    pub indices: &'m [u16], // including 0 and the length of metadata_array
-    pub metadata: &'m [u16],
+    pub indices: &'m [u8], // including 0 and the length of metadata_array
+    pub metadata: &'m [u8],
 }
 
 impl<'m> SideTableView<'m> {
-    pub fn new(parser: &mut crate::valid::Parser<'m>) -> Result<Self, Error> {
+    pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
+        let num_functions =
+            bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&binary[0 .. 2])) as usize;
+        let indices_end = 2 + (num_functions + 1) * 2;
         Ok(SideTableView {
             func_idx: 0,
-            indices: parse_side_table_field(parser)?,
-            metadata: parse_side_table_field(parser)?,
+            indices: &binary[2 .. indices_end],
+            metadata: &binary[indices_end ..],
         })
     }
 
     pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
-            &self.metadata[self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
+            &self.metadata
+                [self.indices[func_idx * 2] as usize .. self.indices[(func_idx + 1) * 2] as usize],
         )
     }
 }
 
-fn parse_u16(data: &[u8]) -> u16 {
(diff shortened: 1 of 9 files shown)


crates/interpreter/src/side_table.rs
```diff
@@ -20,45 +20,38 @@ use crate::module::Parser;
...
+    pub indices: &'m [u8], // including 0 and the length of metadata_array
+    pub metadata: &'m [u8],
     pub branch_table_view: Metadata<'m>,
 }
 
 impl<'m> SideTableView<'m> {
-    pub fn new(parser: &mut crate::valid::Parser<'m>) -> Result<Self, Error> {
+    pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
+        let num_functions =
+            bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&binary[0 .. 2])) as usize;
```
ia0, line 31: We don't need this anymore since `[u8]` is always aligned. We can simply `u16::from_le_bytes(binary[.. 2].try_into().unwrap())` (maybe with some `&` and parentheses).

crates/interpreter/src/side_table.rs
```diff
@@ -20,45 +20,38 @@ use crate::module::Parser;
...
+            metadata: &binary[indices_end ..],
             branch_table_view: Default::default(),
         })
     }
 
     pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
         Metadata(
-            &self.metadata[self.indices[func_idx] as usize .. self.indices[func_idx + 1] as usize],
+            &self.metadata
+                [self.indices[func_idx * 2] as usize .. self.indices[(func_idx + 1) * 2] as usize],
```
ia0, line 44: This is wrong, we need to read 2 bytes. We could use a `parse_u16(offset: usize) -> u16` helper which simply does `u16::from_le_bytes(self.binary[offset..][..2].try_into().unwrap())` or something like that.

Also, we probably shouldn't panic if the side table is corrupted during verification. If some functions are used both in Check and Use mode, then we need to make them generic.

crates/interpreter/src/side_table.rs
```diff
@@ -67,15 +60,15 @@ impl<'m> Metadata<'m> {
...
     }
 
     pub fn parser_range(&self) -> Range<usize> {
-        self.read_u32(1) .. self.read_u32(3)
+        self.read_u32(2) .. self.read_u32(6)
     }
 
     fn read_u32(&self, idx: usize) -> usize {
-        bytemuck::pod_read_unaligned::<u32>(bytemuck::cast_slice(&self.0[idx .. idx + 2])) as usize
+        bytemuck::pod_read_unaligned::<u32>(bytemuck::cast_slice(&self.0[idx .. idx + 4])) as usize
```
ia0, line 71: Same here, we can use `u32::from_le_bytes()`. We probably don't need bytemuck anymore.

ia0 (commented):
Thanks! We shouldn't need to use bytemuck anymore and can write our own `parse_u16` and `parse_u32` functions (probably with toctou).

crates/interpreter/src/side_table.rs
zhouwfang: It seems this function is only used in `next_branch_table()` for the `Verify` mode. Do you mean we should return `Result<Metadata<'m>, Error>` instead?

ia0: It's because `Module::side_table` is currently wrong. It will be the `SideTableView` eventually, and then it will use this function to access the parser range and branch table.

Yes, I mean we should return `MResult<Metadata<'m>, M>`. And `Module::new_unchecked()` should do the initial parsing. This could be done in another PR though.

crates/interpreter/src/side_table.rs
```diff
@@ -20,43 +20,39 @@ use crate::module::Parser;
...
-    pub indices: &'m [u16], // including 0 and the length of metadata_array
-    pub metadata: &'m [u16],
+    pub indices: &'m [u8], // including 0 and the length of metadata_array
+    pub metadata: &'m [u8],
 }
 
 impl<'m> SideTableView<'m> {
-    pub fn new(parser: &mut crate::valid::Parser<'m>) -> Result<Self, Error> {
+    pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
+        let num_functions = u16::from_le_bytes(binary[.. 2].try_into().unwrap()) as usize;
```
ia0, line 29:
```suggestion
        let num_functions = parse_u16(binary, 0) as usize;
```

crates/interpreter/src/side_table.rs
```diff
@@ -20,43 +20,39 @@ use crate::module::Parser;
...
-fn parse_u16(data: &[u8]) -> u16 {
-    bytemuck::pod_read_unaligned::<u16>(bytemuck::cast_slice(&data[0 .. 2]))
-}
-
-fn parse_side_table_field<'m>(parser: &mut crate::valid::Parser<'m>) -> Result<&'m [u16], Error> {
-    let len = parse_u16(parser.save()) as usize;
-    let parser = parser.split_at(len)?;
-    let bytes = parser.save().get(0 .. len * 2).unwrap();
-    Ok(bytemuck::cast_slice::<_, u16>(bytes))
+fn parse_u16(data: &[u8], offset: usize) -> u16 {
```
ia0, line 46: We could make this return `usize` for convenience. The `u16` is just for representation. We could also make this take an index (as if the slice was a u16 slice) instead of an offset, so that the multiplication by 2 is done here instead of all callers.

crates/interpreter/src/side_table.rs
```diff
@@ -20,43 +20,39 @@ use crate::module::Parser;
...
 }
 
 #[derive(Default, Copy, Clone)]
-pub struct Metadata<'m>(&'m [u16]);
+pub struct Metadata<'m>(&'m [u8]);
 
 impl<'m> Metadata<'m> {
     pub fn type_idx(&self) -> usize {
-        self.0[0] as usize
+        u16::from_le_bytes(self.0[.. 2].try_into().unwrap()) as usize
```
ia0, line 55: This can also use parse_u16

crates/interpreter/src/side_table.rs
```diff
@@ -65,15 +61,15 @@ impl<'m> Metadata<'m> {
...
     }
 
     pub fn parser_range(&self) -> Range<usize> {
-        self.read_u32(1) .. self.read_u32(3)
+        self.read_u32(2) .. self.read_u32(6)
     }
 
     fn read_u32(&self, idx: usize) -> usize {
-        bytemuck::pod_read_unaligned::<u32>(bytemuck::cast_slice(&self.0[idx .. idx + 2])) as usize
+        u32::from_le_bytes(self.0[idx .. idx + 4].try_into().unwrap()) as usize
```
ia0, line 72: similarly we could move this to a parse_u32 (however here we can't take an index, we must take an offset, so probably we should also continue taking an offset for parse_u16)

crates/interpreter/src/side_table.rs
```diff
@@ -65,15 +61,15 @@ impl<'m> Metadata<'m> {
     }
 
     pub fn branch_table(&self) -> &[BranchTableEntry] {
-        bytemuck::cast_slice(&self.0[5 ..])
+        bytemuck::cast_slice(&self.0[10 ..])
```
ia0, line 64: We still have the alignment problem here. BranchTableEntry shouldn't use `u16`, it should be `[u8; 6]`. And then we can remove the dependency on bytemuck.

ia0 (commented):
Thanks! It's getting closer.

crates/interpreter/src/side_table.rs
zhouwfang: Do you basically mean [prepare()](https://github.com/zhouwfang/wasefire/blob/1aa931b68be427f449cafe39f17e9eccd18d41b2/crates/interpreter/src/valid.rs#L30) should return `SideTableView`?

crates/interpreter/src/side_table.rs
zhouwfang: Returning `u16` is useful in the new code.

crates/interpreter/src/side_table.rs
```diff
@@ -20,43 +20,37 @@ use crate::module::Parser;
...
 impl<'m> SideTableView<'m> {
-    pub fn new(parser: &mut crate::valid::Parser<'m>) -> Result<Self, Error> {
+    pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
+        let num_functions = parse_u16(binary, 0) as usize;
+        let indices_end = 2 + (num_functions + 1) * 2;
         Ok(SideTableView {
             func_idx: 0,
-            indices: parse_side_table_field(parser)?,
-            metadata: parse_side_table_field(parser)?,
+            indices: &binary[2 .. indices_end],
```
ia0, line 33: The TODO below also applies to this function. In particular, we should use `.get(2 .. indices_end)` to return an error instead of panicking.

crates/interpreter/src/side_table.rs
```diff
@@ -65,15 +59,23 @@ impl<'m> Metadata<'m> {
     }
 
     pub fn branch_table(&self) -> &[BranchTableEntry] {
-        bytemuck::cast_slice(&self.0[5 ..])
+        let entry_size = size_of::<BranchTableEntry>();
+        assert_eq!(
```
ia0, line 63: Same here, we probably want to use `MResult`.

ia0 (approved):
Thanks!

crates/interpreter/src/side_table.rs
ia0: No, `prepare()` is not concerned. It's only `Module::new_unchecked()`. More generally, it's only relevant for parsing (views) and not formatting (`prepare()`).

PR #738, opened 2025-02-13
Add timer::Timeout and use it in opensk

ia0:


examples/rust/opensk/src/env.rs
```diff
@@ -74,3 +77,11 @@ impl Env for WasefireEnv {
 }
 
 impl opensk_lib::api::key_store::Helper for WasefireEnv {}
+
+fn convert_error(error: Error) -> Ctap2StatusCode {
+    match (Space::try_from(error.space()), Code::try_from(error.code())) {
```
ia0-review, line 82: No need to parse the code.

examples/rust/opensk/src/env/hid_connection.rs
```diff
@@ -14,15 +14,34 @@
...
+use wasefire::usb::ctap;
 
-use crate::env::WasefireEnv;
+use crate::env::{WasefireEnv, convert_error};
 
 impl HidConnection for WasefireEnv {
-    fn send(&mut self, _buf: &[u8; 64], _endpoint: UsbEndpoint) -> CtapResult<()> {
-        todo!()
+    fn send(&mut self, packet: &[u8; 64], endpoint: UsbEndpoint) -> CtapResult<()> {
+        match endpoint {
```
ia0-review, line 24: Can we do this with a let?

ia0-review (changes requested):

ia0-review (approved):

PR #744, opened 2025-02-16
Complete implementation for `Verify`

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/valid.rs
+++ b/crates/interpreter/src/valid.rs
@@ -712,11 +712,7 @@ impl<'a, 'm, M: ValidMode> Expr<'a, 'm, M> {
                         let result = self.label().type_.results.len();
                         let mut target = self.branch_target(result);
                         target.branch_table += 1;
-                        M::BranchTable::stitch_branch(
-                            self.branch_table.as_mut().unwrap(),
-                            source,
-                            target,
-                        )?;
+                        self.branch_table.as_mut().unwrap().stitch_branch(source, target)?;
                     }
                     _ => Err(invalid())?,
                 }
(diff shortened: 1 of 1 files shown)


zhouwfang:
@ia0 Could you also take a look at the [follow-up PR](https://github.com/zhouwfang/wasefire/pull/9/files)? Thanks.

crates/interpreter/src/valid.rs
```diff
@@ -132,13 +132,18 @@ impl<'m> BranchTableApi<'m> for &mut Vec<BranchTableEntry> {
     }
 }
 
+struct MetadataView<'m> {
```
ia0, line 135: This is good to have side table views specific to verification in `valid.rs` because `exec.rs` will need different views (it won't need the mutable index used only for verification).

crates/interpreter/src/valid.rs
```diff
@@ -170,26 +175,29 @@ impl<'m> BranchesApi<'m> for Option<SideTableBranch<'m>> {
...
+impl<'m> BranchTableApi<'m> for MetadataView<'m> {
     fn stitch_branch(
         &mut self, source: SideTableBranch<'m>, target: SideTableBranch<'m>,
     ) -> CheckResult {
         check(source == target)
     }
 
     fn patch_branch(&self, mut source: SideTableBranch<'m>) -> Result<SideTableBranch<'m>, Error> {
-        let entry = self.branch_table()[source.branch_table].view();
+        source.branch_table = self.branch_idx;
```
ia0, line 186: I'm a bit surprised by this. In theory, we should have `source.branch_table + 1 == self.branch_idx`, so this smells like an off-by-one error. Let's merge like this but keep it in mind. This might be a source of error.

ia0 (approved):
Thanks! I've pushed an unrelated simplification on top that occurred to me while reviewing.

crates/interpreter/src/valid.rs
zhouwfang: Could you explain why you believe so? 

Before this function, `next_index()` and `allocate_branch()` are called. 

For `Prepare`, `next_index()` returns the length of the branch table, and `allocate_branch()` pushes an invalid entry to the branch table, which `patch_branch` returns.

For `Verify`, to be consistent with `Prepare`, I thought we should also use the next branch in `patch_branch()`.

PR #758, opened 2025-02-23
Serialize side table and integrate `verify`

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/format.rs
+++ b/crates/interpreter/src/format.rs
@@ -23,13 +23,12 @@ pub fn leb128(wasm: &mut Vec<u8>, mut x: usize) {
     wasm.push(x as u8);
 }
 
-pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: Option<&str>) {
+pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: &str) {
     assert!(id <= 12);
     wasm.push(id);
-    leb128(wasm, content.len());
-    if let Some(name) = name {
-        leb128(wasm, name.len());
-        wasm.extend_from_slice(name.as_bytes());
-    }
+    let name_bytes = name.as_bytes();
+    leb128(wasm, 1 + name_bytes.len() + content.len());
+    leb128(wasm, name_bytes.len());
+    wasm.extend_from_slice(name.as_bytes());
     wasm.extend_from_slice(content);
 }
--- a/crates/interpreter/src/format.rs
+++ b/crates/interpreter/src/format.rs
@@ -14,7 +14,7 @@
 
 use alloc::vec::Vec;
 
-pub fn leb128(mut x: usize, wasm: &mut Vec<u8>) {
+pub fn leb128(wasm: &mut Vec<u8>, mut x: usize) {
     assert!(x <= u32::MAX as usize);
     while x > 127 {
         wasm.push(0x80 | (x & 0x7f) as u8);
@@ -26,9 +26,9 @@ pub fn leb128(mut x: usize, wasm: &mut Vec<u8>) {
 pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: Option<&str>) {
     assert!(id <= 12);
     wasm.push(id);
-    leb128(content.len(), wasm);
+    leb128(wasm, content.len());
     if let Some(name) = name {
-        leb128(name.len(), wasm);
+        leb128(wasm, name.len());
         wasm.extend_from_slice(name.as_bytes());
     }
     wasm.extend_from_slice(content);
--- a/crates/interpreter/src/format.rs
+++ b/crates/interpreter/src/format.rs
@@ -23,12 +23,12 @@ pub fn leb128(wasm: &mut Vec<u8>, mut x: usize) {
     wasm.push(x as u8);
 }
 
-pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: &str) {
+pub fn custom_section(wasm: &mut Vec<u8>, id: u8, name: &str, content: &[u8]) {
     assert!(id <= 12);
     wasm.push(id);
-    let name_bytes = name.as_bytes();
-    leb128(wasm, 1 + name_bytes.len() + content.len());
-    leb128(wasm, name_bytes.len());
+    assert!(name.len() < 128);
+    leb128(wasm, 1 + name.len() + content.len());
+    leb128(wasm, name.len());
     wasm.extend_from_slice(name.as_bytes());
     wasm.extend_from_slice(content);
 }
--- a/crates/interpreter/src/format.rs
+++ b/crates/interpreter/src/format.rs
@@ -23,12 +23,11 @@ pub fn leb128(wasm: &mut Vec<u8>, mut x: usize) {
     wasm.push(x as u8);
 }
 
-pub fn custom_section(wasm: &mut Vec<u8>, id: u8, name: &str, content: &[u8]) {
-    assert!(id <= 12);
-    wasm.push(id);
+pub fn custom_section(wasm: &mut Vec<u8>, name: &str, content: &[u8]) {
     assert!(name.len() < 128);
+    wasm.push(0);
     leb128(wasm, 1 + name.len() + content.len());
-    leb128(wasm, name.len());
+    wasm.push(name.len() as u8);
     wasm.extend_from_slice(name.as_bytes());
     wasm.extend_from_slice(content);
 }
--- a/crates/interpreter/src/format.rs
+++ b/crates/interpreter/src/format.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,10 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO(dev/fast-interp): Add debug asserts when `off` is positive and negative, and `toctou`
-// support.
-pub fn offset_front<T>(cur: &[T], off: isize) -> &[T] {
-    unsafe {
-        core::slice::from_raw_parts(cur.as_ptr().offset(off), (cur.len() as isize - off) as usize)
+use alloc::vec::Vec;
+
+pub fn leb128(mut x: usize, wasm: &mut Vec<u8>) {
+    assert!(x <= u32::MAX as usize);
+    while x > 127 {
+        wasm.push(0x80 | (x & 0x7f) as u8);
+        x >>= 7;
     }
+    wasm.push(x as u8);
+}
+
+pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: Option<&str>) {
+    assert!(id <= 12);
+    wasm.push(id);
+    leb128(content.len(), wasm);
+    if let Some(name) = name {
+        leb128(name.len(), wasm);
+        wasm.extend_from_slice(name.as_bytes());
+    }
+    wasm.extend_from_slice(content);
 }
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -19,7 +19,6 @@ use crate::parser::{SkipData, SkipElem};
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
-use crate::valid::verify;
 use crate::*;
 
 /// Valid module.
@@ -50,9 +49,8 @@ impl ImportDesc {
 impl<'m> Module<'m> {
     /// Validates a WASM module in binary format.
     pub fn new(binary: &'m [u8]) -> Result<Self, Error> {
-        verify(binary)?;
-        let module = unsafe { Self::new_unchecked(&binary) };
-        Ok(module)
+        crate::valid::verify(binary)?;
+        Ok(unsafe { Self::new_unchecked(&binary) })
     }
 
     /// Creates a valid module from binary format.
@@ -61,13 +59,10 @@ impl<'m> Module<'m> {
     ///
     /// The module must be valid.
     pub unsafe fn new_unchecked(binary: &'m [u8]) -> Self {
-        let mut parser = unsafe { Parser::new(&binary[8 ..]) };
-        let mut module = Module {
-            // Only keep the sections (i.e. skip the header).
-            binary: &binary[8 ..],
-            types: Vec::new(),
-            side_table: parser.parse_side_table().unwrap(),
-        };
+        // Only keep the sections (i.e. skip the header).
+        let binary = &binary[8 ..];
+        let side_table = unsafe { Parser::new(binary) }.parse_side_table().into_ok();
+        let mut module = Module { binary, types: Vec::new(), side_table };
         if let Some(mut parser) = module.section(SectionId::Type) {
             for _ in 0 .. parser.parse_vec().into_ok() {
                 module.types.push(parser.parse_functype().into_ok());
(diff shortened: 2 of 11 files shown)


crates/interpreter/src/side_table.rs
```diff
@@ -69,7 +71,7 @@ impl<'m> Metadata<'m> {
         unsafe {
             core::slice::from_raw_parts(
                 self.0[10 ..].as_ptr() as *const BranchTableEntry,
-                self.0.len() / entry_size,
+                self.0.len() / entry_size - 1,
```
ia0, line 74: Why is there a `- 1` here?

crates/interpreter/src/side_table.rs
```diff
@@ -86,6 +88,32 @@ pub struct MetadataEntry {
...
+pub fn serialize(side_table: &[MetadataEntry]) -> Result<Vec<u8>, Error> {
+    let mut res = vec![];
+    res.extend_from_slice(&(side_table.len() as u16).to_le_bytes());
+    let mut index = 0;
+    res.extend_from_slice(&(index as u16).to_le_bytes());
+    for entry in side_table {
+        index =
+            try_from::<u16>("index of MetadataEntry", index + 10 + 6 * entry.branch_table.len())?
+                as usize;
+        res.extend_from_slice(&(index as u16).to_le_bytes());
```
ia0, line 100: Looking at this, we could actually have a `push_as<T: TryFrom<usize>>(dst: &mut Vec<u8>, src: usize, msg: &str)` helper instead of `try_from`. This should make this function much easier to read.

ia0 (commented):
Thanks! I think the issues come from the fact that the side-table format is not correct yet.

crates/interpreter/src/module.rs
```diff
@@ -73,6 +67,15 @@ impl<'m> Module<'m> {
                 module.types.push(parser.parse_functype().into_ok());
             }
         }
+        if let Some(mut parser) = module.section(SectionId::Custom) {
+            let section_name = "wasefire-sidetable";
+            if parser.parse_name().into_ok() == section_name {
+                let side_table_view = SideTableView::new(parser.save()).unwrap();
+                module.side_table = side_table_view;
+                let len = 3 + section_name.as_bytes().len() + parser.save().len();
+                module.binary = &binary[8 + len ..];
```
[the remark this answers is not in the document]
zhouwfang: Are you sure? I checked the debugging logs and it does not seem the case.

crates/interpreter/src/module.rs
```diff
@@ -178,8 +178,11 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let MetadataEntry { parser_range, branch_table, .. } = &self.side_table[x as usize];
-        (unsafe { Parser::new(&self.binary[parser_range.clone()]) }, branch_table)
+        let metadata = Box::leak(Box::new(self.side_table.metadata(x as usize)));
+        (
+            unsafe { Parser::new(&self.binary[metadata.parser_range().clone()]) },
```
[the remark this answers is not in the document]
zhouwfang: It does return a `Range`. I removed the `clone()`.

crates/interpreter/src/module.rs
```diff
@@ -178,8 +178,11 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let MetadataEntry { parser_range, branch_table, .. } = &self.side_table[x as usize];
-        (unsafe { Parser::new(&self.binary[parser_range.clone()]) }, branch_table)
+        let metadata = Box::leak(Box::new(self.side_table.metadata(x as usize)));
```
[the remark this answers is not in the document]
zhouwfang: `metadata()` returns a temporary value.

crates/interpreter/src/side_table.rs
zhouwfang: T should also satisfy `ToBytes` for the `to_le_bytes()` to work. But `ToBytes` is not implemented for `u16` and `u32`.

crates/interpreter/src/module.rs
```diff
@@ -73,6 +67,15 @@ impl<'m> Module<'m> {
                 module.types.push(parser.parse_functype().into_ok());
             }
         }
+        if let Some(mut parser) = module.section(SectionId::Custom) {
```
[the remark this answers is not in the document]
zhouwfang: Do you mean not calling `section()` but writing a new `first/sidetable_section()`?

crates/interpreter/src/format.rs
```diff
@@ -0,0 +1,34 @@
...
+pub fn leb128(wasm: &mut Vec<u8>, mut x: usize) {
+    assert!(x <= u32::MAX as usize);
+    while x > 127 {
+        wasm.push(0x80 | (x & 0x7f) as u8);
+        x >>= 7;
+    }
+    wasm.push(x as u8);
+}
+
+pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: &str) {
```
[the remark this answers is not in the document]
zhouwfang: Can I rename it to be `custom_section`?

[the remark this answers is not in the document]
ia0: Yes. In that case, let's move `name` before `content` since that's the order in the binary format.

crates/interpreter/src/format.rs
```diff
@@ -0,0 +1,34 @@
...
+    assert!(id <= 12);
+    wasm.push(id);
+    let name_bytes = name.as_bytes();
+    leb128(wasm, 1 + name_bytes.len() + content.len());
```
ia0, line 30: This assumes `name.len() < 128`. We should assert it.

crates/interpreter/src/format.rs
```diff
@@ -0,0 +1,34 @@
...
+        wasm.push(0x80 | (x & 0x7f) as u8);
+        x >>= 7;
+    }
+    wasm.push(x as u8);
+}
+
+pub fn section(wasm: &mut Vec<u8>, id: u8, content: &[u8], name: &str) {
+    assert!(id <= 12);
+    wasm.push(id);
+    let name_bytes = name.as_bytes();
```
ia0, line 29: No need for this variable: `name.len() == name.as_bytes().len()`.

crates/interpreter/src/module.rs
```diff
@@ -20,17 +20,14 @@ use crate::parser::{SkipData, SkipElem};
 use crate::side_table::*;
 use crate::syntax::*;
 use crate::toctou::*;
-use crate::valid::prepare;
 use crate::*;
 
 /// Valid module.
 #[derive(Debug, Default)]
```
ia0, line 26: Let's merge `dev/fast-interp` to get rid of this `Default`.

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
ia0: We should use `Verify::parse_side_table(&mut Parser::new(module.binary))`. Or better, we should move the body of this function into `parser.rs` returning `MResult`, because in `valid.rs` we use a `Parser<Check>` but here in `module.rs` we use a `Parser<Use>`.

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
ia0: Custom sections are skipped by the `if actual_id == SectionId::Custom` test in `Module::section()`. Not sure what you mean by "not skipped".

crates/interpreter/src/module.rs
```diff
@@ -109,9 +112,6 @@ impl<'m> Module<'m> {
             }
             let actual_id = parser.parse_section_id().into_ok();
             let section = parser.split_section().into_ok();
-            if actual_id == SectionId::Custom {
```
ia0, line 112: Why are you removing this?

crates/interpreter/src/module.rs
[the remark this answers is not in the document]
ia0: Ah the problem is the type of `Metadata::branch_table()`. It should return `&'m [BranchTableEntry]`.

crates/interpreter/src/side_table.rs
```diff
@@ -69,7 +71,7 @@ impl<'m> Metadata<'m> {
         unsafe {
             core::slice::from_raw_parts(
                 self.0[10 ..].as_ptr() as *const BranchTableEntry,
-                self.0.len() / entry_size,
+                (self.0.len() - 10) / entry_size,
```
ia0, line 74: Let's improve this by using an intermediate variable:

```rust
let branch_table = &self.0[10 ..];
assert_eq!(branch_table.len() % entry_size, 0);
from_raw_parts(branch_table.as_ptr().cast(), branch_table.len() / entry_size)
```

crates/interpreter/src/side_table.rs
ia0: I see. There's actually no trait for that sadly. `ToBytes` is for SIMD. We could create our own but then it's not worth it. Let's leave it like that.

ia0, line 100: We should actually have `index: u16`. The casts are at the wrong places. It should be:

```rust
    let mut index = 0u16;
    res.extend_from_slice(&index.to_le_bytes());
    for entry in side_table {
        index = try_from::<u16>(
            "index of MetadataEntry",
            index as usize + 10 + 6 * entry.branch_table.len(),
        )?;
        res.extend_from_slice(&index.to_le_bytes());
    }
```

zhouwfang:
Now that `valid.rs` becomes generic over `ValidMode`, I wonder how I could print debugging logs for generic objects. For example, I'm not sure how to print the next [branch table](https://github.com/zhouwfang/wasefire/blob/bc7d5762ddef7ed46287819ebf7e8e52d5330c2b/crates/interpreter/src/valid.rs#L314). I added `Debug` at some places, but there still seems something missing. Could you show me the code for it so that I could debug over the weekends? Thanks!

ia0:
> Now that `valid.rs` becomes generic over `ValidMode`, I wonder how I could print debugging logs for generic objects.

I've pushed a commit for that (called `debug`). For generic contexts you need the `Debug` bound on the associated types (not just on the final types used for those associated types in the trait implementations).

crates/interpreter/src/valid.rs
```diff
@@ -179,14 +185,18 @@ impl<'m> BranchTableApi<'m> for MetadataView<'m> {
     fn stitch_branch(
         &mut self, source: SideTableBranch<'m>, target: SideTableBranch<'m>,
     ) -> CheckResult {
-        check(source == target)
+        check(
+            source.parser == target.parser
+                && source.result == target.result
+                && source.branch_table == target.branch_table
+                && source.stack <= target.stack,
```
zhouwfang, line 192: @ia0 This is consistent with [the logic](https://github.com/zhouwfang/wasefire/blob/b1e4cb9884707897da37ee5e2ca1db322cb4b2c2/crates/interpreter/src/valid.rs#L1146-L1148) in `pop_cnt()`. To be completely rigorous, we would need to add `pop_cnt == 0`, but it does not seem straightforward to obtain `pop_cnt` whenever `stitch_branch()` is called.

zhouwfang, line 192: @ia0 This is a necessary condition of [the logic](https://github.com/zhouwfang/wasefire/blob/b1e4cb9884707897da37ee5e2ca1db322cb4b2c2/crates/interpreter/src/valid.rs#L1146-L1148) in `pop_cnt()`. The actual equivalent condition should be `source.stack == target.stack || (pop_cnt == 0 && source.stack < target.stack)`, but it does not seem straightforward to obtain `pop_cnt` whenever `stitch_branch()` is called.

zhouwfang: Update: we can store `pop_cnt` as `Option<usize>` in `SideTableBranch`.

zhouwfang:
The only test that fails in the test suite now is [the huge br_table](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107) because of ["index of MetadataEntry overflow"](https://github.com/zhouwfang/wasefire/blob/3b732824890cbdac9089da32f659f3d80279563c/crates/interpreter/src/side_table.rs#L86-L89) (the index is 97252). Do we want to change it to u32?

I suspect the error in `./scripts/hwci.sh` is due to the same reason.

crates/interpreter/src/valid.rs
ia0: I'll need to look into this when I get time. I think the whole situation around pop_cnt feels wrong, but maybe I'm missing something. I've added some TODOs to not lose track of this. This should not block this PR.

ia0 (approved):
Thanks! That's a huge step, but I'm not convinced about the pop_cnt story yet. I'll need to look into it. Ideally we should do some differential fuzzing, I don't trust the test suite to provide good coverage.

ia0:
> The only test that fails in the test suite now is [the huge br_table](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107) because of ["index of MetadataEntry overflow"](https://github.com/zhouwfang/wasefire/blob/3b732824890cbdac9089da32f659f3d80279563c/crates/interpreter/src/side_table.rs#L86-L89) (the index is 97252). Do we want to change it to u32?
> 
> I suspect the error in `./scripts/hwci.sh` is due to the same reason (but I'm not sure how to check that).

I'm not sure we want to support such corner-cases. For now I'm ignoring all 149 tests in `br_table`, but maybe some of them we shouldn't. We can deal with that later.

If hwci is still failing, I'll take a look and work around it to merge. It might be because of the pop_cnt story which might be wrong and is not caught by the test suite.

ia0 (approved):

crates/interpreter/src/valid.rs
```diff
@@ -1122,6 +1149,8 @@ fn delta(
 
 fn pop_cnt(source: SideTableBranch, target: SideTableBranch) -> MResult<u32, Check> {
     let source = source.stack;
+    // TODO(dev/fast-interp): Figure out why we can't simply source.stack - target.stack and
+    // document it.
```
zhouwfang, line 1153: @ia0 I thought we shouldn't always expect `source.stack` to be `>= target.stack`. One example is the [as-if-cond](https://github.com/WebAssembly/spec/blob/2e2e3a9e471d162ce317b6d1153e8ff8b9945218/test/core/br_if.wast#L108) in `br_if.wast`. At `Else`, we have ` source=SideTableBranch { parser: [65, 2, 5, 65, 3, 11, 11, 11], branch_table: 1, stack: 0, result: 0 }, target=SideTableBranch { parser: [65, 3, 11, 11, 11], branch_table: 3, stack: 1, result: 1 }`.

zhouwfang:
> > The only test that fails in the test suite now is [the huge br_table](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107) because of ["index of MetadataEntry overflow"](https://github.com/zhouwfang/wasefire/blob/3b732824890cbdac9089da32f659f3d80279563c/crates/interpreter/src/side_table.rs#L86-L89) (the index is 97252). Do we want to change it to u32?
> > I suspect the error in `./scripts/hwci.sh` is due to the same reason (but I'm not sure how to check that).
> 
> I'm not sure we want to support such corner-cases. For now I'm ignoring all 149 tests in `br_table`, but maybe some of them we shouldn't. We can deal with that later.
> 
> If hwci is still failing, I'll take a look and work around it to merge. It might be because of the pop_cnt story which might be wrong and is not caught by the test suite.

I had checked `br_table.wast` throughly, and [large](https://github.com/WebAssembly/spec/blob/c280b356e0da830ce0b7c636c34ea684f5d1b728/test/core/br_table.wast#L107) is only function that fails due to the `u16` overflow (after I commented out this function, the test suite passes).

`pop_cnt` should be correct for `hwci` because I completed its creation in `valid.rs` and usage in `exec.rs` long time ago, and there has been no issue with `hwci`. The `hwci` error should be due to `u16` overflow.

PR #761, opened 2025-02-26
Reduce compile-time of wasmtime

ia0:
It's taking at least 5 minutes out of `test-0` making it at least 12 minutes.


ia0:
The last commit `63e3207` confirms that adding wasmtime increases `test-0` time from ~6m30 to ~12m30. Using `default-features` didn't help, so trying to cache the build (checking the size).

ia0:
This last version is better. Adding wasmtime only increases the test by 2 minutes. It's the slowest one but not a big issue. We could split from 4 to 5 parallel tests, or apply some shuffling since the variance between test shards is quite high.

ia0-review (approved):

PR #769, opened 2025-03-03
Enforce A/B update process for all platforms

ia0:


crates/cli/CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
...
-## 0.2.1-git
+## 0.3.0-git
+
+### Major
+
+- Remove `platform-update-{metadata,transfer}` in favor of `platform-update`
 
 ### Minor
 
+- Add `platform-update` for A/B platform update
```
ia0-review, line 11: We should make this command take a bundle which contains both A and B, and asks the platform which version it needs.

ia0-review (changes requested):

crates/cli/CHANGELOG.md
ia0: I did a bit simpler. The `PlatformUpdate` action takes an optional second file. If it doesn't, nothing changes. If it does, it will ask the platform which side it is running and use the appropriate file.

ia0-review (approved):

PR #770, opened 2025-03-05
Simplify `pop_cnt()`

zhouwfang:
The previous implementation was unnecessarily complicated. I hope it is convincing now.

#46


ia0:
Waiting for #771 before updating and merging.

PR #775, opened 2025-03-07
Make `metadata()`, `parse_u16()` and `view()` generic

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -26,7 +26,7 @@ use crate::*;
 pub struct Module<'m> {
     binary: &'m [u8],
     types: Vec<FuncType<'m>>,
-    side_table: SideTableView<'m, Use>,
+    side_table: SideTableView<'m>,
 }
 
 impl<'m> Import<'m> {
@@ -112,7 +112,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func_type(&self, x: FuncIdx) -> FuncType<'m> {
-        self.types[self.side_table.metadata(x as usize).unwrap().type_idx()]
+        self.types[self.side_table.metadata::<Use>(x as usize).into_ok().type_idx()]
     }
 
     pub(crate) fn table_type(&self, x: TableIdx) -> TableType {
@@ -173,7 +173,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let metadata = self.side_table.metadata(x as usize).unwrap();
+        let metadata = self.side_table.metadata::<Use>(x as usize).into_ok();
         (unsafe { Parser::new(&self.binary[metadata.parser_range()]) }, metadata.branch_table())
     }
 
--- a/crates/interpreter/src/module.rs
+++ b/crates/interpreter/src/module.rs
@@ -26,7 +26,7 @@ use crate::*;
 pub struct Module<'m> {
     binary: &'m [u8],
     types: Vec<FuncType<'m>>,
-    side_table: SideTableView<'m>,
+    side_table: SideTableView<'m, Use>,
 }
 
 impl<'m> Import<'m> {
@@ -112,7 +112,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func_type(&self, x: FuncIdx) -> FuncType<'m> {
-        self.types[self.side_table.metadata(x as usize).type_idx()]
+        self.types[self.side_table.metadata(x as usize).unwrap().type_idx()]
     }
 
     pub(crate) fn table_type(&self, x: TableIdx) -> TableType {
@@ -173,7 +173,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let metadata = self.side_table.metadata(x as usize);
+        let metadata = self.side_table.metadata(x as usize).unwrap();
         (unsafe { Parser::new(&self.binary[metadata.parser_range()]) }, metadata.branch_table())
     }
 
--- a/crates/interpreter/src/side_table.rs
+++ b/crates/interpreter/src/side_table.rs
@@ -16,7 +16,6 @@ use alloc::vec::Vec;
 use core::ops::Range;
 
 use crate::error::*;
-use crate::module::Parser;
 use crate::toctou::*;
 
 pub const SECTION_NAME: &str = "wasefire-sidetable";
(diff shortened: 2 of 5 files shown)


crates/interpreter/src/side_table.rs
```diff
@@ -14,26 +14,39 @@
...
-    pub fn metadata(&self, func_idx: usize) -> Metadata<'m> {
-        Metadata(
-            &self.metadata[parse_u16(self.indices, func_idx * 2) as usize
-                .. parse_u16(self.indices, (func_idx + 1) * 2) as usize],
-        )
+impl<'m, M: Mode> SideTableView<'m, M> {
+    pub fn metadata(&self, func_idx: usize) -> MResult<Metadata<'m>, M> {
+        try {
+            Metadata(
+                &self.metadata[parse_u16::<M>(
```
zhouwfang, line 36: Could you explain how to make "getting range" generic?

crates/interpreter/src/side_table.rs
```diff
@@ -14,26 +14,39 @@
...
+                    .. parse_u16::<M>(
+                        self.indices,
+                        func_idx
+                            .checked_add(1)
+                            .ok_or_else(|| M::invalid())?
+                            .checked_mul(2)
+                            .ok_or_else(|| M::invalid())?,
+                    ) as usize],
+            )
+        }
```
zhouwfang, line 49: How should I call `ok_or_else(|| M::invalid())` after `try` to only call it once?

crates/interpreter/src/side_table.rs
```diff
@@ -155,8 +168,8 @@ impl BranchTableEntry {
     }
 }
 
-fn parse_u16(data: &[u8], offset: usize) -> u16 {
-    u16::from_le_bytes(data[offset ..][.. 2].try_into().unwrap())
+fn parse_u16<M: Mode>(data: &[u8], offset: usize) -> MResult<u16, M> {
+    Ok(u16::from_le_bytes(data[offset ..][.. 2].try_into().unwrap()))
```
zhouwfang, line 172: I know this is probably not good. Not sure the difference between `Check` and `Use` for this function.

crates/interpreter/src/module.rs
```diff
@@ -112,7 +112,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func_type(&self, x: FuncIdx) -> FuncType<'m> {
-        self.types[self.side_table.metadata(x as usize).type_idx()]
+        self.types[self.side_table.metadata(x as usize).unwrap().type_idx()]
```
ia0, line 115:
```suggestion
        self.types[self.side_table.metadata(x as usize).into_ok().type_idx()]
```

crates/interpreter/src/module.rs
```diff
@@ -173,7 +173,7 @@ impl<'m> Module<'m> {
     }
 
     pub(crate) fn func(&self, x: FuncIdx) -> (Parser<'m>, &'m [BranchTableEntry]) {
-        let metadata = self.side_table.metadata(x as usize);
+        let metadata = self.side_table.metadata(x as usize).unwrap();
```
ia0, line 176:
```suggestion
        let metadata = self.side_table.metadata(x as usize).into_ok();
```

crates/interpreter/src/side_table.rs
```diff
@@ -14,26 +14,39 @@
...
 
 use crate::error::*;
 use crate::module::Parser;
+use crate::toctou::*;
 
 pub const SECTION_NAME: &str = "wasefire-sidetable";
 
 #[derive(Debug)]
-pub struct SideTableView<'m> {
+pub struct SideTableView<'m, M: Mode> {
```
ia0, line 26: Only the functions should be generic. The `SideTableView<'m>` doesn't need to be. It's a bit different from `Parser`.

crates/interpreter/src/side_table.rs
ia0: `self.metadata.get(range)`

crates/interpreter/src/side_table.rs
ia0: You need to put the result of the try block in a variable and then call `ok_or_else` on that variable.

```rust
let metadata: Option<_> = try { ... };
metadata.ok_or_else(M::invalid)
```

crates/interpreter/src/side_table.rs
ia0: The difference is on the `[offset ..][.. 2]` part. Right now there are checks and panics. What we want is:
- With `Check`, it should check and return and error (not panic)
- With `Use`, it should not even check. But using `M::invalid()` we can still check and the compiler will remove the check.

So we just need to implement it with something like `data.get(offset .. offset.checked_add(2)?)?` and then the `ok_or_else(invalid)`.

crates/interpreter/src/side_table.rs
```diff
@@ -137,10 +150,10 @@ impl BranchTableEntry {
     }
 
     pub fn view(self) -> BranchTableEntryView {
-        let pop_val_counts = parse_u16(&self.0, 4);
+        let pop_val_counts = parse_u16::<Use>(&self.0, 4).unwrap();
         BranchTableEntryView {
-            delta_ip: (parse_u16(&self.0, 0) as i16) as i32,
-            delta_stp: (parse_u16(&self.0, 2) as i16) as i32,
+            delta_ip: (parse_u16::<Use>(&self.0, 0).unwrap() as i16) as i32,
+            delta_stp: (parse_u16::<Use>(&self.0, 2).unwrap() as i16) as i32,
```
ia0, line 156: Same here, we want this function to be parametric over the mode. We use it in `patch_branch()` during verification. And actually, we should not panic when indexing with `[source.branch_table]` there. We should also use `get()` (unless we have a test somewhere else).

crates/interpreter/src/side_table.rs
```diff
@@ -14,26 +14,39 @@
...
+            Metadata(
+                &self.metadata[parse_u16::<M>(
+                    self.indices,
+                    func_idx.checked_mul(2).ok_or_else(|| M::invalid())?,
+                ) as usize
+                    .. parse_u16::<M>(
+                        self.indices,
+                        func_idx
+                            .checked_add(1)
+                            .ok_or_else(|| M::invalid())?
```
ia0, line 44: No need for ok_or_else here. This can just be `func_idx.checked_add(1)?.checked_mul(2)?`.

Also it's probably worth it to use some variables for the range start and limit.

crates/interpreter/src/side_table.rs
zhouwfang: It does not work directly with the error " `?` could not convert error type `Result<_, <M as Mode>::Error>` to `<M as Mode>::Error". But I use [M::open()](https://github.com/zhouwfang/wasefire/blob/188766a9e91d0f932b487dc0233d169685e9bd98/crates/interpreter/src/toctou.rs#L52) instead.

crates/interpreter/src/side_table.rs
zhouwfang: I use `M::open()` instead for the reason explained in another comment.

crates/interpreter/src/side_table.rs
zhouwfang: Where is the "indexing" in this function?

crates/interpreter/src/side_table.rs
```diff
@@ -42,7 +43,7 @@ pub struct Metadata<'m>(&'m [u8]);
 
 impl<'m> Metadata<'m> {
     pub fn type_idx(&self) -> usize {
```
ia0, line 43: `type_idx` and `branch_table` also need to be polymorphic, because they are called from `valid.rs` too. But this can be done in another PR.

ia0 (approved):
Thanks! LGTM. We can do `type_idx` and `branch_table` in another PR since this one is good.

crates/interpreter/src/side_table.rs
ia0: Yes, `M::open(|| try { })?` also works fine.

crates/interpreter/src/side_table.rs
ia0: There's no "indexing" in this function. The indexing I mentioned was where it is called in `valid.rs`. That can be done in the next PR. I hope I'll be able to find out all those issues in the final review from `dev/fast-interp` to `main`. There are actually many places where we would panic instead of returning an error on user errors.

PR #776, opened 2025-03-07
Fix branch table logic

ia0:
The main changes are:
- Do not create a branch table entry for unreachable code. This saves space and permits a uniform treatment of branches.
- Use the correct stack length for the Loop branch target (we're taking the branch before being in the label, so the current stack needs to be added).
- Take the Else source and target branches at the correct time (i.e. when the stack is the expected one such that the stack length is correctly computed).


ia0-review (approved):

zhouwfang:
#46

PR #777, opened 2025-03-07
Modify the br_table test before parsing

ia0:
This way we can make sure which part of it is unsupported, namely the large table.


ia0-review (approved):

ia0-review (approved):

zhouwfang:
#46

PR #784, opened 2025-03-07
Enable `hw-host` in CI

zhouwfang:
#46

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
           x     tests-1  pull_request  push  schedule
           x     tests-2  pull_request  push  schedule
           x     tests-3  pull_request  push  schedule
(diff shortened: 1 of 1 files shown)


zhouwfang:
Somehow the CI is still queued after 40min.

zhouwfang:
Do you know why this still fails?

ia0:
> Do you know why this still fails?

It's because we don't prepare the side tables for the applets. I'm sending #786.

ia0:
By the way, to debug you can use in one terminal:

```
RUST_BACKTRACE=1 cargo xtask runner host --features=wasefire-interpreter/debug flash --protocol=unix
```

And in another:

```
cargo xtask applet rust cbc_test install --protocol=unix
```

When there's an error in the interpreter, you'll get a backtrace. You can also add printf statements.

PR #737, opened 2025-02-12
Add initial version of opensk example

ia0:
This is part of #424. This needs more work to fix it.

PR #785, opened 2025-03-08
Improve `offset_front()`

zhouwfang:
#46

### DIFF
--- a/crates/interpreter/src/util.rs
+++ b/crates/interpreter/src/util.rs
@@ -12,9 +12,123 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::error::*;
+
 // TODO(dev/fast-interp): Add debug asserts when `off` is positive and negative, and `toctou`
 // support.
 pub fn offset_front<T>(cur: &[T], off: isize) -> &[T] {
+    unsafe {
+        core::slice::from_raw_parts(cur.as_ptr().offset(off), (cur.len() as isize - off) as usize)
+    }
+}
+
+#[allow(dead_code)]
+pub fn offset_front_check<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> Result<&'a [T], Error> {
+    if beg.is_empty() || cur.is_empty() {
+        #[cfg(feature = "debug")]
+        eprintln!("One or both slices are empty.");
+        return Err(invalid());
+    }
+    if beg.last().unwrap() as *const T != cur.last().unwrap() as *const T {
+        #[cfg(feature = "debug")]
+        eprintln!("The last elements of beg and cur do not match.");
+        return Err(invalid());
+    }
+    let mut cur_offset = beg.len().checked_sub(cur.len()).ok_or_else(|| {
+        #[cfg(feature = "debug")]
+        eprintln!("Cur length {} exceeds beg length {}.", cur.len(), beg.len());
+        invalid()
+    })?;
+    if off > 0 {
+        cur_offset = cur_offset.checked_add(off as usize).ok_or_else(|| {
+            #[cfg(feature = "debug")]
+            eprintln!("Addition overflow");
+            invalid()
+        })?;
+        if cur_offset < cur.len() {
+            #[cfg(feature = "debug")]
+            eprintln!("The offset overflows the slice length.");
+            return Err(invalid());
+        }
+    } else {
+        cur_offset = cur_offset.checked_sub(off as usize).ok_or_else(|| {
+            #[cfg(feature = "debug")]
+            eprintln!("Subtraction overflow");
+            invalid()
+        })?;
+        if cur_offset >= cur.len() {
+            #[cfg(feature = "debug")]
+            eprintln!("The offset overflows below 0.");
+            return Err(invalid());
+        }
+    }
+    Ok(&beg[cur_offset ..])
+}
+
+#[allow(dead_code)]
+#[allow(unused_variables)]
+pub fn offset_front_use<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> &'a [T] {
+    #[cfg(not(feature = "toctou"))]
+    return offset_front_unsafe(beg, off);
+    #[cfg(feature = "toctou")]
+    offset_front_toctou(beg, cur, off)
+}
+
+#[allow(dead_code)]
+fn offset_front_toctou<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> &'a [T] {
+    if beg.is_empty() || cur.is_empty() {
+        #[cfg(feature = "debug")]
+        eprintln!("One or both slices are empty.");
+        panic!();
+    }
+    if beg.last().unwrap() as *const T != cur.last().unwrap() as *const T {
+        #[cfg(feature = "debug")]
+        eprintln!("The last elements of beg and cur do not match.");
+        panic!();
+    }
+    let mut cur_offset = beg
+        .len()
+        .checked_sub(cur.len())
+        .ok_or_else(|| {
+            #[cfg(feature = "debug")]
+            eprintln!("Cur length {} exceeds beg length {}.", cur.len(), beg.len());
+            invalid()
+        })
+        .unwrap();
+    if off > 0 {
+        cur_offset = cur_offset
+            .checked_add(off as usize)
+            .ok_or_else(|| {
+                #[cfg(feature = "debug")]
+                eprintln!("Addition overflow");
+                invalid()
+            })
+            .unwrap();
+        if cur_offset < cur.len() {
+            #[cfg(feature = "debug")]
+            eprintln!("The offset overflows the slice length.");
+            panic!();
+        }
+    } else {
+        cur_offset = cur_offset
+            .checked_sub(off as usize)
+            .ok_or_else(|| {
+                #[cfg(feature = "debug")]
+                eprintln!("Subtraction overflow");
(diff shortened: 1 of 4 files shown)


crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,22 @@
...
-pub fn offset_front<T>(cur: &[T], off: isize) -> &[T] {
-    unsafe {
-        core::slice::from_raw_parts(cur.as_ptr().offset(off), (cur.len() as isize - off) as usize)
-    }
+use crate::toctou::*;
+
+// TODO(dev/fast-interp): Add debug asserts when `off` is positive and negative.
+pub fn offset_front<T, M: Mode>(cur: &[T], off: isize) -> MResult<&[T], M> {
+    M::choose(
+        || unsafe {
```
zhouwfang, line 20: Not sure how to make the check mode "safe".

ia0: That's the difficulty. We need an extra argument, but only in `Check` mode (and also `Use` mode without `toctou`). This needs quite some amount of changes, here's a few:
- Change the signature to `fn offset_front<T, M: Mode>(beg: &[T], cur: &[T], off: isize) -> MResult<&[T], M>`
- For each call-site provide `beg`, which is often not available, so needs additional changes.
- For example, we could have a `MetadataVerify` the same way we have `SideTableVerify` in `impl ValidMode for Verify` and this `MetadataVerify` would hold the parser at the beginning of the function.

This is a bit more problematic for `exec.rs`. We probably want to gate the added code behind `toctou` feature. In particular, since it doesn't seem we call `offset_front` from polymorphic context, we could simply expose 3 functions:
- `offset_front_check(beg, cur, off) -> Result<&[T], Error>`
- `#[cfg(feature = "toctou")] offset_front(beg, cur, off) -> &[T]`
- `#[cfg(not(feature = "toctou")] offset_front(cur, off) -> &[T]`

zhouwfang: What does "beg" mean?

ia0: "begin", it's usually `beg`, `cur`, `end`. But here we don't need `end` because we have the length (which is actually redundant in `beg` and `cur`, but that's fine).

zhouwfang:
Could you first take a look at the three functions? I'll apply them later. Thanks.

crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,124 @@
...
     unsafe {
         core::slice::from_raw_parts(cur.as_ptr().offset(off), (cur.len() as isize - off) as usize)
     }
 }
+
+#[allow(dead_code)]
+pub fn offset_front_check<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> Result<&'a [T], Error> {
+    if beg.is_empty() || cur.is_empty() {
+        #[cfg(feature = "debug")]
+        eprintln!("One or both slices are empty.");
```
ia0: This should not be an error. It should be valid to call `offset_front_check(xs, xs, 0)` where `xs.is_empty()`.

crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,124 @@
...
+        return Err(invalid());
+    }
+    if beg.last().unwrap() as *const T != cur.last().unwrap() as *const T {
+        #[cfg(feature = "debug")]
+        eprintln!("The last elements of beg and cur do not match.");
```
ia0: There's actually a function that does that: https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.subslice_range

You'll need to add `#![feature(substr_range)]` in `lib.rs`.

crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,124 @@
...
+            eprintln!("Subtraction overflow");
+            invalid()
+        })?;
+        if cur_offset >= cur.len() {
+            #[cfg(feature = "debug")]
+            eprintln!("The offset overflows below 0.");
+            return Err(invalid());
+        }
+    }
+    Ok(&beg[cur_offset ..])
```
ia0: I believe it's possible to do much simpler. Pushing a review commit.

crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,124 @@
...
+}
+
+#[allow(dead_code)]
+#[allow(unused_variables)]
+pub fn offset_front_use<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> &'a [T] {
+    #[cfg(not(feature = "toctou"))]
+    return offset_front_unsafe(beg, off);
```
ia0: This is wrong. It should be `cur, off`. But anyway, we don't need this intermediate function.

crates/interpreter/src/util.rs
```diff
@@ -12,10 +12,124 @@
...
+    #[cfg(feature = "toctou")]
+    offset_front_toctou(beg, cur, off)
+}
+
+#[allow(dead_code)]
+fn offset_front_toctou<'a, T>(beg: &'a [T], cur: &'a [T], off: isize) -> &'a [T] {
```
ia0: This could theoretically just be `offset_front_check(beg, cur, off).unwrap()`. But I'm pushing another relatively simple implementation which doesn't check the end.

ia0 (commented):
Looks good, I've simplified a bit.

crates/interpreter/src/util.rs
zhouwfang: The advantage of having this intermediate function seems to simplify the call site code (we would only need to call one function instead of two).

ia0: But we don't want to pass the extra parameter. Also we don't need to call twice, we can gate only the first argument.

zhouwfang: Sorry I didn't really mean "calling twice" but writing the two versions twice at the call site.

The "tradeoff" in my mind is
Option 1: With intermediate function, we only need to write one function at call site. But `beg` is redundant for the "not toctou" case in the intermediate function.
Option 2: Without intermediate function, we need to write two functions at call site (of course only one of them is actually called based on the existence of the toctou flag). But there is no issue with `beg` redundancy for the "not toctou" case.

ia0: It should be possible to do this:

```rust
offset_from(#[cfg(feature = "toctou")] beg_expr, cur_expr, off_expr)
```

And you get both benefits: one call-site, no extra parameter.

zhouwfang: Good idea, but the  compiler is confused about the lifetime. "Not toctou" version has explicit lifetime while toctou version does not have. When I add the lifetime in the intermediate function, the complier asks me to elide it. When I don't add it, the compiler asks me to add it.

ia0: This is easier to address with code. Feel free to just push a WIP commit that I can take a look to. I suspect we might just want to disable the lint, but I need to see the actual problem to be sure.

zhouwfang: Pushed. That's very possible as the error appeared after the test suite passes.

ia0: Ah ok I see. I meant to use `offset_from(#[cfg(feature = "toctou")] beg_expr, cur_expr, off_expr)` at call-site, not definition time. For the definition you can use my commit.

zhouwfang:
After this PR is merged, can we merge `dev/fast-interp` into `main`?

crates/interpreter/src/valid.rs
```diff
@@ -1001,8 +1011,8 @@ impl<'a, 'm, M: ValidMode> Expr<'a, 'm, M> {
                 check(label.type_.params == label.type_.results)?;
                 if let Some(source) = source {
                     let source = self.branch_table.as_ref().unwrap().patch_branch(source)?;
-                    // SAFETY: This function is only called after parsing an End instruction.
-                    target.parser = offset_front(target.parser, -1);
+                    // This function is only called after parsing an End instruction.
+                    target.parser = offset_front_check(self.context.func_body, target.parser, -1)?;
```
ia0: We don't need a check here, since it's not based on user input.

ia0 (approved):
Thanks! I've changed the logic to have the information at the right place (essentially `func_body` should be in `Expr`).

ia0:
> After this PR is merged, can we merge `dev/fast-interp` into `main`?

Yes, I'll create a PR from a fork of this branch in my repo to avoid the PR getting double CI (once for push and once for pull_request).

zhouwfang:
> > After this PR is merged, can we merge `dev/fast-interp` into `main`?
> 
> Yes, I'll create a PR from a fork of this branch in my repo to avoid the PR getting double CI (once for push and once for pull_request).

Do you mind explaining the differences in more detail? You are ahead of me. Thanks!

ia0:
> Do you mind explaining the differences in more detail? You are ahead of me. Thanks!

#737 is the PR where I merged the `dev/opensk` branch into `main`. You can see (if you click `View details` near the end) that there were 40 checks, while in this PR it's 21. There were twice more checks because for each commit I pushed to the branch after the PR was created, a `push` for `dev/opensk` and a `pull_request` for `main` workflow was created. You can see this in the `checks` job (which appears twice like almost all other jobs):
- https://github.com/google/wasefire/actions/runs/13283440451/job/37086733921 we can see `Compute the checks for pull_request`
- https://github.com/google/wasefire/actions/runs/13283439320/job/37086719317 we can see `Compute the checks for push`

PR #786, opened 2025-03-10
Re-enable hwci host in dev/fast-interp

ia0:
Supersedes #784


ia0-review (approved):

ia0:
Ah, the opensk applet is a problem. We don't use wasm yet, so let's try to disable it at least in the `dev/fast-interp` branch.

ia0-review (approved):

PR #814, opened 2025-04-10
Update opentitan from z1 to a2

ia0:
This was actually tested on a1, but there should be no software difference.


ia0:
The flash regions 3 and 4 have been swapped.

ia0:
HW CI passed

ia0-review (approved):

PR #1113, opened 2026-05-06
Use Tombi default rules for TOML formatting

ia0:


ia0:
> The `sed` logic for reordering the `lints` property in the JSON schema appears to be incorrect

It's subtle because it prints the same `    },` twice and skips one.

ia0-review (approved):

PR #1139, opened 2026-08-23
Enable unreachable_pub lint in the interpreter

### LINKED ISSUE #823: Unsound get_unchecked usages at 0.3.0
crates/interpreter/src/toctou.rs:111
```rust
pub fn get(xs: &[u8], i: usize) -> u8 {
    #[cfg(not(feature = "toctou"))]
    return unsafe { *xs.get_unchecked(i) };
    #[cfg(feature = "toctou")]
    xs[i]
}

pub fn split_at(xs: &[u8], mid: usize) -> (&[u8], &[u8]) {
    #[cfg(not(feature = "toctou"))]
    return unsafe { xs.split_at_unchecked(mid) };
    #[cfg(feature = "toctou")]
    xs.split_at(mid)
}
```

Functions `get` and `split_at`are public and safe, they accept parameters and used in unsafe functions without sufficient checks (when certain feature flag is set up), which might cause memory risks. In Rust, we should not face any security risks when merely using safe function.

rootkiller6788:
Fixes #823.

This enables the `unreachable_pub` lint for the interpreter crate as suggested in the issue:

- Remove `interpreter` from the exemption list in `scripts/sync.sh`.
- Regenerate the `[lints]` section of `crates/interpreter/Cargo.toml` (adds `rust.unreachable_pub = "warn"`).
- Fix all 142 `unreachable_pub` warnings by narrowing the visibility of internal items to `pub(crate)` (the lint suggestions), and removing the unnecessary `pub` on the `parse_leb128_is_correct` test.

No public API is affected: only items already re-exported from `lib.rs` remain `pub`. Verified with the commands run by the crate's `test.sh` (via `test-helper.sh`): `cargo check --lib` and clippy equivalents compile warning-free for the default, `debug,toctou`, `float-types,vector-types`, and all-features configurations, for the `thumbv7em-none-eabi` and `riscv32imc-unknown-none-elf` targets, and for the `hello` example and the spec test target. `cargo test --lib --features=debug,toctou` passes and `cargo fmt -- --check` is clean.

### DIFF
--- a/crates/interpreter/CHANGELOG.md
+++ b/crates/interpreter/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Patch
 
+- Enable `unreachable_pub` lint
 - Update dependencies
 
 ## 0.5.0
--- a/crates/interpreter/Cargo.toml
+++ b/crates/interpreter/Cargo.toml
@@ -35,4 +35,5 @@ vector-types = []
 clippy.mod_module_files = "warn"
 clippy.uninlined_format_args = "allow"
 clippy.unit_arg = "allow"
+rust.unreachable_pub = "warn"
 rust.unused_crate_dependencies = "warn"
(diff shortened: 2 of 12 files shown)


ia0 (approved):
Thanks!

Issue #565, opened 2024-08-05
Fix all lints currently disabled in scripts/sync.sh


ia0:
We want to enable the following lints:
- [ ] `rust.elided-lifetimes-in-paths`
- [ ] `rust.missing-debug-implementations` (not sure if good thing)
- [ ] `rust.missing-docs` for all published libraries
- [ ] `rust.unreachable-pub`
- [ ] `rust.unused-results`

devesh-2002:
May I take this issue?

ia0:
Of course, that would be great. That's a good first issue. You can start by changing those 2 lines:

https://github.com/google/wasefire/blob/5da7a5e0de8545825ce6ca023c389f2e35926c02/scripts/sync.sh#L33-L34

into (I'm removing `missing-debug-implementations` because I'm not yet convinced it's a good thing to have):

```rust
  case $crate in
    prelude) add_lint $file warn rust.elided-lifetimes-in-paths ;;
  esac
```

Then run `./scripts/sync.sh` (note that you might need to run `./scripts/setup.sh` the very first time you clone the repository). This will add the lint to the prelude Cargo.toml. Then you can run `./test.sh` from the `crates/prelude` directory and fix any warning.

Once there are no warnings, you can create your first PR.

Then, you can add more and more crates to the case pattern, like:

```rust
  case $crate in
    board|prelude) add_lint $file warn rust.elided-lifetimes-in-paths ;;
  esac
```

And create one PR for each non-trivial crate (you can group multiple crates together if there's nothing or almost nothing to change).

The final state (i.e. when adding the last remaning crate) should simply be:

```rust
  add_lint $file warn rust.elided-lifetimes-in-paths
```

Thanks! Feel free to ask questions if you're blocked.

devesh-2002:
Thank you @ia0, for the detailed instructions!

devesh-2002:
@ia0 , Hello. I ran `./scripts/setup.sh` and then `./scripts/sync.sh`. But it gives me a duplicate key error. Would you please guide me in resolving this error?
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.36s
     Running `target/debug/xtask update-apis`
"cargo" "run" "--manifest-path=crates/xtask/crates/update-api/Cargo.toml" "--features=wasefire-applet-api-desc/full-api" "--" "--lang=assemblyscript" "--output=examples/assemblyscript/api.ts"
[...]

PR #1141, opened 2026-08-27
Enable unreachable_pub lint in the scheduler

56steve:
## What

Enables the `rust.unreachable_pub` lint for the `wasefire-scheduler` crate:

- Removes `scheduler` from the `unreachable_pub` exclusion list in `scripts/sync.sh`.
- Adds `rust.unreachable_pub = "warn"` to `crates/scheduler/Cargo.toml` (matching the ordering `sync.sh` generates).
- Restricts crate-internal items to `pub(crate)` across the `call`, `event`, `applet`, `perf`, and `protocol` modules. The crate's public API (`lib.rs`) is unchanged.
- Adds a CHANGELOG entry.

## Why

Continues the incremental work of #565 ("Fix all lints currently disabled in scripts/sync.sh"), following #1139 which enabled the same lint for the interpreter.

## Verification

Ran the full feature/target matrix from `crates/scheduler/test.sh` on host:

- `cargo check --lib` for: host `wasm,std,log`; `i686-unknown-linux-gnu` `native,std(,log)`; `thumbv7em-none-eabi` `wasm`/`pulley`/`native`, each with and without `defmt` — **all report zero `unreachable_pub` warnings**.
- `cargo test --lib --features=_test,full-api,wasm,std` passes.
- `cargo clippy` and `cargo fmt --check` pass.

Note: the full Linux CI suite (`scripts/ci.sh`) was not run locally.

### DIFF
--- a/crates/scheduler/CHANGELOG.md
+++ b/crates/scheduler/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Patch
 
+- Enable `unreachable_pub` lint
 - Disable `unused-features` lint
 - Use heterogeneous try blocks when needed
 - Update dependencies
--- a/crates/scheduler/Cargo.toml
+++ b/crates/scheduler/Cargo.toml
@@ -346,4 +346,5 @@ internal-hash-context = []
 clippy.mod_module_files = "warn"
 clippy.uninlined_format_args = "allow"
 clippy.unit_arg = "allow"
+rust.unreachable_pub = "warn"
 rust.unused_crate_dependencies = "warn"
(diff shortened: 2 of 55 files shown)


ia0:
By the way, thanks for the PRs. I'm currently on vacation, but will review them when I'm back (some time end of next week).

ia0 (approved):

56steve:
Hi @ia0, welcome back, and thanks for the review!

Same note as on the xtask PR: the sync failure here was only the rustup submodule drift, not this change. I've rebased onto the latest main (past #1144, rustup 1.29.1), so it's resolved. I re-verified every feature and target combination from scheduler/test.sh locally and they all come back with zero unreachable_pub warnings, plus tests, clippy and fmt are clean.

The fork PR seems to be waiting on a maintainer to approve the workflow run before CI runs again. Once you trigger it, it should pass. Thanks!

PR #1146, opened 2026-09-02
Upgrade all dependencies

ia0:


ia0-review (changes requested):
Let's update all CHANGELOGs appropriately.

ia0:
OpenTitan needs to be updated separately because there's a manifest version 3 now apparently.

ia0-review (approved):

